Issue #8, opened 2022-12-09
Input/feature symbol overlap


kylebgorman:
[copied from CUNY-CL/abstractness/issues/127]

If there is any overlap between input symbols and feature symbols, the two are treated the same in several models. The simplest solution would be to use some kind of special string munging to prevent this overlap, perhaps by wrapping the features like `f"{[feature]}`. 

Alternatively, this might motivate using separate encoders for feature-full LSTM and transformer models.

Issue #7 is spiritually related.

**This is a blocker for a post-beta release candidate.**

bonham79:
Closing issue with https://github.com/CUNY-CL/yoyodyne/pull/18

Issue #7, opened 2022-12-09
Unseen symbols during inference


kylebgorman:
[copied from CUNY-CL/abstractness/issues/134]

Currently, the presence of a symbolsnot seen during training causes crash during transducer inference, and I am not sure what happens in the same situation in other architectures. I can imagine two solutions:

1. One could provide an "unk"-ing script which takes train, dev, and test, and then "unk"s (i.e., replaces with the reserved `<UNK>` symbol) any symbols present in dev or test but not in train, and also any symbol below a certain threshold (in case one wants to learn embeddings for this unk character).
2. `yoyodyne-predict` could insert `<UNK>` programmatically where needed during inference.

I think I prefer (2), it's one less step for the user and the unk embedding thing is a bit exotic for our small-vocabulary setting.

**This is a blocker for a post-beta release candidate.**

bonham79:
It would cause a crash in any architecture that has separate features (i.e. pointer-generator). Will be ignored by those models that don't distinguish. However, still a bug since the "<UNK>" character is for input symbols, not features. Likely not the best idea. 

Recent PR makes it that any new feature in validation/test raises an error in the dataloader.

Adamits:
@bonham79 I think one thing that is unique in the transducer though is that it builds an action vocabulary. I believe the action vocabulary will never have an UNK in it, whereas the ptr-generator has *only* the dataset vocabulary, which has an UNK inserted into it manually when creating the `Dataset` class.

I think the ptr-gen issue is related specifically to an UNK feature (as opposed to an UNK character) right?

kylebgorman:
The current behavior (crash with an informative error message on separate-feature models) is superior to the previous state of affairs, but I think it should be handled programmatically and consistently across models instead of resulting in an error.

bonham79:
@Adamits correct, the action vocabulary only will produce vocabulary seen in the training set and has no equivalent of <UNK> (outside of simply copying or deleting an UNK character). 

The ptr-gen issue would be specifically the UNK feature issue. While other models have a more subtler issue of using the UNK _character_ embedding for an UNK feature. This is currently not an issue since the raise error flag stops passing of new features.

[...]

PR #18, opened 2022-12-14
Unk feat

bonham79:
For https://github.com/CUNY-CL/yoyodyne/issues/7 and https://github.com/CUNY-CL/yoyodyne/issues/8 . Now creates an UNK_FEAT symbol for encoding unseen features during validation. Features are also formatted as `<FEAT>` in symbol lookup.

Note, this could also be done by sharing UNK symbol between character and feature. Adding the extra symbol allows later management of unknown features that is separate from unknown characters.

### DIFF
--- a/yoyodyne/datasets.py
+++ b/yoyodyne/datasets.py
@@ -216,7 +216,6 @@ class DatasetNoFeatures(data.Dataset):
         self,
         symbol2i: Dict,
         word: List[str],
-        unk: int,
         add_start_tag: bool = True,
         add_end_tag: bool = True,
     ) -> torch.Tensor:
@@ -241,7 +240,7 @@ class DatasetNoFeatures(data.Dataset):
         if add_end_tag:
             sequence.append(special.END)
         return torch.LongTensor(
-            [symbol2i.get(symbol, unk) for symbol in sequence]
+            [symbol2i.get(symbol, self.unk_idx) for symbol in sequence]
         )
 
     def _decode(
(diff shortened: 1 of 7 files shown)


yoyodyne/datasets.py
```diff
@@ -464,6 +470,9 @@ def _iter_samples(
                 features = self._get_cell(
                     row, self.features_col, self.features_sep
                 )
+                # Use unique encoding for features.
+                # This disambiguates from overlap with source.
+                features = [f"<{f}>" for f in features]
```
kylebgorman, line 475: We should use some other style for this, as it could easily clash with the other special characters (for instance if a feature name is `S` or `E`). I am going to suggest square brackets. Then the unknown feature is literally just `<UNK>` in feature form, namely `[<UNK>]`

yoyodyne/special.py
```diff
@@ -6,5 +6,6 @@
 START = "<S>"
 END = "<E>"
 UNK = "<UNK>"
+UNK_FEAT = "<UNK_FEAT>"
```
kylebgorman, line 9: I am not sure this really belongs here if it ought not to be part of SPECIAL below. Alternatively it could live solely inside of `datasets`.

kylebgorman (commented):
A few comments.

bonham79:
@kylebgorman had to change the transducer around to accommodate unknowns in a different way. Let me know if it's a bit too clunky. (Keep in mind it'll be changed when transducer embeddings get updated.)

kylebgorman (commented):
LGTM. Does this need more testing or can I submit?

The comments seem to compensate for any additional complexity.

kylebgorman (approved):

bonham79:
> LGTM. Does this need more testing or can I submit?
> 
> The comments seem to compensate for any additional complexity.

I ran tests and everything seems kosher.

Issue #16, opened 2022-12-09
Benchmarking


kylebgorman:
We should add a benchmarking suite. I have reserved a separate repo, CUNY-CL/yoyodyne-benchmarks for this.

Here are a list of shared tasks (and related papers) from which we can pull data:

* [Merhav & Ash (2018) transliteration](https://github.com/steveash/NETransliteration-COLING2018/tree/master/data)
* Other transliteration tasks:
   - [Dakshina](https://github.com/google-research-datasets/dakshina)
   - [Japanese/English reverse transliteration](https://github.com/google-research/google-research/tree/master/pair_ngram)
* [SIGMORPHON 2016 inflection](https://sigmorphon.github.io/sharedtasks/2016/)
* [SIGMORPHON 2017 inflection](https://sigmorphon.github.io/sharedtasks/2017/task/)
* [SIGMORPHON 2018 inflection](https://sigmorphon.github.io/sharedtasks/2018/task1/)
* [SIGMORPHON 2020 g2p](https://sigmorphon.github.io/sharedtasks/2020/task1/)
* [SIGMORPHON 2021 g2p](https://github.com/sigmorphon/2021-task1)
* [New York-Boulder abstractness data](https://github.com/CUNY-CL/abstractness)

The benchmark itself is a collection of two tables.

* A "KPI" table, per dataset/language. E.g., "SIGMORPHON 2021 g2p Bulgarian".
* A "study" table, per dataset/language/architecture. E.g., "Transducer ensemble on SIGMORPHON 2021 g2p Bulgarian".

A single script should compute all KPI statistics and dump it out as a TSV. This table should include:

* training set size
* dev set size
* test set size
* average input string length
* average output string length
* whether it has features

While one could imagine a single script which performs all studies, this is probably not wise. Rather, these should be grouped into separate scripts based on their functionality (though it may make sense for there to be multiple studies per study script; e.g., we coudl have one script per dataset/language pair). The results can be dumped out in some structured format (JSON), and a separate script can be used to aggregate the non-ragged portions of all the JSON study reports into a single TSV. This table should include:

* dataset
* language
* model type
* GPU models (e.g., `torch.cuda.get_device_name(number)`)
* wall clock time during training
* wall clock time during inference
* development accuracy of best model
* test accuracy of best model
* hyperparameters for best model (this is the ragged part)
* model size, in KB
* model size, in # of trainable parameters

Then, a separate script is used to aggregate the non-ragged portions of the extant study observations.

[...]

Issue #15, opened 2022-12-09
Sweeping


kylebgorman:
W&B provides a [simple interface for running hyperparameter sweeps](https://docs.wandb.ai/guides/sweeps). To support this, we need to refactor the train and predict functions to read hyperparameter settings from elsewhere, and we probably also want an alternative command-line launcher to `yoyodyne-train`.

This is related to #5, as the refactoring there should make this much easier.

Issue #5, opened 2022-12-09
Testing


kylebgorman:
[copied from CUNY-CL/abstractness/issues/87]

We should add integration tests (I hesitate to call these unit tests), simply limiting ourselves to the model sizes and data quantities we can run on CircleCI's free tier. We get 6,000 compute-minutes per month...all of this is pretty generous except that I am unclear whether we can use their GPU images or are stuck on CPU (ideally we'd parameterize tests on both). I think it ought to be possible to do actual training of the major models using, say, 1,000 examples. Unit tests could include g2p (for feature-less) and inflection (for feature-full) from SIGMORPHON.

The current training and prediction functions are structured to read and write directly to the file system. They should be modularized to take ordinary arguments and return the results:

* for training, a function could simply return the best model (or its path) with metadata (wall clock time, training accuracy, development accuracy), and then the command-line enabled version of that loop could invoke this
* for prediction, a function could simply return the accuracy.

These functions can then be called by the existing (null return type) training and prediction functions, the ones parameterized with `click` flags.

This will also support two other projects (issues coming soon):

* benchmarking
* W&B-enabled hyperparameter sweeping

**This is a blocker for a post-beta release candidate.**

kylebgorman:
[Yoyodyne test strategy.pdf](https://github.com/CUNY-CL/yoyodyne/files/10326390/Yoyodyne.test.strategy.pdf)

The above describes my current thinking about the test strategy.

PR #22, opened 2022-12-28
Adds dataset configuration class

kylebgorman:
This new dataclass and module store information about TSV file generation and parsing. This substantially shortens the very long `datasets` module and turns 7 separate arguments passed to the dataset constructors to 1.

This is part of my larger refactoring to make #5, #15, and #16 easier to implement.

This was tested against our Maori (no features) and Polish (yes features) data.


yoyodyne/dataconfig.py
```diff
@@ -0,0 +1,158 @@
...
+            target (List[str]).
+            features (List[str], optional).
+
+        Returns:
+            List[str].
+        """
+        row = [""] * max(self.source_col, self.target_col, self.features_col)
+        # -1 because we're using base-1 indexing.
+        row[self.source_col - 1] = self.source_sep.join(source)
+        row[self.target_col - 1] = self.target_sep.join(source)
```
bonham79, line 154: This should be `join(target)`

yoyodyne/dataconfig.py
```diff
@@ -0,0 +1,158 @@
...
+        if self.has_features:
+            assert features is not None, "Expected features"
+            row[self.features_col - 1] = self.features_sep.join(source)
```
bonham79, line 157: `join(features)`

yoyodyne/dataconfig.py
```diff
@@ -0,0 +1,158 @@
...
+
+        Args:
+            source (List[str]).
+            target (List[str]).
+            features (List[str], optional).
+
+        Returns:
+            List[str].
+        """
+        row = [""] * max(self.source_col, self.target_col, self.features_col)
```
bonham79, line 151: Possible user error here: since we don't limit the columns to only three, there's an event someone specifies `self.features_col = 5` and this will be jumbled. This could come up if someone wanted to use a more verbose tsv for yoyodyne.

yoyodyne/train.py
```diff
@@ -141,7 +148,6 @@ def main(
     label_smoothing,
     gradient_clip,
     batch_size,
-    eval_batch_size,
```
bonham79, line 144: Removing an `eval_batch_size` is going to make training runs slower. Some of our models evaluate at about 2-5x the size of training. I think the extra variable is worthwhile.

bonham79 (commented):
@Adamits you should still have final review but I left some first pass comments.

yoyodyne/dataconfig.py
kylebgorman: +1 thank you.

yoyodyne/dataconfig.py
kylebgorman: +1 thank you.

yoyodyne/dataconfig.py
kylebgorman: I consider what it does here to be the best thing you could do in that scenario---it inserts blank columns to pad so you get what you asked for. Do you think it should error out instead?

yoyodyne/train.py
kylebgorman: Why don't we set it to 2x batch size? (Or default it to that?) Thoughts?

yoyodyne/dataconfig.py
bonham79: Since you're assuming `targets` it's already going to be length 2 at minimum. So something along the lines of 
```
    row = ["",""]
    if features is not none:
         row.extend("")
```
Or something along those lines. This will remove the need to assume anything about the columns (which shouldn't influence any writing behavior anyhow). 

I also think `row` should be `_row` since it's more of a sub function only intended for writing to file.

yoyodyne/train.py
bonham79: Setting default as 2x sounds safe. Should allow best of both worlds.

yoyodyne/dataconfig.py
kylebgorman: Discussed offline. I think this is operating as intended, @bonham79 has some alternative proposals to work out, going to table this but he may put together a PR or request-for-comments for discussion.

yoyodyne/datasets.py
```diff
@@ -45,100 +29,35 @@ class DatasetNoFeatures(data.Dataset):
...
         special_vocabulary = special.SPECIAL.copy()
         target_vocabulary: Set[str] = set()
         source_vocabulary: Set[str] = set()
-        for source, target in self.samples:
-            source_vocabulary.update(source)
-            # Only updates if target.
-            if self.target_col:
+        if self.config.has_target:
+            for source, target in self.samples:
+                source_vocabulary.update(source)
```
Adamits, line 53: I didn't realize `update` behaved this way when passed a string, nice :).

yoyodyne/dataconfig.py
```diff
@@ -0,0 +1,158 @@
...
+        else:
+            return (
+                self.source_target_samples(filename)
+                if self.has_target
+                else self.source_samples(filename)
+            )
+
+    # Writer help.
+
+    def row(
```
Adamits, line 135: Suggest `make_row`

Adamits (approved):
My comments are nits on the method names, but not sure there is any change necessary.

Otherwise LGTM.

yoyodyne/dataconfig.py
kylebgorman: That's what Travis wanted to do too, it sounds like. I'm not convinced it wouldn't be done easier in pure UNIX if someone wants that. But I'll be overruled if you both feel this way.

yoyodyne/dataconfig.py
```diff
@@ -0,0 +1,158 @@
...
+            tsv_reader = csv.reader(source, delimiter="\t")
+            for row in tsv_reader:
+                source = self._get_cell(row, self.source_col, self.source_sep)
+                features = self._get_cell(
+                    row, self.features_col, self.features_sep
+                )
+                target = self._get_cell(row, self.target_col, self.target_sep)
+                yield source, features, target
+
+    def samples(self, filename: str) -> Iterator[Tuple[List[str], ...]]:
```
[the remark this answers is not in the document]
kylebgorman: I noticed you used `get_*` somewhere else too. What makes a function a `get_*` function for you? I can follow these principles...

yoyodyne/dataconfig.py
kylebgorman: Done.

yoyodyne/dataconfig.py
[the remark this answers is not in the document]
Adamits: I suppose I can't think of an exact principle I follow. I just typically try to prefix methods with some verb unless it is a property.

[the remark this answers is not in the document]
Adamits: I also looked briefly and cannot find any documented suggestion for this 'principle', so feel free to ignore :)

Adamits (approved):

kylebgorman:
Let me ask the obvious @Adamits: do you think this is an improvement? I thought the modularity was nice (and shortening one of the longer files was a nice side effect) but you can be honest...

Adamits:
@kylebgorman I have thought about it, and I think that it *is* an improvement. Disentangling the information for extracting everything from a file (dataconfig), and the object that actually instantiates the vocab and samples (dataset) seems nice.

With this kind of change my main concern would be about adding more pieces that a user needs to understand in order to modify the codebase, but I think this one is pretty intuitive!

PR #24, opened 2023-01-01
Adds simple getter tests

kylebgorman:
These test the class getters for collators, datasets, and models.

PR #22 will require some minor changes to the dataset tests but I do not anticipate any problems.

Addresses, but does not complete, issue #5.


yoyodyne/datasets.py
```diff
@@ -59,21 +59,19 @@ def __init__(
...
+            source_sep (str): separator character between symbols in source
+                string. "" treats each character in source as a symbol.
+            target_sep (str): separator character between symbols in target
+                string. "" treats each character in target as a symbol.
             **kwargs: ignored.
         """
         if source_col < 1:
             raise Error(f"Invalid source column: {source_col}")
         self.source_col = source_col
-        if target_col == 0:
```
bonham79, line 75: I think this logging should still be present (unless we move it to `train`)

yoyodyne/datasets.py
```diff
@@ -465,9 +453,6 @@ def _iter_samples(
                 features = self._get_cell(
                     row, self.features_col, self.features_sep
                 )
-                # Use unique encoding for features.
-                # This disambiguates from overlap with source.
-                features = [f"[{feature}]" for feature in features]
```
bonham79, line 470: Are we no longer encoding features with unique formatting?

yoyodyne/datasets.py
```diff
@@ -532,26 +518,49 @@ def __getitem__(
...
+        Returns:
+            torch.Tensor: the encoded tensor.
+        """
+        sequence = []
+        for feature in features:
+            if feature in self.source_symbol2i:
+                sequence.append(self.source_symbol2i[feature])
+            else:
+                raise Error(
+                    f"Feature {feature!r} seen during inference was not "
```
bonham79, line 551: This ignores cases of unknown features. Did we revert the decision from the previous discussion?

yoyodyne/datasets.py
```diff
@@ -599,16 +609,11 @@ def decode_features(
         """
         # Masking source vocab.
         indices = torch.where(
-            (indices >= self.features_idx) | (indices < len(self.special_idx)),
+            (indices >= self.features_idx) | (indices == self.pad_idx),
```
bonham79, line 612: Technically the second boolean check is unnecessary. It saves a few flops but don't think it's major in the scheme of things.

bonham79 (approved):
Everything looks good to go. Confused about how we decided to manage feature logic though. To my memory:
1. We were using a unique formatting for features to prevent overlap with source codes
2. We wanted to allow features not seen during training. Did we choose to revert this decision?

yoyodyne/datasets.py
kylebgorman: It's already logged in trainer, which is why I got rid of it.

yoyodyne/datasets.py
kylebgorman: This is a merge error of course, fixed.

yoyodyne/datasets.py
kylebgorman: Merge error, fixed.

yoyodyne/datasets.py
kylebgorman: Good call, removed it.

kylebgorman:
No to your question, this isn't changing anything about feature handling, there are just issues in merging...too many PRs out at once.

PR #25, opened 2023-01-06
Migrates to argparse

kylebgorman:
This PR moves the command-line parsing to the built-in `argparse`. This:

* allows us to take advantage of PL's ability to add every option the trainer supports to the argparse parser
* allows us to mostly-automatically build the trainer from these args

Among the cool things this enables: 

* one can set a max wall clock time for training
* one can control training time with (min, max) (epochs, steps) or combinations thereof

And, we no longer need to "mock" trainer args except those we want to keep fixed (which here includes the loggers and callbacks but nothing else), so we get things like gradient clipping (or rather a set of different gradient clipping algorithms and threshold values), numerical precision, batch norm, auto-LR tuning, multi-accelerator training, and so on for free.

Note that for compatibility with the automatically generated flags I am now using `--google_style` instead of `--unix-style`. That's just what Lightning does and I thought the inconsistency would be maddening.

Allow me to draw your attention to the files most affected:

* `train.py`: gets substantial new modularity. This will be used for #5, #15, and #16.
* `predict.py`: ditto
* `base.py`: has the attributes all models share inside of it now. This reduces boilerplate.

To do this well, I felt it necessary to add some substantial modularity to `train.py` and `predict.py` itself. This is thus a minor step towards #5, #15, and #16. In particular I have used getter functions for all the major steps of the process as well as created reusable `train` and `predict` functions. 

Since the parsing of CLI arguments is complicated, I have added helper functions (including class methods where feasible) for this. This just serves to hide complexity. On the other hand, I have also added some functions and static methods to augment the argument parser with appropriate arguments. These have been moved into the highest point in the class hierarchy where said arguments are introduced. I hope that by placing these definitions closer to their use they will be easier to maintain. While this pattern may look weird, it's not uncommon. It also hides a bit of complexity.

This is largely performance-neutral and has been tested on Polish and Maori as before.

This introduces two minor feature regressions:

1. I am not exposing beam width as a flag at present. I take it as obvious that we probably shouldn't until we have good support (and testing) of this for all models. It adds a lot of complexity.
2. I am not exposing dev prediction. It occurs to me that this doesn't really need to live in training anyways---one can just repeat the call to the predictor on dev and test.


Adamits:
I was wondering why I saw a branch floating around changing out `click` for `argparse`. This sounds extremely helpful! I will take a look later today.

kylebgorman:
I took it a bit further than I maybe should and did a bunch of other things on this branch, which make make it hard to review. If you have problems/questions let me know.

yoyodyne/predict.py
```diff
@@ -1,50 +1,142 @@
...
+    Returns:
+        pl.Trainer.
+    """
+    return pl.Trainer(max_epochs=0, **kwargs)
+
+
+def _get_trainer_from_argparse_args(
+    args: argparse.Namespace,
+) -> pl.Trainer:
+    """Creates the trainer from CLI arguments."""
```
Adamits, line 28: Should we add `Args`/`Returns`?

yoyodyne/predict.py
```diff
@@ -1,50 +1,142 @@
...
+        index (str).
+
+    Returns:
+        data.Dataset.
+    """
+    # TODO: Since we don't care about the target column, we should be able to
+    # set config.source_col = 0 and avoid the overhead for parsing it.
+    # This does not work because the modules expect it to be present even if
+    # they ignore it.
+    dataset = datasets.get_dataset(predict, config)
```
Adamits, line 51: This still loads an initial index since we don't have a different dataset to pass, right? I think if we could pass the index (in this case the index file it seems) to the Dataset this could be avoided (future PR?)

yoyodyne/predict.py
```diff
@@ -1,50 +1,142 @@
...
+def _get_loader_from_argparse_args(
+    args: argparse.Namespace,
+) -> data.DataLoader:
+    """Creates the loader from CLI arguments."""
+    config = dataconfig.DataConfig.from_argparse_args(args)
+    dataset = get_dataset(args.predict, config, args.index)
+    return get_loader(dataset, args.arch, args.batch_size)
+
+
+def get_model(
```
Adamits, line 89: What is this for?

yoyodyne/predict.py
```diff
@@ -1,50 +1,142 @@
...
 
 from . import collators, dataconfig, datasets, models, util
 
 
-def write_predictions(
-    model: models.base.BaseEncoderDecoder,
-    loader: torch.utils.data.DataLoader,
-    output: str,
-    accelerator: str,
+def get_trainer(**kwargs) -> pl.Trainer:
```
Adamits, line 13: What is this for?

yoyodyne/predict.py
```diff
@@ -66,119 +155,51 @@ def write_predictions(
...
-    dataset_cls = datasets.get_dataset_cls(config.has_features)
-    dataset = dataset_cls(predict, config)
-    dataset.load_index(model_dir, experiment)
-    util.log_info(f"Source vocabulary: {dataset.source_symbol2i}")
-    util.log_info(f"Target vocabulary: {dataset.target_symbol2i}")
-    collator_cls = collators.get_collator_cls(
-        arch, config.has_features, include_targets=False
+    # Path arguments.
+    parser.add_argument(
+        "--predict",
```
Adamits, line 168: Can this be called `--predict_path` or `--prediction_path`? `--predict` sounds like it'd be a boolean flag to me :D

yoyodyne/symbols.py
```diff
@@ -0,0 +1,39 @@
...
+    symbol2index: Dict[str, int]
+
+    def __init__(self, symbols: Iterable[str]):
+        self._index2symbol = list(symbols)
+        self._symbol2index = {c: i for i, c in enumerate(self._index2symbol)}
+
+    def __len__(self) -> int:
+        return len(self._index2symbol)
+
+    def index(self, symbol: str, unk_idx: Optional[int] = None) -> int:
```
Adamits, line 29: Args/Returns?

yoyodyne/symbols.py
```diff
@@ -0,0 +1,39 @@
...
+        """Looks up index by symbol."""
+        return self._symbol2index.get(symbol, unk_idx)
+
+    def symbol(self, index: int) -> str:
```
Adamits, line 33: Args/Returns?

yoyodyne/train.py
```diff
@@ -214,103 +68,352 @@ def main(
...
-        accelerator=accelerator,
-        logger=logger,
-        max_epochs=max_epochs,
-        gradient_clip_val=gradient_clip,
-        check_val_every_n_epoch=eval_every,
-        enable_checkpointing=True,
+    return trainer_callbacks
+
+
+def get_trainer(
```
Adamits, line 74: What is this for?

yoyodyne/train.py
```diff
@@ -214,103 +68,352 @@ def main(
...
+
+def main() -> None:
+    """Trainer."""
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument(
+        "--experiment", required=True, help="Name of experiment"
+    )
+    # Path arguments.
+    parser.add_argument(
+        "--train",
```
Adamits, line 335: I feel the same about this as `--predict` in the other file. But as long as they are consistent I guess it seems fine!

Adamits (commented):
The argparse stuff is really awesome!

You can ignore my comments about PaddedTensor/PaddedBatch. I have now seen them in use and its definitely an improvement.

I want to confirm, does `--help` still show all possible arguments to the cli? Thanks for all of this work, this is a huge overhaul!

yoyodyne/models/pointer_generator.py
```diff
@@ -43,29 +41,27 @@ def __init__(self, *args, **kwargs):
...
+        embedded = self.source_embeddings(source.padded)
         embedded = self.dropout_layer(embedded)
-        lens = (source_mask == 0).sum(dim=1).to("cpu")
+        lens = (source.mask == 0).sum(dim=1).to("cpu")
         packed = nn.utils.rnn.pack_padded_sequence(
             embedded, lens, batch_first=True, enforce_sorted=False
         )
         # -> B x seq_len x encoder_dim,
         # (D*layers x B x hidden_size, D*layers x B x hidden_size)
-        packed_outs, (H, C) = self.encoder(packed)
```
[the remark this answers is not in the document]
Adamits: It looks like in practice we passed `self.encoder` as `encoder`, so it was not a bug that will effect anything in the current state, luckily.

yoyodyne/models/expert.py
```diff
@@ -394,7 +395,7 @@ def find_prefixes(
 
 def get_expert(
     train_data: data.Dataset,
-    epochs: int = 10,
+    epochs: int = 5,
```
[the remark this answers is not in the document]
kylebgorman: It matches the command-line defaults, is all. I have found that it always converges by 5 iterations on everything I've tried it with.

yoyodyne/datasets.py
```diff
@@ -97,45 +89,97 @@ def _read_pkl(path: str) -> Any:
...
+    config: dataconfig.DataConfig
+    samples: List
+    source_map: symbols.SymbolMap
+    target_map: symbols.SymbolMap
+
+    def __init__(
+        self,
+        filename,
+        config,
+        other: Optional[BaseDataset] = None,
```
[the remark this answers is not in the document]
kylebgorman: Not a bad suggestion.

yoyodyne/batching.py
```diff
@@ -0,0 +1,103 @@
...
+appropriate device. Furthermore, we register the tensors as buffers.
+"""
+
+from typing import List, Optional
+
+import torch
+from torch import nn
+
+
+class PaddedTensor(nn.Module):
```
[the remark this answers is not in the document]
kylebgorman: Yeah. If you want, I could combine PaddedBatch and PaddedTensor into a single class.

yoyodyne/predict.py
kylebgorman: This and `get_trainer` are reasonably needed around if you want to use this tool not-from-the-CLI, I hypothesize. I may make small changes to them when I start writing deeper tests.

yoyodyne/predict.py
kylebgorman: Well, the tool is called predict so I figure... but okay, I can refactor around that.

kylebgorman:
> I want to confirm, does `--help` still show all possible arguments to the cli? Thanks for all of this work, this is a huge overhaul!

Yes, `-h` and `--help` both work.

yoyodyne/predict.py
kylebgorman: Oh great idea. I'll add a TODO.

yoyodyne/predict.py
Adamits: Good point I forgot about that use case ha

yoyodyne/datasets.py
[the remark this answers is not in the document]
kylebgorman: My solution was to make the index its own class. This was a huge simplification. Then, you can just pass the index. There is no longer a need for the training and dev set to communicate through the file system.

yoyodyne/models/pointer_generator.py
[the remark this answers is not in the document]
kylebgorman: Just investigated. What I did here is an improvement in clarity, though I can't say I remember doing it for that or any other reason. I have added a note to the method documentation to explain why we did it this way.

kylebgorman:
I believe I have now addressed everything. Making the indexes their own class was the last thing I needed to make this cohere.

Issue #26, opened 2023-01-08
Use specialized prediction?


kylebgorman:
Currently, our prediction code does [some relatively hairy stuff that is also architecture sensitive](https://github.com/CUNY-CL/yoyodyne/blob/52587bc1735b2112f1b001a41604fa8ca1d35caf/yoyodyne/predict.py#L50). Lightning permits one to define a [`predict_step` method](https://pytorch-lightning.readthedocs.io/en/stable/common/lightning_module.html#inference). If the user does not specify this, prediction is just the same as calling `forward`. 

It seems to me from this that we could likely move some of that code (the squeeze, transpose, and max) into `predict_step` for the appropriate modules, making it so that there's less Torchyness, and so we don't need to pass around or switch on the arch string.

bonham79:
Approve of porting more logic to `predict_step`. Having model specific parts of evaluation is a personal annoyance (also my fault...)

kylebgorman:
Cool, I'm going to assign you. It looks like some can go in the base and some in the transducer and that would cover it?

bonham79:
Should work. Will make some of life easier anyhow. Will get to it later in the week.

kylebgorman:
Closed in #44.

PR #27, opened 2023-01-08
Get out of the output formatting business

kylebgorman:
This rather simple PR takes the stance that the proper solution to https://github.com/CUNY-CL/yoyodyne/issues/26
is to simply not output the source or feature columns at all, which
simplifies the prediction loop considerably.

### DIFF
--- a/yoyodyne/predict.py
+++ b/yoyodyne/predict.py
@@ -1,6 +1,5 @@
 """Prediction."""
 
-import csv
 import os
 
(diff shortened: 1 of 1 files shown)


bonham79 (approved):
LGTM

kylebgorman:
I incorrectly labeled this as closing #26. Obviously it's relevant but that issue is still there.

PR #30, opened 2023-01-22
Removes --attention flag.

### LINKED ISSUE #29: Avoid --attention flag
One extra complexity for model class lookup is that we treat attention (but only for LSTMs) as a separate flag rather than as a separate architecture. This isn't mirrored in the backend: they are separate model classes.

I propose to call the vanilla LSTM `--arch lstm`, the attentive LSTM `--arch attentive_lstm`. Documentation and defaults will also need to be updated, as  will tests. Self-assigning.

kylebgorman:
This is now just part of the architecture specification, like it is underlyingly.

Closes #29.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,14 +99,15 @@ TPUs.
 The user specifies the model using the `--arch` flag (and in some cases
 additional flags).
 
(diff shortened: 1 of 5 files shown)


Adamits (approved):
LGTM

Issue #28, opened 2023-01-20
Linear decay scheduler


kylebgorman:
We only have one decaying scheduler right now: the goofy warmup+inverse square decay strategy from the original transformer paper. I propose that we also add a linear decay to `schedulers.py`.

@Adamits I think this is in your wheelhouse so I have assigned it to you; let me know if that's a problem.

Adamits:
Sounds good! I have a few things for this project i'll have time to do after the ACL deadline tonight!

Adamits:
I am looking at this now. I think we want to do this by adding pytorch schedulers to [this factory](https://github.com/CUNY-CL/yoyodyne/blob/3b1bf93c38dee004805912038c8ba6c40410f94b/yoyodyne/models/base.py#L300), and we only need to add to schedulers.py in the case that we want to wrap something in custom code.

Then, we can remove `warmup_steps` as an arg to the models, and just take a `schedulers_kwargs` dictionary. These will just be passed along to the schedulers, and we can even just defer to pytorch docs/exceptions for what should be passed through.

I am looking at the argparse docs now -- do they have a default way of grouping certain arguments in such a way that after parsing args, we get a specific dictionary of scheduler_kwargs? Does this implementation sound ok to you?

kylebgorman:
> I am looking at this now. I think we want to do this by adding pytorch schedulers to [this factory](https://github.com/CUNY-CL/yoyodyne/blob/3b1bf93c38dee004805912038c8ba6c40410f94b/yoyodyne/models/base.py#L300), and we only need to add to schedulers.py in the case that we want to wrap something in custom code.

Following so far.

> Then, we can remove `warmup_steps` as an arg to the models, and just take a `schedulers_kwargs` dictionary. These will just be passed along to the schedulers, and we can just defer to pytorch docs/exceptions for what should be passed through, though I guess we need to add arguments that differ (e.g. for linear decay we need a start factor and end factor) for every scheduler (same would be true for more optimizers), as we add them.

That would be a major improvement.

> I am looking at the argparse docs now -- do they have a default way of grouping certain arguments in such a way that after parsing args, we get a specific dictionary of scheduler_kwargs? Does this implementation sound ok to you @kylebgorman ? Or else, I can just make that dict after we parse the args.

[...]

PR #31, opened 2023-01-26
Adds LR decay scheduler.

Adamits:
- Adds schedulers.add_argparse_args
- Refactors _get_lr_scheduler

I added some other functions to schedulers, to support converting from yoyodyne argparse args, to the torch scheduler args. We may also want to consider exposing the interval and frequency of schedule updates to the CLI. In this way, the `total_decay_steps` for the decay scheduler could be total_decay_iters, and depend on whether you request that the scheduler potentially update every step or epoch.

Also -- when I run my local `black`, it seems to modify most files, so I reverted that and just used my flake8 config. Is there anywhere I can see the linting setup for this repo, and apply it to my local?

Addresses #28 28

### DIFF
--- a/yoyodyne/schedulers.py
+++ b/yoyodyne/schedulers.py
@@ -124,13 +124,17 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         "--start_factor",
         type=float,
         default=1 / 3,
-        help="The starting multiplier for the LR (lineardecay scheduler only). Default: %(default)s.",
+        help="The starting multiplier for the LR "
+        "(lineardecay scheduler only). "
+        "Default: %(default)s.",
     )
     parser.add_argument(
         "--end_factor",
         type=float,
         default=1.0,
-        help="The multiplier for the LR after total_decay_steps (lineardecay scheduler only). Default: %(default)s.",
+        help="The multiplier for the LR after total_decay_steps "
+        "(lineardecay scheduler only). "
+        "Default: %(default)s.",
     )
     parser.add_argument(
         "--total_decay_steps",
--- a/yoyodyne/schedulers.py
+++ b/yoyodyne/schedulers.py
@@ -34,6 +34,7 @@ class WarmupInverseSquareRootSchedule(optim.lr_scheduler.LambdaLR):
         optimizer: optim.Optimizer,
         warmup_steps,
         last_epoch=-1,
+        **kwargs,
     ):
         """Initializes the LR scheduler.
 
@@ -41,6 +42,7 @@ class WarmupInverseSquareRootSchedule(optim.lr_scheduler.LambdaLR):
             optimizer (optim.Optimizer): optimizer.
             warmup_steps (int): number of warmup steps.
             last_epoch (int): last epoch for the scheduler.
+            **kwargs: ignored.
         """
         self.warmup_steps = warmup_steps
         self.decay_factor = math.sqrt(warmup_steps)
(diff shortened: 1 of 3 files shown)


kylebgorman:
Thanks for this Adam.

Re: the `black` config stuff, the config it runs is [here](https://github.com/CUNY-CL/yoyodyne/blob/2e15a37e4666195fbcf33d75834a8d504e65d3d6/.circleci/config.yml#L51).

Re: this design, I wonder if it would be easier if you just wrapped the Torch thing and renamed things as you feel? That ought to get rid of at least `_map_scheduler_kwargs`?  Then we could make our scheduler constructors ignore `**kwargs` too (like `def __init__(arg_i_care_about: T, **kwargs): ...`), and we don't need `filter_scheduler_kwargs` either, I think. However it is extremely possible if I am missing something here.

I don't really understand how to use this either. I wonder if we can do more with clear help strings here. (Sadly the Torch source code is not very helpful.)

Adamits:
> Re: this design, I wonder if it would be easier if you just wrapped the Torch thing and renamed things as you feel? That ought to get rid of at least _map_scheduler_kwargs? Then we could make our scheduler constructors ignore **kwargs too (like def __init__(arg_i_care_about: T, **kwargs): ...), and we don't need filter_scheduler_kwargs either, I think. However it is extremely possible if I am missing something here.

No I think this does sound better. I will do it this way (probably wont get to it until tmrw).

> I don't really understand how to use this either. I wonder if we can do more with clear help strings here. (Sadly the Torch source code is not very helpful.)

Is this wrt torch schedulers?

kylebgorman:
> > Re: this design, I wonder if it would be easier if you just wrapped the Torch thing and renamed things as you feel? That ought to get rid of at least _map_scheduler_kwargs? Then we could make our scheduler constructors ignore **kwargs too (like def **init**(arg_i_care_about: T, **kwargs): ...), and we don't need filter_scheduler_kwargs either, I think. However it is extremely possible if I am missing something here.
> 
> No I think this does sound better. I will do it this way (probably wont get to it until tmrw).

Cool. I think this would just be straight inheritance---tweaking the `__init__` and that's it.

> Is this wrt torch schedulers?

Yeah. I wonder if we need some docs about how to use the two schedulers we've enabled. Maybe that's all that's necessary. I also was having trouble understanding the flag `help` but also the docs in the `torch.optim` module itself for this class,

yoyodyne/schedulers.py
```diff
@@ -64,6 +71,41 @@ def lr_lambda(self, step: int) -> float:
         return self.decay_factor * step**-0.5
 
 
+class LinearDecay(optim.lr_scheduler.LinearLR):
+    """Linear decay scheduler."""
+
+    optimizer: optim.Optimizer
```
kylebgorman, line 77: I think these type decls are no longer needed since the superclass can take care of it.

yoyodyne/schedulers.py
```diff
@@ -88,63 +130,28 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         "--start_factor",
         type=float,
         default=1 / 3,
-        help="Start factor for lr decay. Default: %(default)s.",
+        help="The starting multiplier for the lr in LinearDecay. Default: %(default)s.",
```
kylebgorman, line 133: suggest "LR"; also on lines 138 and 145.

Above, for  warmup steps, it reads "(warmupinvsqrt scheduler only)". Here it reads "in LinearDecay". I don't care which style we use (yours may be better actually) but we should standardize.

kylebgorman (approved):
I think this is a big improvement. A few notes, mostly about docs and style, but otherwise LGTM.

kylebgorman:
The internet says the issue with CircleCI will resolve if you log out of CircleCI and then log back in with GitHub credentials.

Adamits:
> The internet says the issue with CircleCI will resolve if you log out of CircleCI and then log back in with GitHub credentials.

I think it wasn't even set up in my fork. Have now done so (after logging in via github).

kylebgorman:
CircleCI stuff is working now (either because something I did or something you did, not sure). Can you fix these?

```
./yoyodyne/schedulers.py:6:1: F401 'inspect' imported but unused
./yoyodyne/schedulers.py:128:80: E501 line too long (102 > 79 characters)
./yoyodyne/schedulers.py:134:80: E501 line too long (117 > 79 characters)
```

May need to run `black` on that file too after the fact. Other than that LGTM again.

Adamits:
This is weird.

```
(yoyodyne) (base) adamwiemerslage ~/nlp-projects/morphology/yoyodyne $ black --line-length=79 yoyodyne/schedulers.py 
All done! ✨ 🍰 ✨
1 file left unchanged.
```

kylebgorman:
> This is weird.

It just doesn't break apart strings for you, as you see...wish it did.

PR #33, opened 2023-01-28
Cleanup to `super` calls in schedulers

kylebgorman:
This immediately follows #31. It simplifies the `super` calls in the two schedulers and then tests both against some of our data.


Adamits (approved):
LGTM

kylebgorman:
@Adamits FWIW what you did is was the only correct way under some older system, and is unambiguous. (It is still sometimes necessary if you have a class Foo which inherits (directly, not transitively) from both Bar and Baz since you have be explicit as to which supercalss method you want access to. It's just not necessary here, I found out.

PR #36, opened 2023-02-01
Removes train_set as a model arg

### LINKED ISSUE #32: hparams
When testing, I noticed two things:

1. hparams.yaml is continuously rebuilt while training -- maybe this means we we reinstantiate the model (to load from a checkpoint or something) several times? This could be ok, but struck me as weird.
2. All samples are now in the hparams file, so it is large and gets bigger as the dataset gets bigger. This is via the training dataset property on the model I believe. I think in the past I added something to ignore certain peroperties in `hparams.yaml` but don't remember what I did off the top of my head.

I need to look into both at some point. Putting this issue here as a placeholder for now.

Adamits:
closes #32

### DIFF
--- a/yoyodyne/train.py
+++ b/yoyodyne/train.py
@@ -295,7 +295,6 @@ def get_model(
         scheduler=scheduler,
         scheduler_kwargs=scheduler_kwargs,
         start_idx=train_set.index.start_idx,
-        train_set=train_set,
         vocab_size=train_set.index.source_vocab_size,
     )
 


kylebgorman (approved):
LGTM. Black wants you to reformat yoyodyne/models/base.py for some reason first.

kylebgorman:
It wants me to do that in my branch too, and it's a no-op so I'm confused. Going to ignore it,  you should too.

PR #39, opened 2023-02-01
Upgrading `black` to get beyond CircleCI issue

kylebgorman:
Okay, let's see if this works.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-black>=22.3.0
+black>=22.6.0
 build>=0.9.0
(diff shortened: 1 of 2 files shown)


kylebgorman:
Okay, this is bizarre. What the command returns on CircleCI's side does not match what I see locally:

```
[yoyodyne]$ black --line-length=79 --check .
All done! ✨ 🍰 ✨
26 files would be left unchanged.
```

kylebgorman:
Okay,  now I went to the most current version (since it looks like CircleCI is selecting that).

PR #42, opened 2023-02-02
Fixes mkdir issue in predict

### LINKED ISSUE #41: Outputting to the current directory
In `yoyodyne-predict`, the following will fail: `--output path_to_current_directory_file.txt`. Do you know why? It's because it tries to create the directory before writing. The simple solution, which I will now implement, will not call `os.makedir` if writing to the PWD. (We can just test for an empty string.)

Reported by @liliest

kylebgorman:
Closes #41.

### DIFF
--- a/yoyodyne/predict.py
+++ b/yoyodyne/predict.py
@@ -144,7 +144,7 @@ def _mkdir(output: str) -> None:
     Args:
         output (str): output to output file.
     """
-    dirname = os.output.dirname(output)
+    dirname = os.path.dirname(output)
     if dirname:
         os.makedirs(dirname, exist_ok=True)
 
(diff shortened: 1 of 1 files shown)


yoyodyne/predict.py
```diff
@@ -138,6 +138,17 @@ def _get_model_from_argparse_args(
     return model_cls.load_from_checkpoint(args.checkpoint)
 
 
+def _mkdir(output: str) -> None:
+    """Creates directory for output file if necessary.
+
+    Args:
+        output (str): output to output file.
+    """
+    dirname = os.output.dirname(output)
```
Adamits, line 147: `os.output` --> `os.path`, no?

Adamits (changes requested):
I think the submodule name in os is wrong. Otherwise LGTM

Issue #40, opened 2023-02-02
Sequence length errors


kylebgorman:
This scary error sometimes greets you if you are using the transformer:

```
../aten/src/ATen/native/cuda/IndexKernel.cu:92: operator(): block: [398,0,0], thread: [51,0,0] Assertion `index >= -sizes[i] && index < sizes[i] && "index out of bounds"` failed.
```

I believe this is what happens when you have inputs that exceed `--max_sequence_length` (increasing it to a huge number makes it go away). What should we do about this? A few possibilities:

1. Cowardly refuse to pad a source sequence batch longer than `--max_sequence_length` by throwing an exception. This would have the interesting effect that it would apply not just to transformer anymore (which actually might be desirable---certainly it's clearer).
2. Document in the README that that scary errors means you need to increase `--max_sequence_length`.

I think I prefer (1). What do you think?

Relatedly, it occurs that the following might be clearer:

* `--max_sequence_length` > `--max_source_length`; help docs should also mention that this is for transformers only, unless we adopt (1) in which case it becomes uniform; we should also move this into base.py
* `--max_decode_length` > `--max_target_length`

Adamits:
Does this raise an exception? Or is that message a warning? I believe that is just used to create the positional encoding tensor.

Anyway I also prefer 1). Though we could maybe remove it as an argument all together and just compute it dynamically before initializing the positional encoding. I would need to look into the implications of getting potentially huge positional encodings though.

Adamits:
I **think** setting a `--max_sequence_length` is largely inspired by pretrained transformers that set such limits. This is probably for efficiency reasons, and during pretraining they just chunk documents to be within the threshold. However, I also imagine there is a point at which encoding performance also degrades if a sequence is too long. I don't really know the implications of this for character transduction.

I think if we keep this argument and raise an exception as you say, then that makes the choice of how to handle long sequences here explicit and up to the user.

kylebgorman:
> Does this raise an exception? Or is that message a warning? I believe that is just used to create the positional encoding tensor.

It's a warning, but I think it should be raised to an exception, since it's a major issue.

> Anyway I also prefer 1). Though we could maybe remove it as an argument all together and just compute it dynamically before initializing the positional encoding. I would need to look into the implications of getting potentially huge positional encodings though.

That could be done later, maybe. I don't know if you can actually know how long the positional encoding needs to be at the time we're constructing the transformer model instance. 

[...]

PR #43, opened 2023-02-03
Rename length flags

kylebgorman:
* `--max_sequence_length` -> `--max_source_length`
* `--max_decode_length` -> `--max_target_length`

This addresses commentary in #40. However it leaves open other parts of the bug, namely that it does not add documentation nor raise exceptions when the max lengths are exceeded by training data.


Adamits (approved):
LGTM

PR #44, opened 2023-02-05
Predict upgrade

bonham79:
For https://github.com/CUNY-CL/yoyodyne/issues/26

Major changes:

-- base encoder now contains predict_step that applies to lstm and transformer models.
--Put wrapper around base_encoder.__get_loss method (explained below)

@Adamits  @kylebgorman  This is a minor fix but want to get y'alls thoughts on best practices: While going through this, it looks like we were using two logics for forward passes. For training, tensors were assumed to be of shape `B x vocab x sequence_len` while on prediction they were assumed as `B x sequence_len x vocab` . This isn't ideal and led to a bunch of transpose steps that were somewhat unnecessary. I've tried to unify them under the assumption that alll `forward` passes predict a tensor of shape `B x sequence_len x vocab`

Channeling my innner Guido I want to suggest the following:
1. All `forward` passes (unless in specialized cases), produce tensors of one unified shape. My preference is for `B x seq_len x vocab_size` but am open to other arguments.
2. `base_encoder._get_predictions` should only be called on decoder outputs. Since this is really only a wrapper for `max`, I think it's more important to save semantic sanity than using it for condensing code. 

Anyway, these are just my thoughts.

### DIFF
--- a/yoyodyne/indexes.py
+++ b/yoyodyne/indexes.py
@@ -1,7 +1,7 @@
 """Indexes for symbols."""
 
-import pickle
 import os
+import pickle
 from typing import Dict, List, Optional, Set
 
 from . import special
--- a/yoyodyne/indexes.py
+++ b/yoyodyne/indexes.py
@@ -87,7 +87,9 @@ class IndexNoFeatures(BaseIndex):
     target_map: SymbolMap
 
     def __init__(
-        self, source_vocabulary: List[str], target_vocabulary: List[str],
+        self,
+        source_vocabulary: List[str],
+        target_vocabulary: List[str],
     ):
         """Initializes the index.
 
--- a/yoyodyne/indexes.py
+++ b/yoyodyne/indexes.py
@@ -1,7 +1,6 @@
 """Indexes for symbols."""
-
-import os
 import pickle
+import os
 from typing import Dict, List, Optional, Set
 
 from . import special
@@ -88,9 +87,7 @@ class IndexNoFeatures(BaseIndex):
     target_map: SymbolMap
 
     def __init__(
-        self,
-        source_vocabulary: List[str],
-        target_vocabulary: List[str],
+        self, source_vocabulary: List[str], target_vocabulary: List[str],
     ):
         """Initializes the index.
 
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -112,7 +112,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         embedding_layer = nn.Embedding(num_embeddings, embedding_size)
         # Xavier initialization.
         nn.init.normal_(
-            embedding_layer.weight, mean=0, std=embedding_size ** -0.5
+            embedding_layer.weight, mean=0, std=embedding_size**-0.5
         )
         # Zeroes out pad embeddings.
         if pad_idx is not None:
@@ -161,7 +161,9 @@ class BaseEncoderDecoder(pl.LightningModule):
         raise NotImplementedError
 
     def training_step(
-        self, batch: batches.PaddedBatch, batch_idx: int,
+        self,
+        batch: batches.PaddedBatch,
+        batch_idx: int,
     ) -> torch.Tensor:
         """Runs one step of training.
 
@@ -188,7 +190,9 @@ class BaseEncoderDecoder(pl.LightningModule):
         return loss
 
     def validation_step(
-        self, batch: batches.PaddedBatch, batch_idx: int,
+        self,
+        batch: batches.PaddedBatch,
+        batch_idx: int,
     ) -> Dict:
         """Runs one validation step.
 
@@ -232,7 +236,9 @@ class BaseEncoderDecoder(pl.LightningModule):
         return metrics
 
     def predict_step(
-        self, batch: batches.PaddedBatch, batch_idx: int,
+        self,
+        batch: batches.PaddedBatch,
+        batch_idx: int,
     ) -> torch.Tensor:
         """Runs one predict step.
 
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -235,6 +235,28 @@ class BaseEncoderDecoder(pl.LightningModule):
             self.log(metric, value, prog_bar=True)
         return metrics
 
+    def predict_step(
+        self,
+        batch: batches.PaddedBatch,
+        batch_idx: int,
+    ) -> torch.Tensor:
+        """Runs one predict step.
+
+        This is called by the PL Trainer.
+
+        Args:
+            batch (batches.PaddedBatch).
+            batch_idx (int).
+
+        Returns:
+            torch.Tensor: indices of the argmax at each timestep.
+        """
+        self.eval()
+        predictions = self(batch)
+        # Default tensor: B x seq_len x vocab_size.
+        greedy_predictions = self._get_predicted(predictions)
+        return greedy_predictions
+
     def _get_predicted(self, predictions: torch.Tensor) -> torch.Tensor:
         """Picks the best index from the vocabulary.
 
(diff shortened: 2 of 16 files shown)


kylebgorman:
> @Adamits @kylebgorman This is a minor fix but want to get y'alls thoughts on best practices: While going through this, it looks like we were using two logics for forward passes. For training, tensors were assumed to be of shape `B x vocab x sequence_len` while on prediction they were assumed as `B x sequence_len x vocab` . This isn't ideal and led to a bunch of transpose steps that were somewhat unnecessary. I've tried to unify them under the assumption that alll `forward` passes predict a tensor of shape `B x sequence_len x vocab`

Oof, is that so? Glad you fixed. 

> Channeling my innner Guido I want to suggest the following:
> 
> 1. All `forward` passes (unless in specialized cases), produce tensors of one unified shape. My preference is for `B x seq_len x vocab_size` but am open to other arguments.

No preference so long as it's consistent and batch size is the leading dimension.

yoyodyne/indexes.py
```diff
@@ -1,7 +1,7 @@
 """Indexes for symbols."""
 
-import pickle
 import os
+import pickle
```
kylebgorman, line 4: please revert

yoyodyne/models/base.py
```diff
@@ -235,6 +235,28 @@ def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:
...
+        Args:
+            batch (batches.PaddedBatch).
+            batch_idx (int).
+
+        Returns:
+            torch.Tensor: indices of the argmax at each timestep.
+        """
+        self.eval()
+        predictions = self(batch)
+        # Default tensor: B x seq_len x vocab_size.
```
kylebgorman, line 256: What does "Default tensor" mean? Is this just saying the next step returns something of shape B x seq_len x vocab_size? If so then use Adam's notation:

`# -> B x seq_len x vocab_size.`

yoyodyne/models/transducer.py
```diff
@@ -45,20 +45,21 @@ def __init__(
...
-        # Alternate outputs than dataset targets.
-        kwargs["output_size"] = len(expert.actions)
-        super().__init__(*args, **kwargs)
         # Target embeddings use alternate padding.
         self.target_embeddings = self.init_embeddings(
             num_embeddings=self.output_size,
             embedding_size=self.embedding_size,
             pad_idx=self.actions.end_idx,
         )
+        self.loss_func = TransducerNoFeatures.loss_func
```
kylebgorman, line 62: This requires documentation. You're trying to say that you're going to overload it later? Then this should be no-op. Or are you trying to say something else?

kylebgorman (commented):
This is nice for getting rid of the stuff in predict! 

A few nits/comments (see also main thread). 

Can you standardize on either `seq_len` or (and I prefer this, particularly when these are just comments anyways) `sequence_length` for

kylebgorman:
> > 1. All `forward` passes (unless in specialized cases), produce tensors of one unified shape. My preference is for `B x seq_len x vocab_size` but am open to other arguments.
> 
> No preference so long as it's consistent and batch size is the leading dimension.

Revising my comment after reviewing. If the loss function wants it one way, we should just do it that way everywhere unless there's some other built-in that wants it transposed.

yoyodyne/models/transducer.py
bonham79: I'm having a hard time making this non-obtrusive. The transducer forward pass computes a loss with it, so its loss function is just a dummy wrapper for referencing a tuple. Current inheritance order wants to call `base`'s loss function finder so I'm trying to override that function.

bonham79:
> Revising my comment after reviewing. If the loss function wants it one way, we should just do it that way everywhere unless there's some other built-in that wants it transposed.

That can be done, but many of the forward passes have a default of `B x seq_len x vocab_size`, so doing the other way adds in quite a few `transpose` operations. It also means `base._get_predictions` is more limited in when it can be called (to make semantic sense, it would have to operate on a different dimension, so can't be called with some of the `lstm` outputs until bottom of the forward pass. 

`transpose` is a cheap operation iirc, but I personally find too many of them hard to follow on a code edit. If this is just a me problem I'm fine with the other way.

yoyodyne/models/transducer.py
kylebgorman: Okay, well, you need to write a clear, sentential comment when you use a strange pattern: explain what the pattern does functionally, and why you want to do it.

Secondly, my comments otherwise suggest: we should structure batches to follow what the loss function wants and do away with transposition altogether.

kylebgorman:
> That can be done, but many of the forward passes have a default of `B x seq_len x vocab_size`, 

Can't we just change this too? If so we should. 

> `transpose` is a cheap operation iirc, but I personally find too many of them hard to follow on a code edit. If this is just a me problem I'm fine with the other way.

No you're right, they're awful.

bonham79:
> 
> Can't we just change this too? If so we should.
> 
Nah, this is a product of the pytorch operations. Particularly how lstms have a weird output behavior. Adam did a good job and still you have weirdness like `tensor.tranpose(0,1).tranpose(1,2)`

yoyodyne/indexes.py
bonham79: line 65 `dictionary = pickle.load(source)` line 77 `pickle.dump(vars(self), sink)` ?

kylebgorman: No, lines 2, and 4. You changed the order of imports (pickle was already being imported) to be non-lexicographic.

bonham79: oh gotcha, isort was overzealous

bonham79:
Ignore changes to unrelated files, my linters are having issues for some reason

kylebgorman (approved):
LGTM. @Adamits do you want to take a look?

yoyodyne/models/base.py
```diff
@@ -235,11 +235,33 @@ def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:
...
+        Args:
+            batch (batches.PaddedBatch).
+            batch_idx (int).
+
+        Returns:
+            torch.Tensor: indices of the argmax at each timestep.
+        """
+        self.eval()
+        predictions = self(batch)
+        # -> B x seq_len x output_size.
```
Adamits, line 256: Are you sure this doesn't return `B x seq_len x 1`?

Adamits:
Started reviewing and realized that I am pretty sure there is overlap with #46, which I reviewed first. Should I look at all commits here?

yoyodyne/models/lstm.py
```diff
@@ -339,8 +339,8 @@ def beam_decode(
                 char_likelihoods,
                 decoder_hiddens,
             ) in likelihoods:
-                # -> B x seq_len x outputs.
-                # This is 1 x 1 x outputs since we fixed batch size to 1.
+                # -> B x seq_len x output_size.
```
Adamits, line 342: Maybe just remove this comment.
The one below (sans "this is") should be sufficient.

kylebgorman:
#46 is just a superset of this, I believe; it has one more commit than this one (#44).

yoyodyne/models/transducer.py
```diff
@@ -488,19 +488,16 @@ def log_sum_softmax_loss(
...
 
         This model already produces prediction strings and calculates loss
-        during training. This simply serves as a wrapper to allow compatibility
-        with training_step in the superclass.
+        during training. This simply serves as a wrapper to prevent
+        superclass from creating an unused loss function.
         """
-        _, loss = prediction
-        return loss
+        return lambda predictions, target: predictions[1]
```
Adamits, line 500: Not sure how this is usually handled, but it strikes me as a bit odd that this is conceptually returns a loss function, but we have no docstring for it. If I am a developer making changes to `Transducer` is it clear how to interact with the loss function?

yoyodyne/models/transformer.py
```diff
@@ -212,17 +212,17 @@ def _decode_greedy(
             # We only care about the last prediction in the sequence.
             last_output = output[:, -1, :]
             outputs.append(last_output)
-            pred = self._get_predicted(last_output.unsqueeze(1))
-            predictions.append(pred.squeeze(1))
+            _, pred = torch.max(last_output, dim=1)
```
Adamits, line 215: A comment could be good here:
```
# -> B x 1 x 1
```

yoyodyne/models/base.py
```diff
@@ -323,7 +345,31 @@ def get_loss_func(
                 loss function.
         """
         if self.label_smoothing is None:
-            return nn.NLLLoss(ignore_index=self.pad_idx, reduction=reduction)
+            loss_func = nn.NLLLoss(
+                ignore_index=self.pad_idx, reduction=reduction
+            )
+
+            def _transpose_nlloss(
```
Adamits, line 352: This is all a bit weird, but pretty sure it's updated in the next PR!

yoyodyne/models/base.py
bonham79: Oh, mb.

yoyodyne/models/transducer.py
bonham79: This should be more coherrent in the other PR. Lemme know if you still see issues. Dev work should never touch the loss function since it's part o the forward pass.

bonham79:
Adams recs are done in the other PR to avoid commit conflicts.

kylebgorman:
> Adams recs are done in the other PR to avoid commit conflicts.

So does that mean this is ready?

PR #46, opened 2023-02-05
Multiprocess

bonham79:
Multiprocessor fix for training loop. This should work on multi GPU setup. I confirmed on a 1080 + 3080 (yes, I'm also amazed they're compatible). This was done by moving the inner functions from `get_loss_func` to being top level functions. (Required for pickle.)  Also made namespace tweaks to be more PTL agreeable.

Thing to note: I moved transpose behavior into training and validation loops. See if this is agreeable or we'd prefer to do the change at model level.  (Issue discussed in https://github.com/CUNY-CL/yoyodyne/pull/44)

### DIFF
--- a/yoyodyne/models/transducer.py
+++ b/yoyodyne/models/transducer.py
@@ -1,7 +1,7 @@
 """Transducer model class."""
 
 import math
-from typing import Dict, List, Optional, Tuple
+from typing import Dict, List, Optional, Tuple, Callable
 
 import numpy
 import torch
@@ -33,10 +33,7 @@ class TransducerNoFeatures(lstm.LSTMEncoderDecoder):
     target_embeddings: nn.Embedding
 
     def __init__(
-        self,
-        expert,
-        *args,
-        **kwargs,
+        self, expert, *args, **kwargs,
     ):
         """Initializes transducer model.
 
@@ -59,7 +56,6 @@ class TransducerNoFeatures(lstm.LSTMEncoderDecoder):
             embedding_size=self.embedding_size,
             pad_idx=self.actions.end_idx,
         )
-        self.loss_func = TransducerNoFeatures.loss_func
 
     def forward(
         self, batch: batches.PaddedBatch
@@ -152,10 +148,7 @@ class TransducerNoFeatures(lstm.LSTMEncoderDecoder):
             )
             # Decoding.
             logits, last_hiddens = self.decode_step(
-                encoder_out,
-                last_action,
-                last_hiddens,
-                alignment,
+                encoder_out, last_action, last_hiddens, alignment,
             )
             # If given targets, asks expert for optimal actions.
             optim_action = (
@@ -239,7 +232,7 @@ class TransducerNoFeatures(lstm.LSTMEncoderDecoder):
 
         Args:
             logits (torch.Tensor): logit values from decode_step of shape
-                B x output_size.
+                B x num_actions.
             alignment (torch.Tensor): index of encoding symbols for decoding,
                 per item in batch of shape B x seq_len.
             input_length (torch.Tensor): length of each item in batch.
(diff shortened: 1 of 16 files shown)


yoyodyne/models/transducer.py
```diff
@@ -488,16 +488,38 @@ def log_sum_softmax_loss(
...
+        Args:
+            batch (batches.PaddedBatch)
+            batch_idx (int).
+
+        Returns:
+            torch.Tensor: loss.
         """
-        return lambda predictions, target: predictions[1]
+        # -> B x seq_len x output_size.
+        predictions = self(batch)
```
kylebgorman, line 512: May I suggest: 

```python
_, loss = self(batch)
```

would be the clearest possible, since it is a surprise on line 514 to learn that something you call predictions has the loss in its second element.

kylebgorman (approved):
I reviewed this by reading the last (of 11) commits, bc4be28fd40548d5011656bc4c815aa6d71dfe52. This seems like a straightforward improvement. (I also like that we have one fewer subdefinition, just in general.) LGTM.

Can I ask about the name change from `self.loss_func` to just `self.loss`? I think the old one was better but there may be some justification I'm missing.

bonham79:
> Can I ask about the name change from `self.loss_func` to just `self.loss`? I think the old one was better but there may be some justification I'm missing.

Self.loss is more common in lightningese. That's all.

bonham79:
> Can I ask about the name change from `self.loss_func` to just `self.loss`? I think the old one was better but there may be some justification I'm missing.

Self.loss is more common in lightningese. That's all.

bonham79:
> I reviewed this by reading the last (of 11) commits, [bc4be28](https://github.com/CUNY-CL/yoyodyne/commit/bc4be28fd40548d5011656bc4c815aa6d71dfe52). This seems like a straightforward improvement. (I also like that we have one fewer subdefinition, just in general.) LGTM.
> 
Sorry bout that!  There was a version of 'black' in my 'bin' that decided to take supremacy over my env....

kylebgorman:
So this is fine but I'm going to wait until the earlier PR is approved by @Adamits before I merge.

Adamits (approved):
I forgot about the inner function multiprocessing issue! Good catch!

I think moving the transposes is an improvement. I can't really evaluate if the transpose changes are correct without running this -- but it sounds like you tested all of this in both training and prediction?

I also agree with Kyle that I like `loss_func` more, is there a doc anywhere demonstrating that `loss` should be used for the name of the function? Or is it just a general thing you've picked up from working on other PTL codebases? Either way I am fine with it if you are pretty sure it's justified :D.

LGTM

kylebgorman:
okay @bonham79 should we merge this?

(I may revert `loss_func` later.)

bonham79:
> I forgot about the inner function multiprocessing issue! Good catch!
> 
> I think moving the transposes is an improvement. I can't really evaluate if the transpose changes are correct without running this -- but it sounds like you tested all of this in both training and prediction?
> 

I did some quick testing on the english dataset and there were no flags set off by predict. 

> I also agree with Kyle that I like `loss_func` more, is there a doc anywhere demonstrating that `loss` should be used for the name of the function? Or is it just a general thing you've picked up from working on other PTL codebases? Either way I am fine with it if you are pretty sure it's justified :D.

Just something I picked up from other codebases. I don't think it violates any style rules, just strikes me as odd. I'm fine with reversion if the tribunal is 2-to-1.

bonham79:
> okay @bonham79 should we merge this?
> 
> (I may revert `loss_func` later.)

@Adamits Are there any issues in the other PR that aren't caught by this one?

Adamits:
> > okay @bonham79 should we merge this?
> > (I may revert `loss_func` later.)
> 
> @Adamits Are there any issues in the other PR that aren't caught by this one?

I don't think so, thanks!

bonham79:
@kylebgorman Commit away!

PR #49, opened 2023-02-08
--precision docs in README

kylebgorman:


Adamits (approved):
Didn't realize we had this option (I guess it's probably part of the PTL args?)!  

LGTM

PR #51, opened 2023-02-09
Adds Exception for inputs > maximum_source_length

Adamits:
- Adds warning for maximum_target_length.
- Adds log_warning function
- Addresses #40

### DIFF
--- a/tests/collator_test.py
+++ b/tests/collator_test.py
@@ -4,30 +4,42 @@ from yoyodyne import collators, dataconfig
 
 
 @pytest.mark.parametrize(
-    ["arch", "has_features", "has_target", "expected_separate_features"],
     [
-        ("feature_invariant_transformer", True, True, False),
-        ("feature_invariant_transformer", True, False, False),
-        ("lstm", True, True, False),
-        ("lstm", False, True, False),
-        ("lstm", True, False, False),
-        ("lstm", False, False, False),
-        ("pointer_generator_lstm", True, True, True),
-        ("pointer_generator_lstm", False, True, False),
-        ("pointer_generator_lstm", True, False, True),
-        ("pointer_generator_lstm", False, False, False),
-        ("transducer", True, True, True),
-        ("transducer", False, True, False),
-        ("transducer", True, False, True),
-        ("transducer", False, False, False),
-        ("transformer", True, True, False),
-        ("transformer", False, True, False),
-        ("transformer", True, False, False),
-        ("transformer", False, False, False),
+        "arch",
+        "has_features",
+        "has_target",
+        "expected_separate_features",
+        "max_src_length",
+        "max_tgt_length",
+    ],
+    [
+        ("feature_invariant_transformer", True, True, False, 128, 128),
+        ("feature_invariant_transformer", True, False, False, 128, 128),
+        ("lstm", True, True, False, 128, 128),
+        ("lstm", False, True, False, 128, 128),
+        ("lstm", True, False, False, 128, 128),
+        ("lstm", False, False, False, 128, 128),
+        ("pointer_generator_lstm", True, True, True, 128, 128),
+        ("pointer_generator_lstm", False, True, False, 128, 128),
+        ("pointer_generator_lstm", True, False, True, 128, 128),
+        ("pointer_generator_lstm", False, False, False, 128, 128),
+        ("transducer", True, True, True, 128, 128),
+        ("transducer", False, True, False, 128, 128),
+        ("transducer", True, False, True, 128, 128),
+        ("transducer", False, False, False, 128, 128),
+        ("transformer", True, True, False, 128, 128),
+        ("transformer", False, True, False, 128, 128),
+        ("transformer", True, False, False, 128, 128),
+        ("transformer", False, False, False, 128, 128),
     ],
 )
 def test_get_collator(
-    arch, has_features, has_target, expected_separate_features
+    arch,
+    has_features,
+    has_target,
+    expected_separate_features,
+    max_src_length,
+    max_tgt_length,
 ):
     config = dataconfig.DataConfig(
         features_col=3 if has_features else 0,
@@ -37,6 +49,8 @@ def test_get_collator(
         1,  # pad_idx, but it doesn't matter here.
         config,
         arch,
+        max_src_length,
+        max_tgt_length,
     )
     assert collator.has_target == has_target
     assert collator.separate_features == expected_separate_features
--- a/yoyodyne/collators.py
+++ b/yoyodyne/collators.py
@@ -4,7 +4,11 @@ from typing import List
 
 import torch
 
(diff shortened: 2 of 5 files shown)


yoyodyne/collators.py
```diff
@@ -16,23 +20,70 @@ class Collator:
...
+        """Checks if a source length in the batch is greater than the
+        maximum allowed source length.
+
+        Args:
+            itemlist (List[datasets.Item]).
+
+        Raises:
+            LengthError.
+        """
+        batch_max_source_length = max([len(item.source) for item in itemlist])
```
kylebgorman, line 63: The square brackets here are otiose:

`... = max(len(item.source) for item in itemlist)`

yoyodyne/collators.py
```diff
@@ -16,23 +20,70 @@ class Collator:
...
+
+    def _check_target_length(self, itemlist: List[datasets.Item]):
+        """Checks if a target length in the batch is greater than the
+        maximum allowed target length for inference. Logs a warning suggesting
+        that this probably indicates the max_target_length is too short.
+
+        Args:
+            itemlist (List[datasets.Item]).
+        """
+        batch_max_target_length = max([len(item.target) for item in itemlist])
```
kylebgorman, line 78: ditto

yoyodyne/train.py
```diff
@@ -424,7 +433,12 @@ def main() -> None:
     train_set.index.write(index)
     util.log_info(f"Index: {index}")
     train_loader, dev_loader = get_loaders(
-        train_set, dev_set, args.arch, args.batch_size
+        train_set,
+        dev_set,
+        args.arch,
+        args.batch_size,
```
kylebgorman, line 439: I think it may be time to write `_get_loaders_from_args` or whatever, and have that call `get_loaders`.

yoyodyne/collators.py
```diff
@@ -16,23 +20,70 @@ class Collator:
...
+    max_source_length: int
+    max_target_length: int
 
-    def __init__(self, pad_idx, config: dataconfig.DataConfig, arch: str):
+    def __init__(
+        self,
+        pad_idx,
+        config: dataconfig.DataConfig,
+        arch: str,
+        max_source_length: int,
```
kylebgorman, line 31: Do you want to pass in sensible defaults here? They are both 128 at the flag level, I believe, which seems like a decent default.

yoyodyne/util.py
```diff
@@ -16,6 +16,18 @@ def log_info(msg: str) -> None:
...
+def log_warning(msg: str) -> None:
+    """Logs a warning msg to sys.stderr.
+
+    We can additionally consider logging to a file, or getting a handle to
+    the PL logger.
+
+    Args:
+        msg (str): the message to log.
+    """
+    log_info(Warning(msg))
```
kylebgorman, line 28: I am wondering if this is really useful to convert to a `Warning` object. I have not found them useful in general.

kylebgorman (approved):
Okay, I like this, it definitely takes care of the issue. 

Why do this in the collator rather than the dataset object? Just asking questions: I don't have any particular argument why it is better to do it in one or the other place.

tests/collator_test.py
```diff
@@ -4,30 +4,42 @@
...
-        ("transformer", False, False, False),
+        "arch",
+        "has_features",
+        "has_target",
+        "expected_separate_features",
+        "max_src_length",
+        "max_tgt_length",
+    ],
+    [
+        ("feature_invariant_transformer", True, True, False, 128, 128),
```
kylebgorman, line 16: If you take my advice and make 128 the default for both, this won't be necessary, it occurs to me.

Adamits:
> Okay, I like this, it definitely takes care of the issue.
> 
> Why do this in the collator rather than the dataset object? Just asking questions: I don't have any particular argument why it is better to do it in one or the other place.

Yeah I had a lot of trouble deciding. I think I had thought I could just check the batch tensor, rather than looping over List. But ultimately I do not do that, so maybe the `Dataset` would be best? I could add to the initialization in the `Item` class. Do you think that'd be better?

yoyodyne/collators.py
Adamits: I never liked setting defaults in two different places. But since this could potentially be used as a CLI or a python library -- I have no idea what is typical here. I guess defaults here does make the most sense.

yoyodyne/util.py
Adamits: Do you suggest just doing 

```
util.log_info(Warning(msg))
```
in the collator then?

yoyodyne/collators.py
kylebgorman: I don't love this either, but we could have `defaults.py` which could contain these values as constants. What do you think? (That would be a separate PR.)

(So far I have just repeated ourselves in most places since there are other reasons to have defaults, e.g., for testing, and I want this library to be very useable from Python too.)

yoyodyne/util.py
kylebgorman: Oh, no, I am suggesting that this should not raise a warning at all, it should just log it to stderr like other things.

kylebgorman:
The way I was thinking about this, the checks should be on the padded data, since we're really checking whether we have _enough_ padding or not, with the understanding that there has to be some finite upper bound to this or the batches will grow too large. 

The collator is one place you could do this, since it sees the padding. The other place you could do this is [around here](https://github.com/CUNY-CL/yoyodyne/blob/c3154d1b913a64f4d445e5b17079a45bb03028de/yoyodyne/batches.py#L13), though you'd have some passing around to do to make that work. Do you think that'd be better?

Adamits:
> The way I was thinking about this, the checks should be on the padded data, since we're really checking whether we have _enough_ padding or not, with the understanding that there has to be some finite upper bound to this or the batches will grow too large.
> 
> The collator is one place you could do this, since it sees the padding. The other place you could do this is [around here](https://github.com/CUNY-CL/yoyodyne/blob/c3154d1b913a64f4d445e5b17079a45bb03028de/yoyodyne/batches.py#L13), though you'd have some passing around to do to make that work. Do you think that'd be better?

Right, `PaddedTensor` was my first choice too, but I didn't like having to tell it whether it is a source or target tensor. I think I can figure out a clean way to do this though. I will rework this PR and you can see what you think.

kylebgorman:
> Right, `PaddedTensor` was my first choice too, but I didn't like having to tell it whether it is a source or target tensor. I think I can figure out a clean way to do this though. I will rework this PR and you can see what you think.

If you want, keep this PR up so we can compare side-by-side. This seems like an important decision.

yoyodyne/collators.py
Adamits: `defaults.py` sounds like a good solution.

kylebgorman:
Closing in favor of #53.

PR #53, opened 2023-02-09
Maximum padded length errors

Adamits:
Addresses #40 

Adds the length comparison and message as a callback, which the padded tensor calls.

I also snuck in a little fix to `datasets.Item`.

### DIFF
--- a/yoyodyne/datasets.py
+++ b/yoyodyne/datasets.py
@@ -41,8 +41,8 @@ class Item(nn.Module):
         return self.features is not None
 
     @property
-    def has_targets(self):
-        return self.targets is not None
+    def has_target(self):
+        return self.target is not None
 
 
 class BaseDataset(data.Dataset):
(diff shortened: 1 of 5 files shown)


yoyodyne/datasets.py
```diff
@@ -41,8 +41,8 @@ def has_features(self):
         return self.features is not None
 
     @property
-    def has_targets(self):
-        return self.targets is not None
+    def has_target(self):
```
kylebgorman, line 44: good catch

kylebgorman (approved):
LGTM.

Honestly, this design is kind of fun. Nice use of callbacks.

Can you think of any arguments for it over the other one? Should we ask Travis?

Adamits:
> Honestly, this design is kind of fun. Nice use of callbacks.
Haha I thought it was fun too. Thanks.

> Can you think of any arguments for it over the other one? Should we ask Travis?
Hmmm, I think this one avoids an extra loop. Let's ask @bonham79 what he thinks between this and #51

bonham79:
@Adamits @kylebgorman Howdy y'all, (sorry, github didn't flag my name getting called). I approve of this version more than the former. It's cleaner and keeps track of possible issues in a less head scratcy way.

PR #54, opened 2023-02-10
Defaults module

### LINKED ISSUE #52: Defaults module
[Spun-off from discussion in #51.]

Create a `defaults.py` that defines defaults as constants. These are then used in `train.py` and `predict.py`, and in some cases, in the architecture modules themselves.

kylebgorman:
Adds a module for default constants and uses them.

This closes #52.

### DIFF
--- a/yoyodyne/collators.py
+++ b/yoyodyne/collators.py
@@ -27,8 +27,8 @@ class Collator:
         pad_idx,
         config: dataconfig.DataConfig,
         arch: str,
-        max_source_length=defaults.MAX_SOURCE_LENGTH,
-        max_target_length=defaults.MAX_TARGET_LENGTH,
+        max_source_length: int = defaults.MAX_SOURCE_LENGTH,
+        max_target_length: int = defaults.MAX_TARGET_LENGTH,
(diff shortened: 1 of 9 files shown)


yoyodyne/collators.py
```diff
@@ -28,8 +27,8 @@ def __init__(
         pad_idx,
         config: dataconfig.DataConfig,
         arch: str,
-        max_source_length: int = 128,
-        max_target_length: int = 128,
+        max_source_length=defaults.MAX_SOURCE_LENGTH,
```
Adamits, line 30: Is removing the type annotation here intentional?

Adamits:
Thanks for doing this, and for adding the collators argparse stuff!

I did not check exhaustively if any args were missed, but this LGTM.

Adamits (approved):
Thanks for doing this, and for adding the collators argparse stuff!

I did not check exhaustively if any args were missed, but this LGTM.

yoyodyne/collators.py
kylebgorman: Reverted.

PR #55, opened 2023-02-25
Adds automated LR tuning

kylebgorman:
The PTL trainer supports two kinds of auto-tuning as an alternative to actual training: one for batch size (which just tries to find the biggest value) and one for LR (which is a bit more sophisticated). Here, I have made the --auto_lr_find flag work.

If this flag is invoked, we run the tuner instead of the trainer's fit method, log the result, and then quit. Otherwise, training proceeds as normal.

I did not bother to enable the batch size one as it is substantially more work.

Only question is: do we want this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-Yoyodyne 🪀
-==========
+# Yoyodyne 🪀
 
 [![PyPI
 version](https://badge.fury.io/py/yoyodyne.svg)](https://pypi.org/project/yoyodyne)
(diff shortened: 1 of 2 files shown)


Adamits:
So this does some kind of hyperparameter sweep? And for batch size, it tries to find biggest that will fit in memory?

I think this LR one makes sense, but we probably need some documentation on it. As for batch size, maybe I need to read the PTL docs. In my experience, batch_size and LR have some trade-off in terms of getting a model to converge.

kylebgorman:
It tries an exponential series of learning rates then picks the best one using [this algorithm](https://arxiv.org/abs/1506.01186), logging it to stderr and terminating. If you like the code itself I'll add a paragraph of docs to the README. It's easy to explain.

This does not add support the batch size scaling thing (the other thing the tuner can do), and it'd be a decent amount of work to do so since the batch size has to live inside the modules then and we didn't do it anything like that.

Adamits:
I see. I think this is good, and we should not add the batch size one then!

Adamits (approved):

kylebgorman:
I have added README docs and also made it so the batch autoscaling flag (which is added by the trainer automatically) raises an exception.

PR #57, opened 2023-03-02
Fix lstm encoder layers

Adamits:
LSTM variants and pointer generator now works for >1 layers. There was a bug in the `init_hiddens` that used the number of encoder layers, not decoder. We also need to, for pointer generator, request only the last layer to use with the feature attentions.

However, the pointer generator now breaks when `encoder_layers != decoder_layers` because we follow the paper in initializing the decoder with the last encoder state. I *could* probably get the `n` last layers where `n` is `decoder_layers`, but this seems weird and also means that we still have the constraint that `encoder_layers >= decoder_layers`.

I think we either need to assert that `encoder_layers != decoder_layers`, or diverge from the paper and initialize the decoder like how we do the other LSTM models. I don't expect the latter to have much negative impact, but of course we'd want to test and compare results on some languages.

### DIFF
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -365,10 +365,12 @@ class PointerGeneratorLSTMEncoderDecoderFeatures(
         )
         # -> B x 1 x 4*hidden_size.
         context = torch.cat([source_context, feature_context], dim=2)
+        # num_layers x B x hidden_size
         _, (h, c) = self.decoder(
             torch.cat((embedded, context), 2), (last_h, last_c)
         )
-        hidden = h.transpose(0, 1)
+        # -> B x 1 x hidden_size
+        hidden = h[-1, :, :].unsqueeze(1)
         output_probs = self.classifier(torch.cat([hidden, context], dim=2))
         # Ordinary softmax, log will be taken at the end.
         output_probs = nn.functional.softmax(output_probs, dim=2)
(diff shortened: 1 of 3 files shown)


kylebgorman:
Having not reviewed the code carefully yet I propose that we raise a `pointer_generator.Error` exception if encoder layers does not match decoder layers in the ptr-gen init and not overthink things.

Adamits:
That sounds good! Will add that in a little bit.

yoyodyne/models/pointer_generator.py
```diff
@@ -9,6 +9,10 @@
 from . import attention, generation_probability, lstm
 
 
+class PointerGeneratorError(Exception):
```
kylebgorman, line 12: Just call this `Error`. It'll have the module name in front of it already.

yoyodyne/models/pointer_generator.py
```diff
@@ -39,6 +44,17 @@ def __init__(self, *args, **kwargs):
...
+    def _check_layer_sizes(self):
+        """Checks that encoder and decoder layers are the same number.
+
+        Raises:
+            PointerGeneratorError: _description_
+        """
+        if self.encoder_layers != self.decoder_layers:
+            msg = "encoder_layers needs to be the same as decoder_layers."
+            msg += f" {self.encoder_layers} != {self.decoder_layers}."
+            raise PointerGeneratorError(msg)
```
kylebgorman, line 56: Fix to match above suggestion.

kylebgorman (approved):
One nit otherwise LGTM. I take it this tests fine?

kylebgorman:
Gentle ping: do you need anything from me here?

Adamits:
> Gentle ping: do you need anything from me here?

No I just got busy and this slipped my mind. Thanks for the ping :). The tests pass -- I will fix the Error now.

PR #59, opened 2023-03-10
Fixes 16bit pointer-generator

Adamits:
- Changes var name to ptr probs
- instantiates ptr_probs to use dtype of source_attention_weights


kylebgorman (approved):
Cool! So this works with `--precision 16` or `--precision bfloat16` on `--accelerator cpu`? 

I agonized about how to make `ptr_probs` with `zeros_like` (as that also inherits device and dtype from the thing it's "like") but there's no existing object of the right shape as far as I can tell, so that won't work. 

LGTM.

Adamits:
> 

Right! Do you know if there is reason to ever use f16 instead bf16 by the way? I didn't test, but I imagine bf16 causes fewer issues?

kylebgorman:
> Right! Do you know if there is reason to ever use f16 instead bf16 by the way? I didn't test, but I imagine bf16 causes fewer issues?

Older GPUs don't support bf16, and I think some CPUs also. That's it though.

Issue #12, opened 2022-12-09
Transducer feature embedding


kylebgorman:
[copied from CUNY-CL/abstractness/issues/96]

Transducers make it somewhat difficult to condition generation on features. Makarov & Clematide use a one-hot embedding of features, concatenated itemwise to the input string embeddings, but @bonham79's pilot results suggest this is ineffectual. We can imagine at least a couple other ways to do it:

1. Features could be embedded, combined via mean, and then used to initialize the decoder state.
2. Features could be embedded, combined via mean, and then concatenated itemwise to the input string embeddings.
3. Features could be encoded and this encoding used to initialize the decoder state.
4. Features could be encoded, and then concatenated itemwise to the input strring embeddings.

bonham79:
So just cause it's been a bit, want to clarify 1 and 3. For 3, I can see how the hidden states from an encoder can be passed as initial states to the decoder, which would be pretty nifty thing to try. However, I'm not understanding how that would carry for embeddings. Am I misinterpreting?

kylebgorman:
#1 would constrain the embedding to a certain dimension unless you added some additional layer to reshape them how you wanted. #3 is the interesting case for me anyways---the one most likely to work.

bonham79:
Ran some tests on SIGMORPHON Basque data. Turns out merging 3 and 4 work the best (i.e. pass hidden states to decoder **and** concatenate to each character). Was able to get 91% accuracy with only five epochs of training. (Compared to 89% with just hidden states). I imagine this sorta makes sense, provides global knowledge of features to the whole decoder and local knowledge for each edit.

Outside  that, using embeddings instead of n_hot encoding does second best.

What are we feeling PR wise? Have a flag that allows multiple options or just use the best performing one?

kylebgorman:
I'm fine with just one solution.

How much worse does just using embeddings give us? And are you combining
with mean or what?

bonham79:
Just embeddings gives about 84/85% performance. So the encoding does a fair bit of work.

Embeddings are averaged. Hidden encoding are summed (borrowed from pointer generator.)

Issue #61, opened 2023-04-10
Feature indexing


bonham79:
`IndexFeatures` has an incorrect alignment for `feature_idx`. See https://github.com/CUNY-CL/yoyodyne/blob/18f19ae1858b46e1befd745f220b26106f2eb99f/yoyodyne/indexes.py#L160

 Easy fix but impacts https://github.com/CUNY-CL/yoyodyne/issues/47  Do we want to maintain a separate `features_map` to ease chance of these bugs or keep merged with `source_map`.

kylebgorman:
I don't understand the bug (why is this wrong? ) so I am assigning the issue to you---is this related to poor performance of the feature-ified transducer?

bonham79:
The way the index is set up, 'souece_map' includes feature vocab, so this is just providing the last index of the combined vocabularies, as opposed to the start of the feature vocabularies.

This would impact all models that treat features separately

kylebgorman:
@bonham79 can I close this too?

bonham79:
@kylebgorman yep1

Issue #47, opened 2023-02-07
Feature encoding enhancements


kylebgorman:
[This continues the discussion in #12.]

Both the transducer and the pointer-generator treat features in architecture-specific ways; issue #12 deals with their ideal treatment in the transducer, since the Makarov/Clematide mechanism of treating them as one-hot embeddings appears to be ineffectual. In contrast, they are just concatenated by the LSTM and transformer models. My proposal is that we should, in attentive LSTM and transformer, have separate (LSTM and transformer, respectively) encoders for features and these encoders should then be lengthwise concatenated. To be more explicit, imagine the source tensor is of size batch_size x hidden_size x source_length and the feature tensor is of size batch_size x hidden_size x feature_length. Then I propose that we concatenate these to form a tensor of size batch_size x hidden_size x (source_length + feature_length), and attention operates over that larger tensor.

As a result, we can use the features column to do multi-source translation in general. Furthermore, the dataset getter is no longer is conditioned on architecture: it just has features and no-feature variants, which makes things a bit simpler.

(Note that this will not work for the inattentive LSTM; something else will have to be done or we can just dump it.)

A distant enhancement is that it would be possible, in theory, to have different encoders for source and feature (LSTM vs. GRU vs. transformer); an even more distant enhancement would be to allow these to have different dimensionalities and use linear projection to map the feature encoding back onto the source encoding. I do not actually think we should do either of these, but it's a thing we could do...

Adamits:
> My proposal is that we should, in attentive LSTM and transformer, have separate (LSTM and transformer, respectively) encoders for features and these encoders should then be lengthwise concatenated.

That sounds great. I *do* think we should keep the option to do it the current way for 1) reproducibility of papers that do it that way (though I suspect it was often done this way out of convenience for using popular MT libraries in the past :D), and 2) perhaps to make comparisons wrt runtime?

Another note: then in the `FeatureInvariantTransformer`, we'd only need to embed the character embeddings with positional encodings, and would not need to track the feature idx in the vocab.

kylebgorman:
> > My proposal is that we should, in attentive LSTM and transformer, have separate (LSTM and transformer, respectively) encoders for features and these encoders should then be lengthwise concatenated.
> 
[...]

PR #63, opened 2023-04-10
Fixed feature bug and added embeddings to transducer.py

bonham79:
Fixes https://github.com/CUNY-CL/yoyodyne/issues/12 and https://github.com/CUNY-CL/yoyodyne/issues/61 . Lays groundwork for https://github.com/CUNY-CL/yoyodyne/issues/47 .

Fixes are as follows:
-- Addresses feature bug by adding a `features_map` to `IndexFeatures` that tracks encodings for features in dataset indexes. This can be utilized down the road when we expand feature encoding functionality.
-- This feature map encodes `special` encodings that parallels encoding for `source` and `target` strings. Doing so allows sharing of index locations across mappings. Please do not reorder the appending of `special.SPECIAL` in the `SymbolMap` class or things will break.
-- Adds a `feature_offset` to collator for when features and source strings are collated as a single string. This provides 'just in time' concatenation of the two to avoid further headaches when managing feature encodings at the dataloader level.
-- Updates `Collator` instantiation to utilize just a `dataset` object instead of multiple attributes from same object. 
-- Adds a `feature_embeddings` to `pointer_generator` to take advantage of direct referencing of `feature vocabulary size.` This will come in handy when we add in more feature customization.
--Updates `transducer` to use embeddings and brings its behavior more consistent with other models. Further enhancements can be made after we add more feature utility.

@Adamits  Would you make sure there was no breaking of `feature_invariant_transformer` functionality? I'm not up to date on that architecture so want to double check.


kylebgorman:
Would prefer you to edit the tests to make stuff pass. Feel free to expand on them within reason.On Apr 10, 2023, at 6:43 PM, Travis Bartley ***@***.***> wrote:﻿
@kylebgorman Since I messed around with Collator some tests are gonna be broken. Want me to defer to instantiation that conforms to the tests or edit the tests to conform to new instantiation?

yoyodyne/models/pointer_generator.py
```diff
@@ -501,10 +508,10 @@ def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
         """
         assert batch.has_features
         source_encoded, (h_source, c_source) = self.encode(
-            batch.source, self.encoder
+            batch.source, self.source_embeddings, self.encoder
         )
         features_encoded, (h_features, c_features) = self.encode(
-            batch.features, self.feature_encoder
+            batch.features, self.feature_embeddings, self.feature_encoder
```
Adamits, line 514: Question: Did batch.features previously index the embeddings matrix incorrectly? Do you have any idea when this was introduced?

bonham79: I remember indexing behavior being correct prior to the `indexes` implementation. So it may have been introduced during that update.

yoyodyne/train.py
```diff
@@ -296,10 +309,10 @@ def get_model(
         encoder_layers=encoder_layers,
         end_idx=train_set.index.end_idx,
         expert=expert,
-        features_vocab_size=getattr(
-            train_set.index, "features_vocab_size", -1
-        ),
-        features_idx=getattr(train_set.index, "features_idx", -1),
+        features_vocab_size=features_vocab_size,
+        features_idx=train_set.index.source_vocab_size
```
Adamits, line 313: Is this only here for the feature invariant transformer now? I think we could make the same change to that module. The features_idx is only used to zero out positional embeddings for features, since their positional info is irrelevant.

yoyodyne/train.py
```diff
@@ -283,6 +285,17 @@ def get_model(
     scheduler_kwargs = schedulers.get_scheduler_kwargs_from_argparse_args(
         **kwargs
     )
+    separate_features = train_set.config.has_features and arch in [
+        "pointer_generator_lstm",
+        "transducer",
+    ]
+    features_vocab_size = getattr(train_set.index, "features_vocab_size", 0)
+    vocab_size = (
```
Adamits, line 293: Should this perhaps be in the `Index`? Or does it need to be here since it depends on the model config (do we want to separate features or not)?

yoyodyne/models/transducer.py
```diff
@@ -617,31 +617,16 @@ def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
...
-        # B x 1 x n_feat -> B x seq_len x n_feat.
-        n_hot_features = n_hot_features.expand(-1, encoder_out.size(1), -1)
-        # Concatenates n-hot encoding onto each symbol.
-        encoder_out_feat = torch.cat((encoder_out, n_hot_features), dim=2)
+        # Prepares feature embeddings.
+        features_out = self.feature_embeddings(batch.features.padded)
+        features_out = torch.sum(features_out, dim=1)
+        denom = torch.sum(~batch.features.mask, dim=1, keepdim=True)
+        denom = denom.expand(-1, features_out.size(1))
+        features_out = features_out / denom
```
Adamits, line 625: Was this averaging in one of the publications for this model? If not, we should leave an option for the original behavior for reproducibility (I ask b/c it seems like features were handled differently before this PR).

yoyodyne/models/pointer_generator.py
bonham79: Ah, I think I know why this behavior wasn't affected. Since `pointer_generator` just uses one embedding and was instantiating from vocabulary, it was already the size of the conjoined `source + feature` space.  Since `batch.features` was still indexed in order with source vocabulary, it would have had the proper encodings.

yoyodyne/collators.py
```diff
@@ -83,14 +85,14 @@ def _target_length_warning(self, padded_length: int):
...
-    @staticmethod
     def concatenate_source_and_features(
+        self,
         itemlist: List[datasets.Item],
     ) -> List[torch.Tensor]:
         """Concatenates source and feature tensors."""
         return [
             (
-                torch.cat((item.source, item.features))
+                torch.cat((item.source, item.features + self.features_offset))
```
Adamits, line 95: Does this get run if there are no features? What does `cat` do in that case?

Adamits (commented):
I left a couple inline comments. They may not require any changes, but I want to be sure before I approve.

WRT feature invariant transformer, as long as the features_index is correctly computed, everything should work fine. 

I tried to make sure that I understand how everything works and it seems good to me, but there is certainly a chance that I missed something -- especially since I have not touched the indexing stuff in a long time (probably not since long before the `Indexes.py` stuff was added). Do we have a reliable way to test this right now?

As I understand this, we now index characters and features totally separately, and we combine them in the collator, by adding the source index length to all feature indices, for models that expect a single input string, and have a single embedding matrix? I think this seems good.

bonham79:
>  Do we have a reliable way to test this right now?

I ran all models over SIGMORPHON 2021 Polish data set and all instances where feature was included I saw a performance increase (same seed). That's the only test I had atm though. Can try on Basque since I remember that dataset _only_ sees improvement if features are included.

> As I understand this, we now index characters and features totally separately, and we combine them in the collator, by adding the source index length to all feature indices, for models that expect a single input string, and have a single embedding matrix? I think this seems good.

That is correct. It seems an easier way to refer to features without needing to do indexing math and leaves the point of failure only to the `Collator` method.

tests/collator_test.py
```diff
@@ -29,13 +29,18 @@
 def test_get_collator(
     arch, has_features, has_target, expected_separate_features
 ):
+    dataset = datasets.BaseDataset()
     config = dataconfig.DataConfig(
         features_col=3 if has_features else 0,
         target_col=2 if has_target else 0,
     )
+    index = indexes.BaseIndex()
+    index.pad_idx = 1
```
kylebgorman, line 38: Index doesn't have a `pad_idx` or `source_vocab_size` in either your version or mine.  Not sure what's going on here.

bonham79:
@kylebgorman @Adamits apparently my comments the first go around never went through, can you confirm if you received them this time?

kylebgorman:
No recent comments from you are here on GitHub nor in my inbox.

bonham79:
> No recent comments from you are here on GitHub nor in my inbox.

Oh that's annoying, all were pending because git thought I started a review...

kylebgorman:
It won't let me comment on any of your self-review comments, but I can add that you can TODO the boilerplate stuff in the test like so:

```
# TODO: reduce boilerplate.
```

bonham79:
> It won't let me comment on any of your self-review comments, but I can add that you can TODO the boilerplate stuff in the test like so:
> 
> ```
> # TODO: reduce boilerplate.
> ```

It actually gets fixed in my other PR. Should i add the TODO just for version history tracking or nah?

kylebgorman:
In that case no need, trust you on that ;)

K

bonham79:
> In that case no need, trust you on that ;) K

aight, think this is good to push then.

tests/collator_test.py
kylebgorman: Okay I'll come back to this later.

yoyodyne/datasets.py
```diff
@@ -82,6 +82,7 @@ def __init__(
         self.config = config
         self.samples = list(self.config.samples(filename))
         self.index = index if index is not None else self._make_index()
+        assert self.config.has_features is self.index.has_features
```
kylebgorman, line 85: Is this still needed? If so it should be `==` not `is`, I think

kylebgorman (approved):
This looks good to me now. Maybe add a TODO with the collator test thing. LGTM.

PR #67, opened 2023-05-29
Removes optimizer details" printout

kylebgorman:
This printout doesn't seem to serve much purpose since one can read the same information off of the `Arguments` printout, and I'm not sure what most of the information printed here means anyways...


Adamits (approved):
LGTM

PR #69, opened 2023-05-29
Updates versions of various helper libraries

kylebgorman:
This brings our `requirements.txt` up to date (modulo the huge migration to PyTorch 2.0 we still need to do).

`requirements.txt` is essentially notes about the environment we use to develop; `pyproject.toml` actually imposes a much looser set of restrictions for users.


Adamits (approved):
I don't really know which changes over time have necessitated these updates, so I am not sure there's much I can review. Seems ok tome though!

kylebgorman:
TBF it is not necessitated, I am just keeping things up to date as it were.

PR #71, opened 2023-06-05
Fixes decoding twice in validation

### LINKED ISSUE #70: Decoding twice in validation step
See here: https://github.com/CUNY-CL/yoyodyne/blob/master/yoyodyne/models/base.py#L219

Previously, I did this so I could first predict with greedy search to get an accuracy, and then predict with teacher forcing to compute a loss taht is comparable to how the train loss is computed. Currently, we predict twice with whatever is in the batch from the validation dataloader. IIRC this should have the gold targets in it.

This means that we are making identical validation predictions twice, neither of which is greedy -- they both have access to the gold history and use teacher forcing. I think we probably want to compute validation accuracy without the gold history, for which we would need to pass the batch through with no targets on it.

Then, we can either keep predicting a 2nd time to compute a loss with teacher forcing, or just compute the loss directly from the greedy predictions (and thus only decode once during validation).

@kylebgorman thoughts on this?

Adamits:
- Updates validation step to only call model forawrd once.
- Adds teacher_forcing param to  forward and decode methods.
- Calls validation step with teacher_forcing=False.
- Updates  methods to decode until EOS AND at least length of target in teacher_forcing mode.
- Calls predict step with teacher_forcing=False.

Closes #70 

Adds `teacher_forcing`, which tells the model to decode using the gold history, rather than the predicted history. This is typically how encoder-decoder inflection models are trained, but during validation/prediction, we force teacher_forcing to be False.

I additionally implemented a check that makes validation more efficient: we supply the target tensor, but only use it for early stopping on decoding if the decoder has predicted an EOS token for every sequence in the batch. This ensures that we can decode greedily up to at least the length of the targets, and then compute a loss in which we need `len(predicted == len(targets)`.

*IMPORTANT* 
- For Transducer, teacher_forcing means we get the gold actions from the expert. @bonham79 does this make sense? Or should I ignore this new parameter entirely in that architecture?
- Transducer errors with:

```
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/expert.py", line 105, in lookup
    return self.w2i[word]
KeyError: ConditionalIns(new=0)
```

I reverted all changes to test, and still reproduce the error. This is on the ConLL-SIGMORPHON 2017 inflection shared task English medium setting data, with the dev set. I suspect this is related to an old issue that happened when there was an OOV in the eval set. @bonham79 -- do you know if that issue still exists in the transducer?

### DIFF
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -435,11 +435,6 @@ class BaseEncoderDecoder(pl.LightningModule):
             type=float,
             help="Coefficient for label smoothing.",
         )
-        parser.add_argument(
-            "--teacher_forcing",
-            default=defaults.TEACHER_FORCING,
-            help="Use teacher forcing while training. Default: %(default)s.",
-        )
         # TODO: add --beam_width.
         # Model arguments.
         parser.add_argument(
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -209,17 +209,16 @@ class BaseEncoderDecoder(pl.LightningModule):
         """
         # Greedy decoding.
         # -> B x seq_len x output_size.
-        predictions = self(batch)
         target_padded = batch.target.padded
+        greedy_predictions = self(batch, teacher_forcing=False)
         accuracy = self.evaluator.val_accuracy(
-            predictions, target_padded, self.end_idx, self.pad_idx
+            greedy_predictions, target_padded, self.end_idx, self.pad_idx
         )
-        # We rerun the model with teacher forcing so we can compute loss.
-        # TODO: Update to run the model only once.
-        forced_predictions = self(batch)
         # -> B x output_size x seq_len. For loss.
-        forced_predictions = forced_predictions.transpose(1, 2)
-        loss = self.loss_func(forced_predictions, target_padded)
+        greedy_predictions = greedy_predictions.transpose(1, 2)
+        # Truncates predictions to the size of the target.
+        greedy_predictions = torch.narrow(greedy_predictions, 2, 0, target_padded.size(1))
+        loss = self.loss_func(greedy_predictions, target_padded)
         return {"val_accuracy": accuracy, "val_loss": loss}
 
     def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:
@@ -256,7 +255,8 @@ class BaseEncoderDecoder(pl.LightningModule):
         Returns:
             torch.Tensor: indices of the argmax at each timestep.
         """
-        predictions = self(batch)
+        batch.target.padded = None
+        predictions = self(batch, teacher_forcing=False)
         # -> B x seq_len x 1.
         greedy_predictions = self._get_predicted(predictions)
         return greedy_predictions
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -30,6 +30,7 @@ class BaseEncoderDecoder(pl.LightningModule):
     # Regularization arguments.
     dropout: float
     label_smoothing: Optional[float]
+    teacher_forcing: bool
     # Decoding arguments.
     beam_width: int
     max_target_length: int
@@ -59,6 +60,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         scheduler_kwargs=None,
         dropout=defaults.DROPOUT,
         label_smoothing=None,
+        teacher_forcing=defaults.TEACHER_FORCING,
         beam_width=defaults.BEAM_WIDTH,
         max_target_length=defaults.MAX_TARGET_LENGTH,
         decoder_layers=defaults.DECODER_LAYERS,
@@ -81,6 +83,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         self.scheduler_kwargs = scheduler_kwargs
         self.dropout = dropout
         self.label_smoothing = label_smoothing
+        self.teacher_forcing = teacher_forcing
         self.beam_width = beam_width
         self.max_target_length = max_target_length
         self.decoder_layers = decoder_layers
@@ -210,7 +213,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         # Greedy decoding.
         # -> B x seq_len x output_size.
         target_padded = batch.target.padded
-        greedy_predictions = self(batch, teacher_forcing=False)
+        greedy_predictions = self(batch)
         accuracy = self.evaluator.val_accuracy(
             greedy_predictions, target_padded, self.end_idx, self.pad_idx
         )
@@ -257,8 +260,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         Returns:
             torch.Tensor: indices of the argmax at each timestep.
         """
-        batch.target.padded = None
-        predictions = self(batch, teacher_forcing=False)
+        predictions = self(batch)
         # -> B x seq_len x 1.
         greedy_predictions = self._get_predicted(predictions)
         return greedy_predictions
@@ -432,6 +434,11 @@ class BaseEncoderDecoder(pl.LightningModule):
             "--label_smoothing",
             type=float,
             help="Coefficient for label smoothing.",
+        )
+        parser.add_argument(
+            "--teacher_forcing",
+            default=defaults.TEACHER_FORCING,
+            help="Use teacher forcing while training. Default: %(default)s.",
         )
         # TODO: add --beam_width.
         # Model arguments.
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -217,7 +217,9 @@ class BaseEncoderDecoder(pl.LightningModule):
         # -> B x output_size x seq_len. For loss.
         greedy_predictions = greedy_predictions.transpose(1, 2)
         # Truncates predictions to the size of the target.
-        greedy_predictions = torch.narrow(greedy_predictions, 2, 0, target_padded.size(1))
+        greedy_predictions = torch.narrow(
+            greedy_predictions, 2, 0, target_padded.size(1)
+        )
         loss = self.loss_func(greedy_predictions, target_padded)
         return {"val_accuracy": accuracy, "val_loss": loss}
 
--- a/yoyodyne/models/lstm.py
+++ b/yoyodyne/models/lstm.py
@@ -247,6 +247,9 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             # In teacher forcing mode the next input is the gold symbol
             # for this step.
             if teacher_forcing:
+                assert (
+                    target is not None
+                ), "Teacher forcing requested but no target provided"
                 decoder_input = target[:, t].unsqueeze(1)
             # Otherwise we pass the top pred to the next timestep
             # (i.e., student forcing, greedy decoding).
--- a/yoyodyne/models/lstm.py
+++ b/yoyodyne/models/lstm.py
@@ -199,6 +199,7 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
         encoder_mask: torch.Tensor,
         encoder_out: torch.Tensor,
         target: Optional[torch.Tensor] = None,
+        teacher_forcing: Optional[bool] = True
     ) -> torch.Tensor:
         """Decodes a sequence given the encoded input.
 
@@ -211,6 +212,7 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             encoder_out (torch.Tensor): batch of encoded inputs.
             target (torch.Tensor, optional): target symbols; if None, then we
                 decode greedily with 'student forcing'.
+            teacher_forcing (bool, optional): Whether or not to decode with teacher forcing.
 
         Returns:
             predictions (torch.Tensor): tensor of predictions of shape
@@ -239,12 +241,12 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
                 decoder_input, decoder_hiddens, encoder_out, encoder_mask
             )
             predictions.append(output.squeeze(1))
-            # If we have a target (training) then the next input is the gold
+            # In teacher forcing mode the next input is the gold
             # symbol for this step (i.e., teacher forcing).
-            if target is not None:
+            if teacher_forcing:
                 decoder_input = target[:, t].unsqueeze(1)
-            # Otherwise, it must be inference time and we pass the top pred to
-            # the next next timestep (i.e., student forcing, greedy decoding).
+            # Otherwise, it we pass the top pred to the next next 
+            # timestep (i.e., student forcing, greedy decoding).
             else:
                 decoder_input = self._get_predicted(output)
                 # Updates to track which sequences have decoded an EOS.
@@ -252,7 +254,10 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
                     finished, (decoder_input == self.end_idx)
                 )
                 # Breaks when all batches predicted an EOS symbol.
-                if finished.all():
+                # If we have a target (and are thus computing loss), 
+                # we only break when we have decoded at least the the 
+                # same number of steps as the target length.
+                if finished.all() and (target is None or decoder_input.size(-1) >= target.size(-1)):
                     break
         predictions = torch.stack(predictions)
         return predictions
@@ -382,11 +387,16 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
         else:
             return predictions
 
-    def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
+    def forward(
+        self,
+        batch: batches.PaddedBatch,
+        teacher_forcing: Optional[bool] = True,
+    ) -> torch.Tensor:
         """Runs the encoder-decoder model.
 
         Args:
             batch (batches.PaddedBatch).
+            teacher_forcing (bool, optional): Whether or not to decode with teacher forcing.
 
         Returns:
             predictions (torch.Tensor): tensor of predictions of shape
@@ -402,7 +412,7 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             )
         else:
             predictions = self.decode(
-                len(batch), batch.source.mask, encoder_out, batch.target.padded
+                len(batch), batch.source.mask, encoder_out, batch.target.padded, teacher_forcing
             )
         # -> B x seq_len x output_size.
         predictions = predictions.transpose(0, 1)
--- a/yoyodyne/models/lstm.py
+++ b/yoyodyne/models/lstm.py
@@ -199,7 +199,7 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
         encoder_mask: torch.Tensor,
         encoder_out: torch.Tensor,
         target: Optional[torch.Tensor] = None,
-        teacher_forcing: Optional[bool] = True
+        teacher_forcing: Optional[bool] = True,
     ) -> torch.Tensor:
         """Decodes a sequence given the encoded input.
 
@@ -245,7 +245,7 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             # symbol for this step (i.e., teacher forcing).
             if teacher_forcing:
                 decoder_input = target[:, t].unsqueeze(1)
-            # Otherwise, it we pass the top pred to the next next 
+            # Otherwise, it we pass the top pred to the next next
             # timestep (i.e., student forcing, greedy decoding).
             else:
                 decoder_input = self._get_predicted(output)
@@ -254,10 +254,12 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
                     finished, (decoder_input == self.end_idx)
                 )
                 # Breaks when all batches predicted an EOS symbol.
-                # If we have a target (and are thus computing loss), 
-                # we only break when we have decoded at least the the 
+                # If we have a target (and are thus computing loss),
+                # we only break when we have decoded at least the the
                 # same number of steps as the target length.
-                if finished.all() and (target is None or decoder_input.size(-1) >= target.size(-1)):
+                if finished.all() and (
+                    target is None or decoder_input.size(-1) >= target.size(-1)
+                ):
                     break
         predictions = torch.stack(predictions)
         return predictions
@@ -412,7 +414,11 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             )
         else:
             predictions = self.decode(
-                len(batch), batch.source.mask, encoder_out, batch.target.padded, teacher_forcing
+                len(batch),
+                batch.source.mask,
+                encoder_out,
+                batch.target.padded,
+                teacher_forcing,
             )
         # -> B x seq_len x output_size.
         predictions = predictions.transpose(0, 1)
--- a/yoyodyne/models/lstm.py
+++ b/yoyodyne/models/lstm.py
@@ -212,7 +212,8 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             encoder_out (torch.Tensor): batch of encoded inputs.
             target (torch.Tensor, optional): target symbols; if None, then we
                 decode greedily with 'student forcing'.
-            teacher_forcing (bool, optional): Whether or not to decode with teacher forcing.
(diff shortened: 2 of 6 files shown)


tbartley94:
>For Transducer, teacher_forcing means we get the gold actions from the expert. @bonham79 does this make sense? Or should I ignore this new parameter entirely in that architecture?

I never used that flag with the Transducer so it should not have an effect. The teacher forcing for that uses a different query from the SED parameters. 

>  I suspect this is related to an old issue that happened when there was an OOV in the eval set. @bonham79 -- do you know if that issue still exists in the transducer?

The OOV issue (iirc) was related to OOV _features_ , are you referring to that?

Adamits:
> > For Transducer, teacher_forcing means we get the gold actions from the expert. @bonham79 does this make sense? Or should I ignore this new parameter entirely in that architecture?
> 
> Yes that makes sense. The query from the expert is equivalent to teacher forcing. The main difference is that instead of the gold being static, it's recalculated at each step of transcription to allow correction during bad predictions.
> 
> > I suspect this is related to an old issue that happened when there was an OOV in the eval set. @bonham79 -- do you know if that issue still exists in the transducer?
> 
> The OOV issue (iirc) was related to OOV _features_ , are you referring to that?

Ah, yes possibly. Do you remember if we fixed that? I can debug this a bit, but I don't think its related to my changes to would rather not make changes on this PR.

kylebgorman:
Sorry y'all, I should be able to get to this tomorrow.

It does sound like the transducer is broken at head from your description---that should be solved separately once Travis is recovered from his ICASSP jaunt---happy to consult on that.

tbartley94:
> Ah, yes possibly. Do you remember if we fixed that? I can debug this a bit, but I don't think its related to my changes to would rather not make changes on this PR.

iirc no since it's being addressed in feature encoder update. that should be done next week when i'm stateside.

tbartley94:
> It does sound like the transducer is broken at head from your description---that should be solved separately once Travis is recovered from his ICASSP jaunt---happy to consult on that.

yeah, i'm trying to integrate the transducer more into basic lstm behavior so updates are easier. i give it about two more major updates and issues like these shouldnt' be popping up as often.

yoyodyne/models/pointer_generator.py
```diff
@@ -164,6 +164,7 @@ def decode(
         source_enc: torch.Tensor,
         source_mask: torch.Tensor,
         target: Optional[torch.Tensor] = None,
+        teacher_forcing: Optional[bool] = True,
```
kylebgorman, line 167: ditto from above.

yoyodyne/models/pointer_generator.py
```diff
@@ -203,29 +205,40 @@ def decode(
...
+                ):
                     break
         predictions = torch.stack(predictions)
         return predictions
 
-    def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
+    def forward(
+        self,
+        batch: batches.PaddedBatch,
+        teacher_forcing: Optional[bool] = True,
```
kylebgorman, line 234: ditto

yoyodyne/models/transformer.py
```diff
@@ -219,22 +236,33 @@ def _decode_greedy(
...
+            ):
                 break
         # -> B x seq_len x output_size.
         return torch.stack(outputs).transpose(0, 1)
 
-    def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
+    def forward(
+        self,
+        batch: batches.PaddedBatch,
+        teacher_forcing: Optional[bool] = True,
```
kylebgorman, line 253: ditto

kylebgorman (commented):
Thanks for this, this is a big update of sorts.

I have to ask whether there's some function we could write that would reduce the boilerplate...is there?

Other than my nits, I'd suggest you go ahead and submit even if you're getting an OOV error on some data in the transducer, so long as 1) you believe it to be independent and 2) there's a bug report/issue reported for that issue so we don't forget about it.

Adamits:
> > It does sound like the transducer is broken at head from your description---that should be solved separately once Travis is recovered from his ICASSP jaunt---happy to consult on that.
> 
> yeah, i'm trying to integrate the transducer more into basic lstm behavior so updates are easier. i give it about two more major updates and issues like these shouldnt' be popping up as often.

Cool -- sorry for the false alarm! Enjoy ICCASP.

Adamits:
> I have to ask whether there's some function we could write that would reduce the boilerplate...is there?

Ah good point. I think there is. In fact,  I think the `decode(...)` implementation for LSTM models is setup fairly differently than transformer and we could make these much more similar in a way that would reduce repeated code. iirc Travis had pointed this out in a PR or issue a while back.

I can work out those updates sometime before the week ends.

tbartley94:
> > I have to ask whether there's some function we could write that would reduce the boilerplate...is there?
> 
> Ah good point. I think there is. In fact, I think the `decode(...)` implementation for LSTM models is setup fairly differently than transformer and we could make these much more similar in a way that would reduce repeated code. iirc Travis had pointed this out in a PR or issue a while back.
> 
> I can work out those updates sometime before the week ends.

Don't worry about the boilerplate for models, I have it cleaned up in an upcoming PR.

Adamits:
Updated per the PR comments. Leaving updates around reduction of boilerplate and conforming `decode` methods to Travis' PR.

yoyodyne/models/base.py
```diff
@@ -256,7 +257,8 @@ def predict_step(
         Returns:
             torch.Tensor: indices of the argmax at each timestep.
         """
-        predictions = self(batch)
+        batch.target.padded = None
```
kylebgorman, line 260: I notice that `batch` and `predictions` aren't passed in. Is it supposed to work like that, getting them from context? Seems scary.

yoyodyne/models/lstm.py
```diff
@@ -239,21 +242,27 @@ def decode(
                 decoder_input, decoder_hiddens, encoder_out, encoder_mask
             )
             predictions.append(output.squeeze(1))
-            # If we have a target (training) then the next input is the gold
-            # symbol for this step (i.e., teacher forcing).
-            if target is not None:
+            # In teacher forcing mode the next input is the gold symbol
+            # for this step.
+            if teacher_forcing:
```
kylebgorman, line 247: Why can't this switch on whether or not `target` is null?

yoyodyne/models/lstm.py
```diff
@@ -199,6 +199,7 @@ def decode(
         encoder_mask: torch.Tensor,
         encoder_out: torch.Tensor,
         target: Optional[torch.Tensor] = None,
+        teacher_forcing: bool = True,
```
kylebgorman, line 202: Why can't this switch on whether or not `target` is null? The documentation for `target` seems to suggest that's what it does.

yoyodyne/models/lstm.py
```diff
@@ -402,7 +417,11 @@ def forward(self, batch: batches.PaddedBatch) -> torch.Tensor:
             )
         else:
             predictions = self.decode(
-                len(batch), batch.source.mask, encoder_out, batch.target.padded
+                len(batch),
+                batch.source.mask,
+                encoder_out,
+                batch.target.padded,
```
kylebgorman, line 423: Suggest making this:

`batch.target.padded if teacher_forcing else None`

and the removing the following argument, and adjusting interface to respect  that.

yoyodyne/models/pointer_generator.py
```diff
@@ -203,29 +205,40 @@ def decode(
...
                 finished = torch.logical_or(
                     finished, (decoder_input == self.end_idx)
                 )
-                # Breaks when all batches predicted an END symbol.
-                if finished.all():
+                # Breaks when all batches predicted an EOS symbol.
+                # If we have a target (and are thus computing loss),
+                # we only break when we have decoded at least the the
+                # same number of steps as the target length.
+                if finished.all() and (
```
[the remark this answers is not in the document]
kylebgorman: Make this consistent with the rest of them, whatever you do here.

kylebgorman (commented):
See my comments about on `validation_epoch_end` and how we're twice encoding whether or not we're using teacher forcing in the helpers.

yoyodyne/models/base.py
kylebgorman: Also wouldn't you want to null out the whole target portion of the batch and not just the padded vector?

bonham79:
> Ah, yes possibly. Do you remember if we fixed that? I can debug this a bit, but I don't think its related to my changes to would rather not make changes on this PR.

Checked the transducer, it's not really a bug but more because that dataset is poorly formatted, so there's character symbols in dev that don't occur in train. Due to the transducer predicting edits, not actual characters, there's no logic in place for predicting characters that have never be seen. I can fix it but for me that falls under the category of data processing.

Side note: That, uh, CONLL set had some interesting words in it...

kylebgorman:
> Checked the transducer, it's not really a bug but more because that dataset s poorly formatted, so there's character symbols in dev that don't occur in train. 

So can we fix this dataset automatically? Can you include the dataset formatting bug report here?

> Due to the transducer predicting edits, not actual characters, there's no logic in place for predicting characters that have never be seen. I can fix it but for me that falls under the category of data processing.

I wonder if we can catch and convert this error to something more informative?

bonham79:
>
> So can we fix this dataset automatically? Can you include the dataset formatting bug report here?

The fix would just be standard post processing (force single casing, stripping punctuation). Best I can provide for a bug report is just a comparison of dev vocabulary set vs train vocabulary set. 

It's now fixed in https://github.com/CUNY-CL/yoyodyne/pull/72 by inserting a <UNK> character when the desired output is outside training. 
 
> 
> I wonder if we can catch and convert this error to something more informative?

We could compare indexes from dev and train when doing an initial dataset instantiation just to flag if vocabularies aren't subsets. Issue is that I think this is just a transducer issue so a bit wasteful. Also, this strikes me as something that users should be responsible for in the data curation stage.

kylebgorman:
> It's now fixed in #72 by inserting a character when the desired output is outside training

If that's so let's forget about it. You do that with the bit about the unk index, right?

> We could compare indexes from dev and train when doing an initial dataset instantiation just to flag if vocabularies aren't subsets. Issue is that I think this is just a transducer issue so a bit wasteful. Also, this strikes me as something that users should be responsible for in the data curation stage.

I agree the user should be responsible for this but the problem I see is that the error message is not informative and gives no hint to the user how to fix it. The solution should be simple and cheap: catch the `KeyError` (if that's what it is) and convert it to more specific error message. (I think that try/except code is very cheap at runtime.)

bonham79:
> If that's so let's forget about it. You do that with the bit about the unk index, right?
> 
yep

> I agree the user should be responsible for this but the problem I see is that the error message is not informative and gives no hint to the user how to fix it. The solution should be simple and cheap: catch the `KeyError` (if that's what it is) and convert it to more specific error message. (I think that try/except code is very cheap at runtime.)

Error handling is strangely expensive for some reason. It's fine for instantiation but for forward passes it's a resource drain.  I think the fix with UNK makes it more consistent with other behavior so should be good.

kylebgorman:
I approve this modulo the unaddressed issues from my earlier review. I think @Adamits is busy with something else for a few days though...

yoyodyne/models/base.py
Adamits: > I notice that batch and predictions aren't passed in. Is it supposed to work like that, getting them from context? Seems scary.

I don't follow what you mean here.

> Also wouldn't you want to null out the whole target portion of the batch and not just the padded vector?

Ideally, I think yes. I can't remember which model behavior forced me to do this. Will retrace my steps and update (either with a fix, or clearer justification :D)

yoyodyne/models/lstm.py
Adamits: See my email: we can still use the target for early stopping if provided, even when evaluating without teacher forcing. Is that too tricky/confusing?

yoyodyne/models/base.py
Adamits: Not sure if this is what you mean here, but thinking about it more, it might be a bad idea to force remove targets here? On the one hand, I don't think `predict` should ever let the model see the targets. On the other hand, this should probably be handled elswhere that is less implicit.

The problem is that currently there is not obvious place in the code to say "don't use the target" (I guess setting teacher forcing to False handles this...). And I think its reasonable to pass in a test file with targets and expect the code to not use them at runtime when predicting.

kylebgorman: I mean i don't understand why this doesn't complain that `batch` is undefined. The following snippet of Python should give a `NameError` on `baz` because `baz` hasn't been defined. Similarly, `batch` and `prediction` are undefined in this method. We don't rely on this kind of scoping anywhere else in the entire project to my knowledge: we pass in the arguments we refer to.

```python
def foo(bar):
   return baz + baz
```

Yes, we should pass this as a kwarg and not forcibly remove the target.

kylebgorman (commented):
More notes.

yoyodyne/models/base.py
kylebgorman: Ah that's right, sorry to waste your time.

Adamits:
> Checked the transducer, it's not really a bug but more because that dataset is poorly formatted, so there's character symbols in dev that don't occur in train. Due to the transducer predicting edits, not actual characters, there's no logic in place for predicting characters that have never be seen. I can fix it but for me that falls under the category of data processing.

Is OOV poor formatting? I think of it as a valid test case.

> Side note: That, uh, CONLL set had some interesting words in it...

Yes i've noticed that too :D.

Looking further down the thread about this being on the user:

> I agree the user should be responsible for this but the problem I see is that the error message is not informative and gives no hint to the user how to fix it. The solution should be simple and cheap: catch the KeyError (if that's what it is) and convert it to more specific error message. (I think that try/except code is very cheap at runtime.)

Could we catch it when loading data? I think our code should load dev/test data, then load the train index. After loading the train index, we should be able to find all OOVs, and log a warning? And then do you think the user is supposed to preprocess their data and replace any OOV with some <UNK> symbol? We could alternatively supply an option in the library, not totally sure what is right. Actually come to think of it I thought we already had an UNK symbol in the special vocab that we use to encode stuff not in the index so maybe I am misunderstanding the problem...

yoyodyne/models/lstm.py
Adamits: Sorry I am a bit all over the place today, but also just realized that a user *should* be able to specify `--target_col=0` to control this from the CLI. Idk that we ever actually tested that though.

kylebgorman: Yes, that'd be great if we could just document that and have it work that way! (I'd also log the teacher/student status in the `train` or `predict` main.)

kylebgorman: On reflection I don't think this will (or could) work for training, just prediction. In training, we need to read in the target whether or not we're using it during decoding. In prediction though, this might...

Adamits: Right agreed. Maybe I need to think this through better from the perspective of what is most logical for a user.

I think the teacher forcing flag is useful for making it easy to understand the code during training (check the validation step code, see that we set teacher forcing to False during validation) -- and that this is a pretty natural behavior.

During prediction, we _could_ require a user to set the target col to 0, but then again, I think it is also a natural assumption that `predict_step` should not have access to the targets, even if they are in the file. Maybe the best course of action is to leave it up to the user, but log a warning if target is not zero'd out. 

My only concern is if this warning will be understandable :D. I think the default behavior would be that we predict with teacher_forcing=False, but that, as in validation, we can stop early once we have predicted at least as many tokens as the target length.

kylebgorman: At the highest level I think the interface should be:

* `--forcing` defaults to "default" (teacher during training, student during prediction)
* `--forcing student` does what you think regardless of whether we're doing training or prediction
* `--forcing teacher` ... ditto

Regardless of how the dataset args are passed, you should need a valid target column for teacher forcing and it shouldn't matter for student forcing.

Adamits: Been trying to add this as an argument that we call from train/predict, but I am unsure of how pass an argument through the training loop in lightning, and cannot find anything in docs/issues.

Now I am wondering if we should consider `teacher_forcing` as only valid during training, and add it as a model attribute (if we later extend it to a teacher forcing "ratio", that can still live on the model). Then, the question still remains, where should we tell the forward/decode method to ignore the gold symbols if they exist?

One clean solution (1) is to always force `teacher_forcing=False` in the validate/predict steps like we are doing now. Instead, I suppose we could (2) check if the model is in `eval()` mode? I think (1) makes the most sense. But I keep worrying if the semantics of this is confusing. 

Additional point: I would rather keep the argument as `--teacher_forcing` and have it as either a bool or float if we want a ratio. I think having forcing types is a bit weird since IDK if `student_forcing` is a common term, not if there could be any other possible type of forcing that is not expressed by an "amount" of teacher forcing.

Adamits:
Ok, I have commited an update to the design.

`teacher_forcing` is now only relevant as a training argument, and is set on the model. I have overloaded it as a local variable to the `decode` method (plz let me know if this is problematic and maybe we can come up with a new name. Since this PR is really just about validation issues, maybe I can deal with tweaks to this in a separate issue?).

I think after Travis' large refactor, maybe I should change teacher_forcing to a `float`, and add a method that `decode` calls before each step to decide whether or not to pass the previous gold step or not. It will default to 1, and be set to 0 in eval mode -- future work though.

yoyodyne/models/lstm.py
bonham79: Jumping in on this, unless it was changed a few prs ago, I believe I wrote into `predict` that `target-col` is automatically set to `0` just for these concerns. So during the predict stage, the upstream should be stopping any presence of target.

bonham79:
Oh sorry, email was clogged and didn't see these comments.
 
> Is OOV poor formatting? I think of it as a valid test case.
> 

Mmmm, the `UNK` edit character allows management so it's not too hard  to do. But, given that our domain is string transduction, I don't see a good rationale for it. Why would you train your model on a dataset that doesn't contain the alphabet of your potential strings? You should be able to find a dataset with minimal coverage (outside of cryptography and potentially Hanji/Kanji, and even then...) 

And in the case you don't find that dataset...what do you hope to learn? Idk, it seems like too low a bar to accommodate for imo.

> 
> Looking further down the thread about this being on the user:
> 
> > I agree the user should be responsible for this but the problem I see is that the error message is not informative and gives no hint to the user how to fix it. The solution should be simple and cheap: catch the KeyError (if that's what it is) and convert it to more specific error message. (I think that try/except code is very cheap at runtime.)
> 
> Could we catch it when loading data? I think our code should load dev/test data, then load the train index. After loading the train index, we should be able to find all OOVs, and log a warning? And then do you think the user is supposed to preprocess their data and replace any OOV with some symbol? We could alternatively supply an option in the library, not totally sure what is right. Actually come to think of it I thought we already had an UNK symbol in the special vocab that we use to encode stuff not in the index so maybe I am misunderstanding the problem...

The issue was that for transducer there was an UNK symbol but not an UNK edit action. So it knew the symbol existed, it just didn't have a way to respond to it. Now it should have the option to choose an UNK symbol, so the error shouldn't pop up.

Re: OOV checking. I like that idea. Thinking it would work best as (yet anothe!) CLI flag. Probably like `--check-oov` that reads an index for the `dev` set and raises a warning. Would be a new PR though.

kylebgorman:
I’m fine pushing OOV handling into another PR.

K

bonham79:
> I think after Travis' large refactor, maybe I should change teacher_forcing to a `float`, and add a method that `decode` calls before each step to decide whether or not to pass the previous gold step or not. It will default to 1, and be set to 0 in eval mode -- future work though.

That's gonna be a bit of a headache in future updates. `teacher_forcing` reads like such a `bool` variable that having it serve as a `float` will cause some bugs. (Not nay-saying, just  acknowledging that I'm enough of an idiot that I will make that mistake.)

kylebgorman:
I am now thinking we may have jumped the gun exposing it to the user, and would be better off just forcing teacher forcing during training and student forcing during inference. I think we should actually hyperparameter-experiment with a probabilistic mix of student forcing during training (i.e., try to see if that Nicolai paper is right for the kind of problems we care about) and only then it out as a flag if it makes an impact.

yoyodyne/models/lstm.py
```diff
@@ -239,21 +244,27 @@ def decode(
                 decoder_input, decoder_hiddens, encoder_out, encoder_mask
             )
             predictions.append(output.squeeze(1))
-            # If we have a target (training) then the next input is the gold
-            # symbol for this step (i.e., teacher forcing).
-            if target is not None:
+            # In teacher forcing mode the next input is the gold symbol
+            # for this step.
+            if teacher_forcing:
                 decoder_input = target[:, t].unsqueeze(1)
```
kylebgorman, line 250: I have a minor suggestion. Put an assertion (or module-specific error) here to catch when `teacher_forcing` is True but `target` is None. It shouldn't be reachable but if it is that'd be better than `TypeError: 'NoneType' object is not subscriptable`.

```python
if teacher_forcing:
   assert target is not None, "Teacher forcing requested but no target provided"
   ...
```

yoyodyne/models/lstm.py
kylebgorman: I think what we have here is good for now...resolving.

yoyodyne/models/transformer.py
```diff
@@ -187,8 +188,24 @@ def decode(
...
+        Args:
+            encoder_hidden (torch.Tensor): Hidden states from the encoder.
+            source_mask (torch.Tensor): Mask for the encoded source tokens.
+            targets (torch.Tensor, optional): The optional target tokens,
+                which is only used for early stopping during validation
+                if the decoder has predicted [EOS] for every sequence in
+                the batch.
+
+        Returns:
+            torch.Tensor: predicitons from the decoder.
```
kylebgorman, line 207: typo: `predicitons`

kylebgorman (approved):
Okay LGTM.

I suggest assertions to catch teacher forcing requested without target passed. 

I note that there is some inconsistency in interface:

* the LSTM and pointer generator accept explicit `teacher_forcing` as an argument to `decode`
* the transformer reads it from `self.teacher_forcing`
* the transducer does both, I think. 

Is there anything to be done about this?

Adamits:
> Okay LGTM.
> 
> I suggest assertions to catch teacher forcing requested without target passed.

Ok, I will add an assertion good idea.

> I note that there is some inconsistency in interface:
> 
> * the LSTM and pointer generator accept explicit `teacher_forcing` as an argument to `decode`
> * the transformer reads it from `self.teacher_forcing`
> * the transducer does both, I think.
> 
> Is there anything to be done about this?

I believe transformer is the only one that is inconsistant. This is because the interface is already inconsistent, which, I think @bonham79  is changing anyway?

kylebgorman:
> I believe transformer is the only one that is inconsistant. This is because the interface is already inconsistent, which, I think @bonham79 is changing anyway?

He confirms this.

bonham79:
> > Okay LGTM.
> > I suggest assertions to catch teacher forcing requested without target passed.
> 
> Ok, I will add an assertion good idea.
> 
> > I note that there is some inconsistency in interface:
> > 
> > * the LSTM and pointer generator accept explicit `teacher_forcing` as an argument to `decode`
> > * the transformer reads it from `self.teacher_forcing`
> > * the transducer does both, I think.
> > 
> > Is there anything to be done about this?
> 
> I believe transformer is the only one that is inconsistant. This is because the interface is already inconsistent, which, I think @bonham79 is changing anyway?

Correct.

tbartley94:
> > I believe transformer is the only one that is inconsistant. This is because the interface is already inconsistent, which, I think @bonham79 is changing anyway?
> 
> Correct.

To follow up, if we do this pr before the other one, it'll give me a chance to clean up the interface further for consistency. So assume interface changes will happen in follow up.

tbartley94:
> FYI I **believe** need to make some fixes to the predict script (different PR) -- not sure if I should wait for your PR to merge first? I also have some work adding wandb sweeps, but that should not interfere too much with your work AFAIK.
 
IIRC there's no change to predict behavior (modules are loaded as part of the model) so you're good to go to make changes async to other pr.

kylebgorman:
Okay should I merge?

Adamits:
> Okay should I merge?

Yes!

tbartley94:
> Okay should I merge?

yeeep

PR #74, opened 2023-06-28
Fixes predict no target

Adamits:
Adds the non-existant (but still called -- oh the perils of writing software in interpreted languages!) `source_features_samples`, and makes fixes for when no target is passed through.

Also adds @kylebgorman's suggestion for passing `batch.target.padded` to `decode` when there is no target.


kylebgorman (approved):
LGTM *chef's kiss*.

PR #79, opened 2023-06-30
Removes label smoothing nullability

kylebgorman:
Instead, label smoothing is disabled when it is set to 0.0, which is the default.


Adamits (approved):
LGTM

PR #82, opened 2023-06-30
Adds docs for making releases

kylebgorman:


Adamits (approved):
Just want to double check for releasing, no `push` is needed?

Otherwise, LGTM

kylebgorman:
> Just want to double check for releasing, no `push` is needed?

Good catch, that's my error.

PR #91, opened 2023-07-04
Data modules refactor

kylebgorman:


yoyodyne/data/tsv.py
```diff
@@ -0,0 +1,175 @@
...
+        if self.source_col < 1:
+            raise Error(f"Invalid source column: {self.source_col}")
+        if self.target_col < 0:
+            raise Error(f"Invalid target column: {self.target_col}")
+        if self.target_col == 0:
+            util.log_info("Ignoring targets in input")
+
+    @staticmethod
+    def _tsv_reader(path: str) -> Iterator[str]:
+        with open(path, "r") as source:
```
Adamits, line 54: Not really important, but `source` is semantically overloaded in data modules (the file, or the source string).

yoyodyne/models/lstm.py
```diff
@@ -355,9 +355,9 @@ def beam_decode(
                 char_likelihoods,
                 decoder_hiddens,
             ) in likelihoods:
-                # This is 1 x 1 x target_vocab_size since we fixed batch size
-                # to 1. We squeeze off the first 2 dimensions to get a tensor
-                # of target_vocab_size.
+                # This is 1 x 1 x target_vocab_size since batch size is 1.
+                # We squeeze off the fist 2 dimensions to get a tensor of
```
Adamits, line 359: fist -> first

yoyodyne/models/pointer_generator.py
```diff
@@ -232,7 +232,7 @@ def decode(
                 finished = torch.logical_or(
                     finished, (decoder_input == self.end_idx)
                 )
-                # Breaks when all batches predicted an EOS symbol.
+                # Breaks when all data predicted an EOS symbol.
```
Adamits, line 235: I think we want `data` -> `batches` and this just happened from find + replace?

yoyodyne/models/pointer_generator.py
```diff
@@ -562,7 +561,7 @@ def forward(
             batch.source, self.source_embeddings, self.encoder
         )
         features_encoded, (h_features, c_features) = self.encode(
-            batch.features, self.feature_embeddings, self.feature_encoder
+            batch.features, self.features_embeddings, self.feature_encoder
```
Adamits, line 564: Looks like it is initialized as `self.feature_embeddings`, so one of them needs to be updated if I am not mistaken.

Adamits (changes requested):
This is a big change! The `DataModule` change looks really nice. Left some comments and questions. Most importantly -- I am not sure that we still support loading from a particular index when making predictions on a new dataset.

yoyodyne/data/datasets.py
```diff
@@ -0,0 +1,234 @@
...
+            index (indexes.Index).
+            cell_parser (tsv.CellParser).
+        """
+        super().__init__()
+        self.samples = samples
+        self.index = index
+        self.cell_parser = cell_parser
+
+    # Encoding.
+
```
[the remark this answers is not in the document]
kylebgorman: I use a blank line when it scopes over multiple methods... and no blank otherwise. IDK if that's standard but I think it's defensible.

yoyodyne/data/datamodules.py
```diff
@@ -0,0 +1,189 @@
...
+    # Required API.
+
+    def _dataset(self, path: str) -> datasets.Dataset:
+        return datasets.Dataset(
+            list(self.tsv_parser.samples(path)),
+            self.index,
+            self.cell_parser,
+        )
+
+    def train_dataloader(self) -> data.DataLoader:
```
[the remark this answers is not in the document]
kylebgorman: I will confirm this. If I'm wrong I can modify this so that it acts in some relevant properties like it's cached.

yoyodyne/data/datasets.py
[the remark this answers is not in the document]
Adamits: Ah I see, makes sense to me.

yoyodyne/models/lstm.py
kylebgorman: +1

yoyodyne/models/pointer_generator.py
kylebgorman: This inconsistency bothers me so I'll put a consistency fix to this module in a separate PR so there's less irrelevant change in this PR.

Out of curiosity and while you're here, is there any good argument for having a separate features embedding in the pointer generator? It seems like one ought to be able to share one for source and features since they are already guaranteed to be disjoint.

yoyodyne/data/datasets.py
```diff
@@ -0,0 +1,234 @@
...
+        """Decodes a source tensor.
+
+        Args:
+            indices (torch.Tensor): 2d tensor of indices.
+
+        Yields:
+            str: Decoded source strings.
+        """
+        for symbols in self._decode(indices, self.index.source_map):
+            yield self.cell_parser.source_string(symbols)
```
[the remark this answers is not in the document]
kylebgorman: That's right.

yoyodyne/models/pointer_generator.py
Adamits: I believe this was added to support some changes for tracking features and tokens separately? This just avoids clashing indices, but I also think @bonham79 may have updated s.t. features/tokens can be tracked separately, but then combined in the collator?

kylebgorman: The indices can't crash anyways because we obfuscate features. But I guess it's harmless and maybe also fixed by Travis's PR so no worries.

Adamits: I think it was added here: https://github.com/CUNY-CL/yoyodyne/commit/f6cff4639916a2475fc330309c73ad8c918b3273 -- so guessing it was needed in some previous feature handling, but is no longer needed?

yoyodyne/data/datasets.py
```diff
@@ -0,0 +1,234 @@
...
+                for symbol in symbols
+            ],
+            dtype=torch.long,
+        )
+
+    def encode_source(self, cell: str) -> torch.Tensor:
+        """Encodes a source string, padding with start and end tags.
+
+        Args:
+            cell (str).
```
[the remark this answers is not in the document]
bonham79: ditto on `cell` being ambiguous to parse.

yoyodyne/data/tsv.py
```diff
@@ -0,0 +1,175 @@
...
+
+
+class Error(Exception):
+    """Module-specific exception."""
+
+    pass
+
+
+@dataclasses.dataclass
+class TsvParser:
```
bonham79, line 23: for separate PR, but i'm wondering how trivial it would be to also pass `json` now we're including other dataprocessing libraries like `csv`

yoyodyne/train.py
```diff
@@ -103,114 +93,49 @@ def get_trainer_from_argparse_args(
...
-        batch_size=batch_size,
-        shuffle=True,
-        num_workers=1,  # Our data loading is simple.
-    )
-    dev_loader = data.DataLoader(
-        dev_set,
-        collate_fn=collator,
-        batch_size=2 * batch_size,  # Because we're not collecting gradients.
-        num_workers=1,
+) -> data.DataModule:
```
bonham79, line 98: god this is beautiful

kylebgorman:
Closing this in favor of #110, which I think is closer to bug-free. Could never figure out what was broken with this. However I have taken all the comments: I also have confirmed that the datasets and loaders are only created once and each file only parsed once.

PR #94, opened 2023-07-05
Update README.md

kylebgorman:
This doesn't work so I'm removing the reference to it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,9 +79,6 @@ Alternatively, for the SIGMORPHON 2016 shared task data format:
 
 this format is specified by `--features-col 2 --features-sep , --target-col 3`.
 
-In order to ensure that targets are ignored during prediction, one can specify
-`--target_col 0`.
-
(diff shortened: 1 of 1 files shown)


Adamits:
I am fairly certain that I tested it in several setups for all architectures in this PR https://github.com/CUNY-CL/yoyodyne/pull/74 and it worked? Either my memory is wrong or maybe we have a regression?

kylebgorman:
I actually didn't test it---we jsut have a TODO that says it's not working:

https://github.com/kylebgorman/yoyodyne/blob/61fbb9a085d081c64736343cdb51225765f736cc/yoyodyne/predict.py#L53

If it is working then we should revert this and remove that TODO.

Adamits:
Ah, I should have removed that TODO then! I can test to be sure.

kylebgorman:
Closing in favor of #97.

PR #95, opened 2023-07-05
Standardizes use of "features" rather than "feature" when referring to the third column

kylebgorman:
This is for pointer-generator only.


kylebgorman:
There is a spurious flakiness in PyPI hashes that is causing errors on CircleCI. I am have manually confirmed that the code passes flake8 so I overriding it.

PR #98, opened 2023-07-05
Adds ReduceLROnPlateua scheduler

Adamits:


yoyodyne/schedulers.py
```diff
@@ -108,6 +113,54 @@ def __repr__(self) -> str:
...
+        Args:
+            optimizer (optim.Optimizer): optimizer.
+            reduceonplateau_mode (str): Determines whether to reduce the
+                LR when the metric stops decreasing (`min`) or increasing (`max`).
+            reduceonplateau_factor (float): Factor by which the
+                learning rate will be reduced. new_lr = lr * factor.
+            reduce_lr_patience (int): Number of epochs with no improvement
+                before reducing LR.
+            min_lr (float): Lower bound on the learning rate.
+            **kwargs
```
kylebgorman, line 139: can you add `: ignored.` here?

yoyodyne/schedulers.py
```diff
@@ -149,6 +202,37 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         help="Number of iterations until the LR multiplier reaches "
         "--end_factor (lineardecay scheduler only). Default: %(default)s.",
     )
+    parser.add_argument(
```
kylebgorman, line 205: Can you put these flag default values in `defaults` instead?

kylebgorman (commented):
Cool, I assume you tested this out extensively on the hyperparameters project? 

A few nits etc., suggesting a different flag terminology too though I haven't thought deeply about this yet.

yoyodyne/models/base.py
```diff
@@ -336,6 +337,18 @@ def _get_lr_scheduler(
             "interval": "step",
             "frequency": 1,
         }
+        # For reduce on plateau min, we reduce when val loss stops decreasing.
```
kylebgorman, line 340: This is really unexpected for me---is this standard practice to have both? If so why not call the two modes something like `--reduceonplateau_mode loss` and `--reduceonplateau_mode accuracy` instead?
If you want to stick with this, maybe you could spell this out in slightly less telegraphic language? I think something like:

```
# When using `reduceonplateau_mode min`, we reduce when validation loss stops decreasing. When using `reduceonplateau_mode max`, we reduce when validation accuracy stops increasing.
```

Adamits: The scheduler pretty much just follows torch: https://pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.ReduceLROnPlateau.html. I assume that `max` mode would only make sense for accuracy. However, I would expect `min` on validation loss to be the norm.

I can either force it to not expose the other options, or update the comment. Do you have a preference?

kylebgorman: I jsut think _calling them_ max and min is burying the lede, when what you're doing is "tracking plateauing loss" vs. "tracking plateauing accuracy". The underlying behavior from PyTorch doesn't really apply in the sense that you can pass any metric you want to it but we're just supporting these two. (It'd be crazy to reduce on min accuracy or max loss so let's perish the thought.)

Adamits: Ah I see. In that case I like your suggestions for `loss` or `accuracy` better, but I am also slightly hesitant to change too much from the pytorch interface. Right now I lean changing it though.

kylebgorman: There is an alternative (though I don't recommend it) in which you have both flags for the choice of metric and min vs. max. I don't think this is a good idea because I don't think you should expose crazy things to the user

  don't think min entails "validation loss" or max entails "validation accuracy" so it certainly should be more explicit than the current form.

Adamits: Does it look ok following just the first suggestion?

yoyodyne/schedulers.py
```diff
@@ -108,6 +115,55 @@ def __repr__(self) -> str:
...
+            patience=reduceonplateau_patience,
+            min_lr=min_lr,
+        )
+
+    def __repr__(self) -> str:
+        return (
+            f"{self.__class__.__name__}("
+            f"{self.optimizer}, {self.mode}, "
+            f"{self.factor}, {self.min_lrs}), "
+            f"{self.patience}"
```
kylebgorman, line 163: Possibly dumb question but do you mean to have that closed parentheses before the end? This will look like `ReduceOnPlateau(Adam, "loss", .1, 0), 5`.

kylebgorman (approved):
LGTM (except maybe the thing re: `repr` let me know).

yoyodyne/schedulers.py
Adamits: Oh no, that was a mistake good catch.

PR #101, opened 2023-07-06
Calls the features stuff "features", with a plural.

kylebgorman:
We are somewhat inconsistent but I think this is the best parallel and follows what's done at the CLI level.


Adamits (approved):
LGTM

PR #102, opened 2023-07-06
Punctuation fixes to flags (mostly)

kylebgorman:
I also note that min_lr is surely a float in [0, 1) not an integer (even though PyTorch gives it as 0, it is typed a float).


Adamits (approved):
You are right, it should definitely be a float. Just a note, many of these LR-related hyperparameters support "parameter groups" in torch. If we ever want to support that, some of them, including min_lr will need to be `List`s. Would need to think about the use-case, and what all would need to be updated before caring to prioritize that, though.

LGTM!

PR #105, opened 2023-07-07
More precision in checkpoint names

kylebgorman:
This saves the checkpoint accuracies to 3 digits instead of 2, and stores 3-digit epoch counts.


Adamits (approved):
Nice, LGTM

PR #106, opened 2023-07-07
Improves model logging

### LINKED ISSUE #104: Informative model logging
Right now we log both the "model" (decoder and default encoder) and the "module" (encoder) lookups [here](https://github.com/CUNY-CL/yoyodyne/blob/54a7e75d942f4ef3f257c6123103fadf7484efc6/yoyodyne/models/__init__.py#L35) and [here](https://github.com/CUNY-CL/yoyodyne/blob/54a7e75d942f4ef3f257c6123103fadf7484efc6/yoyodyne/models/modules/__init__.py#L53) respectively. This gives us the following printout, for a pointer-generator with a transformer encoder layer (cf. Singer & Kann 2020):

```
Model: PointerGeneratorLSTMEncoderDecoder
Model: TransformerEncoder
```

This is not as informative as it could be. I submit it would be better as something like:

```Model: transformer encoder, pointer-generator LSTM decoder```

I think this could be done by adding properties to each "model" and "module" class and then concatenating them together like:

```python
util.log_info(f"Model: {encoder.name} encoder, {model.name} decoder")
```

Thoughts? If this proposal appeals it should be easy for me to implement.

kylebgorman:
Instead of logging at lookup, logs at the end of class construction. Example:

```
Model: pointer-generator
Source encoder: LSTM
Features encoder: linear
Decoder: attentive LSTM
```

Closes #104.

### DIFF
--- a/yoyodyne/models/__init__.py
+++ b/yoyodyne/models/__init__.py
@@ -2,7 +2,6 @@
 
 import argparse
 
-from .. import util
 from .base import BaseEncoderDecoder
 from .lstm import AttentiveLSTMEncoderDecoder, LSTMEncoderDecoder
 from .pointer_generator import PointerGeneratorLSTMEncoderDecoder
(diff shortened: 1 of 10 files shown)


Adamits (approved):
This is awesome. LGTM

PR #113, opened 2023-07-19
Improves sweeps example

### LINKED ISSUE #66: Updates and Documentation for WandB sweeps
Based on other work were doing, we should add some documentation and make necessary tweaks for running a W&B sweep with this codebase.

- Add documentation and examples of running WandB sweeps with Yoyodyne.
- Make updates to codebase so PTL and WandB play nice wrt logging hyperparameters, etc.
- Update PTL to log max validation accuracy.

kylebgorman:
I have modified the sweeps example to make it slightly easier to use:

* I make the `--entity` and `--project` fields mandatory. I was unable to get this working without both.
* I use the nice YAML format and the `wandb` CLI tool to create the sweep---Python is used to run it (I don't see any way around this) and to retrieve the results.
* I modified the docs slightly to reflect.

I consider this to close #66.


kylebgorman:
Actually the last step can be done (to .csv not .tsv) directly on the wandb website: looking at a given project, click the downward-facing download arrow then select "CSV Export" and follow the dialogue. I wonder if I should just include those instructions rather than a command to do it.

Adamits:
Yeah I actually knew about that (and then forgot), but originally just used
the API to see if there was any other useful info I could pull. Ultimately
I don't think using the API adds much for us iirc, so I guess downloading
the CSV and writing a script to work with it directly is probably a fine
option.

Adam

PR #116, opened 2023-07-19
Better support for not parsing irrelevant fields

kylebgorman:
This makes it so that `yoyodyne-predict` works with `--features_col 0`. That is, if this is specified it doesn't bother to parse the features column at all, which saves you a little time.

It does this by delegating `has_features` and `has_target` to the parser rather than to the index, and fixes a sneaky bug on the last line of the datasets module.


Adamits (approved):
So the parser already defines `has_features` and `has_target`?

Nice good catch on these. LGTM.

Issue #78, opened 2023-06-29
W&B config warnings when training a sweep agent


Adamits:
Currently, when training a sweep agent, we need to start the agent within a hyperparameter config, which automatically logs all of the hyperparameters to W&B. We also need to initialize the pytorch-lightning `WandbLogger`, which under the hood attempts to log the config again. See here for details: https://github.com/wandb/wandb/issues/2641.

Ideally, we would solve this by not updating twice, however, this may be out of our control as we rely on `wandb` and `pytorch-lightning` for those behaviors. 

A work-around is to suppress the W&B warning message. So far, `warnings.filterwarnings` does not work -- but we should investigate this more.

kylebgorman:
l believe this is irrelevant as of #308; closing.

PR #117, opened 2023-07-20
Silences warning about multiple WandB loggers

kylebgorman:
As I understand it this is not the warning discussed in #78 but it's good to silence anyways.

The full error reads:

```
UserWarning: There is a wandb run already in progress and newly created instances of `WandbLogger` will reuse this run. If this is not desired, call `wandb.finish()` before instantiating `WandbLogger`.
```

@Adamits please let me konw if we want this.


Adamits (approved):
Yeah this is a separate warning, but due to the same issue. Suppressing it seems good.

LGTM

PR #118, opened 2023-07-20
Updates docs on hyperparameter sweeps

kylebgorman:
We now recommend users just grab the CSV themselves using the web API.
This seems like it'll be good enough for virtually everyone.

Note that #117 has leaked into this. I'll do that one first.


Adamits (approved):
Yeah using the python API seems unnecessary.

LGTM

Issue #119, opened 2023-07-21
Validation Accuracy Aggregation


Adamits:
Currently, our `validation_step` method on the `BaseEncoderDecoder` computes a per batch accuracy and aggregates them at the [end of each epoch](https://github.com/CUNY-CL/yoyodyne/blob/master/yoyodyne/models/base.py#L312). Because of this, we get a macro average accuracy that will depend on the batch size.

I noticed something must be strange when using evaluation sets of size 1000 and getting validation accuracies to many decimal places (like `0.9247395...`). I think we probably want to accumulate raw counts of correct/incorrect dev samples per batch, and then aggregate those into an accuracy at the end of each epoch. 

The impact of this should be small, but still, I believe we are getting slightly incorrect accuracies according to the expected micro accuracy.

kylebgorman:
Yes, I agree we want micro-accuracy not macro even though the only way I think they can differ is w.r.t. a partial batch.

Adamits:
Oh yeah good point. I guess we could also get some loss of precision too. Anyway, micro certainly seems preferable.

PR #120, opened 2023-07-21
WIP: Fixes eval accuracy on epoch

Adamits:
Addresses #119

Marking this as a WIP as it's a little clunky, but wanted to get a PR up. Probably we should refactor the evaluator so the semantics of getting correct/total counts is clearer?

This fixes an issue where we report a macro accuracy over per-batch accuracies on the validation set, rather than a micro accuracy over all predictions.

### DIFF
--- a/yoyodyne/evaluators.py
+++ b/yoyodyne/evaluators.py
@@ -1,5 +1,8 @@
 """Evaluators."""
 
+from __future__ import annotations
+from dataclasses import dataclass
+
 import torch
 from torch.nn import functional
 
@@ -8,16 +11,55 @@ class Error(Exception):
     pass
 
 
+@dataclass
+class EvalItem:
+    num_correct: int
+    num_predicted: int
+
+    @property
+    def accuracy(self) -> float:
+        """Computes the accuracy."""
+        return self.num_correct / self.num_predicted
+
+    def __add__(self, other_eval: EvalItem) -> EvalItem:
+        """Adds two EvalItems by summing along both attributes.
+
+        Args:
+            other_eval (EvalItem): The other eval item to add to self.
+
+        Returns:
+            EvalItem.
+        """
+        return EvalItem(
+            self.num_correct + other_eval.num_correct,
+            self.num_predicted + other_eval.num_predicted,
+        )
+
+    def __radd__(self, start_val: int) -> EvalItem:
+        """The initial add when calling `sum` on an iterable of EvalItems.
+        A zero values start value is provided.
+
+        Args:
+            start_val (int): A zero for calling the first add in an iterable.
+
+        Returns:
+            EvalItem.
+        """
+        return EvalItem(
+            self.num_correct + start_val, self.num_predicted + start_val
+        )
+
+
 class Evaluator:
     """Evaluates predictions."""
 
-    def val_accuracy(
+    def evaluate(
         self,
         predictions: torch.Tensor,
         golds: torch.Tensor,
         end_idx: int,
         pad_idx: int,
-    ) -> float:
+    ) -> EvalItem:
         """Computes the exact word match accuracy.
 
         Args:
@@ -38,14 +80,14 @@ class Evaluator:
         _, predictions = torch.max(predictions, dim=2)
         # Finalizes the predictions.
         predictions = self.finalize_predictions(predictions, end_idx, pad_idx)
-        return self.accuracy(predictions, golds, pad_idx)
+        return self.get_eval_item(predictions, golds, pad_idx)
 
     @staticmethod
-    def accuracy(
+    def get_eval_item(
         predictions: torch.Tensor,
         golds: torch.Tensor,
         pad_idx: int,
-    ) -> float:
+    ) -> EvalItem:
         if predictions.size(1) > golds.size(1):
             predictions = predictions[:, : golds.size(1)]
         elif predictions.size(1) < golds.size(1):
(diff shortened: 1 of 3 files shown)


yoyodyne/evaluators.py
```diff
@@ -60,7 +60,7 @@ def accuracy(
         )[0].size()[0]
         # Gets the batch size (total_count).
         total_count = predictions.size(0)
-        return corr_count / total_count
+        return corr_count, total_count, corr_count / total_count
```
kylebgorman, line 63: You probably need to change the return type too.

I wonder if this should be a sort of `struct` or lightweight class rather than a tuple.

It looks like the third element of the tuple is never used.

yoyodyne/models/base.py
```diff
@@ -294,7 +294,7 @@ def validation_step(
             greedy_predictions, 2, 0, target_padded.size(1)
         )
         loss = self.loss_func(greedy_predictions, target_padded)
-        return {"val_accuracy": accuracy, "val_loss": loss}
+        return {"val_correct": correct, "val_total": total, "val_loss": loss}
```
kylebgorman, line 297: does this end up logged on the TQDM slider thingy? if so can we suppress it?

yoyodyne/models/base.py
```diff
@@ -305,11 +305,17 @@ def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:
...
-        keys = validation_step_outputs[0].keys()
-        # Shortens name to do comprehension below.
-        # TODO: there has to be a more elegant way to do this.
-        V = validation_step_outputs
-        metrics = {k: sum([v[k] for v in V]) / len(V) for k in keys}
+        num_steps = len(validation_step_outputs)
+        avg_val_loss = (
+            sum([v["val_loss"] for v in validation_step_outputs]) / num_steps
+        )
+        num_correct = sum([v["val_correct"] for v in validation_step_outputs])
```
kylebgorman, line 312: You don't need the `[]` inside the `sum` expression:

```python
num_correct = sum(v["val_correct"] for v in validation_step_outputs)
```

And same thing on the follwing lnie.

kylebgorman (commented):
Just some notes on implementation while you think about this.

yoyodyne/models/base.py
Adamits: It doesn't. Only if we call `self.log`

Adamits:
Ok updated to use a `dataclass` for eval. I realize the custom adding is possibly more confusing than to be worth it, but thought it was fun so was hoping you could take a look :). I can also revert, and use the `dataclass` strictly as a struct, and just compute the numerator/denominator sums in the `validation_epoch_end` method.

yoyodyne/evaluators.py
```diff
@@ -1,5 +1,8 @@
 """Evaluators."""
 
+from __future__ import annotations
+from dataclasses import dataclass
```
kylebgorman, line 4: Please import this at the module level like we do elsewhere.

yoyodyne/evaluators.py
```diff
@@ -8,16 +11,55 @@ class Error(Exception):
...
+            EvalItem.
+        """
+        return EvalItem(
+            self.num_correct + other_eval.num_correct,
+            self.num_predicted + other_eval.num_predicted,
+        )
+
+    def __radd__(self, start_val: int) -> EvalItem:
+        """The initial add when calling `sum` on an iterable of EvalItems.
+        A zero values start value is provided.
```
kylebgorman, line 40: It doesn't look like this statement is true.

yoyodyne/models/base.py
```diff
@@ -305,11 +305,16 @@ def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:
...
+        num_steps = len(validation_step_outputs)
+        avg_val_loss = (
+            sum([v["val_loss"] for v in validation_step_outputs]) / num_steps
+        )
+        epoch_eval = sum(v["val_eval_item"] for v in validation_step_outputs)
+        metrics = {
+            "val_loss": avg_val_loss,
+            "val_accuracy": epoch_eval.accuracy,
+        }
+
```
kylebgorman, line 317: Please remove blank line.

kylebgorman (commented):
I'm cool with it. A few nits then go for it.

yoyodyne/evaluators.py
Adamits: Maybe it is my comment that is confusing, but here is what I thought happens when sum is called on an iterable of `EvalItem`s:

1. (`sum`)[https://docs.python.org/3/library/functions.html#sum] calls __add__ between the first `EvalItem` and the `start` values, which defaults to 0.
2. `__add__` for `int` raises NotImplementedError because `int` does not define `__add__` with `EvalItem`, and sum first calls `__add__` for the left type. 
3. Some python magic happens and it then checks for the "reverse add", that is, `__radd__` on the right type. I have defined it above so that it takes the default `0` and adds it to numerator and denominator, behaving very similarly to adding two `EvalItems`.

Note that removing this `__radd__` implementation causes:

```
TypeError: unsupported operand type(s) for +: 'int' and 'EvalItem'
```

to be raised.

kylebgorman: > Maybe it is my comment that is confusing, but here is what I thought happens when sum is called on an iterable of `EvalItem`s:
> 
> 1. [sum](https://docs.python.org/3/library/functions.html#sum) calls **add** between the first `EvalItem` and the `start` value, which defaults to 0.
> 2. `__add__` for `int` raises NotImplementedError because `int` does not define `__add__` with `EvalItem`, and sum first calls `__add__` for the left type.

Minor correction is that I think it `return NotImplemented` which is the new way to do this.

> 3. Some python magic happens and it then checks for the "reverse add", that is, `__radd__` on the right type. I have defined it above so that it takes the default `0` and adds it to numerator and denominator, behaving very similarly to adding two `EvalItems`.

Yeah that's all right. I just didn't find the comment helpful, since this is true of the `sum`/`radd` interaction in general?

Adamits: Yeah good point. The comment is confusing. Does the updated version help at all? Since this behavior is built-in to python, I guess we do not need to document too much of it in the method implementation.

kylebgorman (approved):
LGTM.

PR #125, opened 2023-08-02
Removes tied vocabulary flag

### LINKED ISSUE #124: Tied vocabulary flag: no-op?
Correct me if I'm wrong, but: the `--tied_vocabulary` flag has no effect except in the construction of the index (and this has no downstream impact).

It isn't "tied" in the stronger sense that source and target symbols share an embedding, so that a source "a" and a target "a" receive the same representation. Is that what was intended?

If this is correct, I propose we either:

* remove the flag, so as not to confuse people
* or, cause it to have some effect.

kylebgorman:
This can be restored at a later date when it is made to do something.

Closes #124.


Adamits (approved):
Thanks for making the update, LGTM!

PR #131, opened 2023-08-14
Adds documentation for multi-batch accumulation.

### LINKED ISSUE #129: Documentation for accumulating multiple batches
PTL allows us to simulate batch sizes larger than what will fit in your accelerator's core by accumulate gradients from multiple ("mini")batches. For instance to simulate an effective batch size of 8192 with smaller batches of 1024 one could do `--batch_size 1024 --accumulate_grad_batches 8`. Alternatively, if batch size is not known at runtime (e.g., because one is doing hyperparameter optimization) one can do `--max_batch_size 1024` and anything larger than that will be simulated like so. 

We should document this briefly.

kylebgorman:
Closes #129.


Adamits (approved):
LGTM!

I will plan to extend the simulating large batches section after we merge my PR with the "dynamic" accumulation stuff.

Issue #111, opened 2023-07-17
Pointer-generator behavior with features


kylebgorman:
Reporting some funky behavior with the pointer-generator with features:

1. With `--arch pointer_generator_lstm` and features enabled (i.e., `--features_col 3` or something other than the default `0`), I get the following report:

```
Model: pointer-generator
Encoder: LSTM
Decoder: attentive LSTM
LOCAL_RANK: 0 - CUDA_VISIBLE_DEVICES: [0]

  | Name                   | Type                  | Params
-----------------------------------------------------------------
0 | loss_func              | NLLLoss               | 0     
1 | dropout_layer          | Dropout               | 0     
2 | source_encoder         | LSTMEncoder           | 333 K 
3 | decoder                | LSTMAttentiveDecoder  | 283 K 
4 | classifier             | Linear                | 12.0 K
5 | log_softmax            | LogSoftmax            | 0     
6 | generation_probability | GenerationProbability | 601   
-----------------------------------------------------------------
```

Since it doesn't list a feature encoder in either place, this makes me think it has ignored features even though I have explicitly requested them.

2. With `--arch pointer_generator_lstm --source_encoder lstm --features_encoder lstm` (which should be the same thing) we get a crash:

```
  File "/home/kbg/.miniconda3/lib/python3.10/site-packages/yoyodyne/models/pointer_generator.py", line 370, in forward
    return predictions
UnboundLocalError: local variable 'predictions' referenced before assignment
```

The reason from this should be clear from [the code](https://github.com/CUNY-CL/yoyodyne/blob/28ff0e5bc34a7402332b020599cce8d910079175/yoyodyne/models/pointer_generator.py#L345): the branch that begins at line 345 doesn't define `predictions`.

3.  `--arch pointer_generator_lstm --source_encoder lstm --features_encoder linear` works, though after hill climbing for a while both losses go `nan` (e.g., on our Polish data).

4. Same story as (3) with `--arch pointer_generator_lstm --source_encoder transformer --features_encoder linear`.
 
I am assigning this to @bonham79; I think the fix will be quite small.

kylebgorman:
> I am testing pointer generator for #80 and sneaking the `predictions` fix in there (unless it turns out to be more complicated). 

+1.

> However, I will comment that it is very unintuitive that setting the features column, but no features encoder tells the model to simply ignore features.

Yeah, got a different way to do that? 

[...]

PR #133, opened 2023-08-15
Loss function updates

Adamits:
- Changes loss functions to use `CrossEntropyLoss`.
- Overrides this for `PointerGenerator`, which needs to use `NLLLoss`
- Patches the main bug in #111

### DIFF
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -371,7 +371,7 @@ class PointerGeneratorLSTMEncoderDecoder(lstm.LSTMEncoderDecoder):
         return predictions
 
     def _get_loss_func(
-        self
+        self,
     ) -> Callable[[torch.Tensor, torch.Tensor], torch.Tensor]:
         """Returns the actual function used to compute loss.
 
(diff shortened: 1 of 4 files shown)


yoyodyne/models/pointer_generator.py
```diff
@@ -358,18 +358,75 @@ def forward(
...
-                    source_encoded,
-                    batch.source.mask,
-                    batch.source.padded,
-                    last_hiddens,
-                    self.teacher_forcing if self.training else False,
-                    features_enc=features_encoded,
-                    features_mask=batch.features.mask,
-                    target=batch.target.padded if batch.target else None,
-                )
+            predictions = self.decode(
```
kylebgorman, line 361: Is this a fix of sorts for #111?

yoyodyne/models/pointer_generator.py
```diff
@@ -358,18 +358,75 @@ def forward(
...
+                features_mask=batch.features.mask,
+                target=batch.target.padded if batch.target else None,
+            )
         return predictions
 
+    def _get_loss_func(
+        self,
+    ) -> Callable[[torch.Tensor, torch.Tensor], torch.Tensor]:
+        """Returns the actual function used to compute loss.
+
```
kylebgorman, line 377: Could you add some brief documentation here that we're overriding the base version because we need NLLLoss for reasons? It could go into the docstring or even just a comment at line 381-382 or so.

kylebgorman (approved):
LGTM, suggested slight documentation though.

yoyodyne/models/pointer_generator.py
Adamits: Yes. I should mark it in the PR description

PR #134, opened 2023-08-16
Adds documentation for schedulers

### LINKED ISSUE #128: Documentation for schedulers
We have three schedulers, each with their own arguments. They probably merit a paragraph or two in the README describing how they're used. I am assigning this to myself.

kylebgorman:
Closes #128.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
-# Yoyodyne 🪀
+Yoyodyne 🪀
+==========
 
 [![PyPI
 version](https://badge.fury.io/py/yoyodyne.svg)](https://pypi.org/project/yoyodyne)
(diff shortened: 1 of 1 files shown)


kylebgorman:
My attempt to reflow tweaked some of the headers but it's a no-op change. Sorry about that.

Adamits (approved):
Did not know about doing markdown headers this way.

LGTM.

kylebgorman:
> Did not know about doing markdown headers this way.

My reflow-er (I use `pandoc -f markdown -t markdown --columns=80` as a shell alias) does that automatically.  The following two are equivalent in Markdown: 

```
Foo
===

# Foo
```

PR #135, opened 2023-08-16
WIP: Adds transformer pointer generator.

Adamits:
- Adds a pointer generator with a transformer encoder-decoder.
- Introduces a WIP TransformerDecoder that sets a hook to override the torch implementations default behavior of not returning the MHA weights.
- Uses the MHA weights to get pointer probabilities
- Currently does NOT handle features. I believe this will required a second multiheaded attention from each decoder step to the encoded source tokens.  I will note that this may require subclassing the torch `TransformerDecoder` to implement, which will make my work with the hook/`patch_attention` unneeded (we could just hardcode the behavior directly into the subclassed decoder). I have not thought about this much yet though.

This is a WIP, but has a working implementation as far as I can tell, for a transformer encoder-decoder when there are not features. teacher-forced training works by getting all pointer and generator distributions for each target, and then combining them in one go. Greedy decoding for inference works more like the LSTM, where we first decode a single target, then combine pointer and generator probabilities, and then iteratively do it again until we hit EOS or `max_target_length`.

Looking for any obvious mistakes or feedback on the general structure.

### DIFF
--- a/yoyodyne/models/__init__.py
+++ b/yoyodyne/models/__init__.py
@@ -4,7 +4,10 @@ import argparse
 
 from .base import BaseEncoderDecoder
 from .lstm import AttentiveLSTMEncoderDecoder, LSTMEncoderDecoder
-from .pointer_generator import PointerGeneratorLSTMEncoderDecoder
+from .pointer_generator import (
+    PointerGeneratorLSTMEncoderDecoder,
+    PointerGeneratorTransformerEncoderDecoder,
+)
 from .transducer import TransducerEncoderDecoder
 from .transformer import TransformerEncoderDecoder
 
@@ -26,6 +29,7 @@ def get_model_cls(arch: str) -> BaseEncoderDecoder:
         "attentive_lstm": AttentiveLSTMEncoderDecoder,
         "lstm": LSTMEncoderDecoder,
         "pointer_generator_lstm": PointerGeneratorLSTMEncoderDecoder,
+        "pointer_generator_transformer": PointerGeneratorTransformerEncoderDecoder,
         "transducer": TransducerEncoderDecoder,
         "transformer": TransformerEncoderDecoder,
     }
@@ -67,6 +71,7 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             "attentive_lstm",
             "lstm",
             "pointer_generator_lstm",
+            "pointer_generator_transformer",
             "transducer",
             "transformer",
         ],
--- a/yoyodyne/models/modules/transformer.py
+++ b/yoyodyne/models/modules/transformer.py
@@ -1,7 +1,7 @@
 """Transformer model classes."""
 
 import math
-from typing import Optional
+from typing import Optional, Tuple
 
 import torch
 from torch import nn
@@ -71,6 +71,26 @@ class PositionalEncoding(nn.Module):
         pad_mask = symbols.ne(self.pad_idx).unsqueeze(2)
         out = out * pad_mask
         return out
+
+
+class AttentionOutput:
+    """Tracks an attention output during the forward pass.
+    
+    This object can be passed into a hook to the Transformer forward pass
+    in order to modify it's behavior such that certaina ttention weights are 
+    returned, and then stored in self.outputs."""
+    def __init__(self):
+        self.outputs = []
+
+    def __call__(self, module, module_in, module_out):
+        """Stores the second return argument of `module`.
+        
+        This is intended to be called on a multiehaded attention, which returns
+        both the contextualized representation, and the attention weights."""
+        self.outputs.append(module_out[1])
+
+    def clear(self):
+        self.outputs = []
 
 
 class TransformerModule(base.BaseModule):
@@ -286,7 +306,7 @@ class TransformerDecoder(TransformerModule):
             memory_key_padding_mask=source_mask,
             tgt_key_padding_mask=target_mask,
         )
-        return base.ModuleOutput(output)
+        return base.ModuleOutput(output, embeddings=target_embedding)
 
     def get_module(self) -> nn.TransformerDecoder:
         decoder_layer = nn.TransformerDecoderLayer(
@@ -323,3 +343,40 @@ class TransformerDecoder(TransformerModule):
     @property
     def name(self) -> str:
         return "transformer"
+
+
+class TransformerDecoderWithMhaWeights(TransformerDecoder):
+    """TransformerDecoder with an `attention_output` property.
+    
+    `attention_output` tracks the output of multiheaded attention from each decoder step
+    wrt the encoded input. This is achieved with a hook into the forward pass.
+
+    After:
+        Hook idea is taken from 
+        https://gist.github.com/airalcorn2/50ec06517ce96ecc143503e21fa6cb91"""
+
+    def __init__(self, *args, **kwargs):
+        """Initializes the TransformerDecoderWithMhaWeights object."""
+        super().__init__(*args, **kwargs)
+        # Call this to get the actual cross attentions
+        self.attention_output = AttentionOutput()
+        # multihead_attn refers to the attention from decoder to encoder.
+        self.patch_attention(self.module.layers[-1].multihead_attn)
+        self.hook_handle = self.module.layers[-1].multihead_attn.register_forward_hook(
+            self.attention_output
+        )
+
+    def patch_attention(self, attention_module: torch.nn.Module) -> None:
+        """Wraps a module's forward pass such that `need_weights` is True.
+
+        Args:
+            attention_module (torch.nn.Module): The module from which we want
+                to track multiheaded attention weights.
+        """
+        forward_orig = attention_module.forward
+
+        def wrap(*args, **kwargs):
+            kwargs["need_weights"] = True
+            return forward_orig(*args, **kwargs)
+
+        attention_module.forward = wrap
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -7,7 +7,7 @@ import torch
 from torch import nn
 
 from .. import data
-from . import lstm, modules
+from . import lstm, modules, transformer
 
 
 class Error(Exception):
@@ -21,7 +21,7 @@ class GenerationProbability(nn.Module):
 
     W_attention: nn.Linear
     W_hs: nn.Linear
-    W_inp: nn.Linear
+    W_emb: nn.Linear
     bias: nn.Parameter
 
     def __init__(
@@ -35,35 +35,38 @@ class GenerationProbability(nn.Module):
             attention_size (int): dimensions of combined encoder attentions.
         """
         super().__init__()
-        self.W_attention = nn.Linear(attention_size, 1, bias=False)
+        self.W_emb = nn.Linear(embedding_size, 1, bias=False)
         self.W_hs = nn.Linear(hidden_size, 1, bias=False)
-        self.W_inp = nn.Linear(embedding_size, 1, bias=False)
+        self.W_attention = nn.Linear(attention_size, 1, bias=False)
         self.bias = nn.Parameter(torch.Tensor(1))
         self.bias.data.uniform_(-self.stdev, self.stdev)
 
     def forward(
         self,
-        h_attention: torch.Tensor,
-        decoder_hs: torch.Tensor,
-        inp: torch.Tensor,
+        attention_context: torch.Tensor,
+        decoder_hiddens: torch.Tensor,
+        target_embeddings: torch.Tensor,
     ) -> torch.Tensor:
-        # TODO(Adamits): improve documentation.
-        """Computes Wh * ATTN_t + Ws * HIDDEN_t + Wy * Y_{t-1} + b.
+        """Computes the generation probability
+        
+        This is a function of the context vectors, decoder hidden states, and
+        target embeddings, where each is first mapped to a scalar value by a learnable
+        weight matrix.
 
         Args:
-            h_attention (torch.Tensor): combined context vector over source and
-                features of shape B x 1 x attention_size.
-            decoder_hs (torch.Tensor): decoder hidden state of shape
-                B x 1 x hidden_size.
-            inp (torch.Tensor): decoder input of shape B x 1 x embedding_size.
+            attention_context (torch.Tensor): combined context vector over source and
+                features of shape B x sequence_length x attention_size.
+            decoder_hiddens (torch.Tensor): decoder hidden state of shape
+                B x sequence_length x hidden_size.
+            target_embeddings (torch.Tensor): decoder input of shape B x sequence_length x embedding_size.
 
         Returns:
             (torch.Tensor): generation probability of shape B.
         """
-        # -> B x 1 x 1.
-        p_gen = self.W_attention(h_attention) + self.W_hs(decoder_hs)
-        p_gen += self.W_inp(inp) + self.bias.expand(h_attention.size(0), 1, -1)
-        # -> B.
+        # -> B x sequence_length x 1.
+        p_gen = self.W_attention(attention_context) + self.W_hs(decoder_hiddens)
+        p_gen += self.W_emb(target_embeddings) + self.bias.expand(attention_context.size(0), 1, -1)
+        # -> B x sequence_length.
         p_gen = torch.sigmoid(p_gen.squeeze(1))
         return p_gen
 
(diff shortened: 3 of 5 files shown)


yoyodyne/models/modules/transformer.py
```diff
@@ -73,6 +73,26 @@ def forward(
         return out
 
 
+class AttentionOutput:
+    """Tracks an attention output during the forward pass.
+    
+    This object can be passed into a hook to the Transformer forward pass
+    in order to modify it's behavior such that certaina ttention weights are
```
kylebgorman, line 80: typos: `it's` and `certaina ttention`

yoyodyne/models/modules/transformer.py
```diff
@@ -73,6 +73,26 @@ def forward(
...
+    This object can be passed into a hook to the Transformer forward pass
+    in order to modify it's behavior such that certaina ttention weights are 
+    returned, and then stored in self.outputs."""
+    def __init__(self):
+        self.outputs = []
+
+    def __call__(self, module, module_in, module_out):
+        """Stores the second return argument of `module`.
+        
+        This is intended to be called on a multiehaded attention, which returns
```
kylebgorman, line 88: typo: `multihaded`

yoyodyne/models/modules/transformer.py
```diff
@@ -73,6 +73,26 @@ def forward(
...
+        both the contextualized representation, and the attention weights."""
+        self.outputs.append(module_out[1])
+
+    def clear(self):
+        self.outputs = []
```
kylebgorman, line 93: Why not just `self.outputs.clear()`? Probably slightly cheaper, certainly clearer to me.

yoyodyne/models/pointer_generator.py
```diff
@@ -444,3 +447,238 @@ def _reshape_hiddens(
...
+        )
+        # A matrix of 'context' vectors from applying attention 
+        # to the encoder representations wrt each decoder step.
+        context = torch.bmm(mha_outputs, encoder_outputs)
+        # Probability of generating (from output_probs).
+        gen_probs = self.generation_probability(
+            context, decoder_output, tgt_embeddings
+        )
+        ptr_scores = (1 - gen_probs) * ptr_probs
+        gen_scores = gen_probs * output_probs
```
kylebgorman, line 549: consider using `.mul_` here

yoyodyne/models/pointer_generator.py
```diff
@@ -444,3 +447,238 @@ def _reshape_hiddens(
...
+        return gen_scores + ptr_scores
```
kylebgorman, line 550: consider using `.add_` here

yoyodyne/models/pointer_generator.py
```diff
@@ -444,3 +447,238 @@ def _reshape_hiddens(
...
+                output = self._decode_greedy(
+                    source_encoded,
+                    batch.source.mask,
+                    batch.source.padded,
+                    batch.target.padded if batch.target else None,
+                )
+        return output
+    
+    @property
+    def name(self) -> str:
```
kylebgorman, line 683: This can be two words, so "pointer-generator transformer" seems fine. Also, you should change the pointer generator LSTM's `name` property to return `pointer-generator LSTM` then I think.

kylebgorman (commented):
Other than style stuff (I think you know what most of this is, I can call out particular things later) and some notes below, good idea.

I sort of forgot that one could have a true pointer-generator transformer seq2seq model separate from what kind of encoders you have...

Adamits:
Thought a bit about adding features to this. I think the two clearest options are:

1. *Best analog to our LSTM implementation* Modify the `TransformerDecoderLayer`s to have a second multiheaded attention that attends to the features. Concatenate the result of this to the lemma attention. Note that we then need to choose a number of attention heads for the features attention, and need to make our model bigger to accomodate this concatenation (recall the transformer represents the "context" vector as the concatenation of all attention ehad results). 
2. Add the features to the input sequence (as in the regular transformer encoder decoder), and mask those positions in the (averaged) multiheaded attention that is used to compute pointer probabilities.

Do you have any thoughts/preference?

kylebgorman:
The first one seems more obvious to me. You could then just make it so you have `--source_attention_heads` and `--features_attention_heads` (and get rid of the old `--attention_heads`). The other one seems fine too though honestly. Maybe @bonham79 has an opinion.

Adamits:
New commit obviously needs lots of cleanup but:

- This implements 1) from my previous comment
- To achieve this, I subclassed `TransformerDecoderLayer` and `TransformerDecoder`. The layer subclass has a second multiheaded_attention that attends just to the features.  
- In order to combine feature and source attention without breaking the feedforwad layer, nor the residual connection, I have added two linear layers: 1 to map symbol attention to half of the hidden_size, and one to map feature attention to half of the hidden size, and then these are concatenated.

Right now this is hacky in the sense that I probably need to enforce some stuff in the embedding/hidden size configurations, and the code to choose whether to use features or not is sloppy and redundant. Obviously I also need to cleanup documentation. I ran this with a sanity checking script below, and it was converging pretty quickly.

```
#!/bin/bash

LANGUAGE=english

readonly ROOT="${USER}/nlp-projects/morphology"
readonly DATA="${ROOT}/conll2017/all/task1/"
readonly TRAIN="${DATA}/${LANGUAGE}-train-high"
readonly DEV="${DATA}/${LANGUAGE}-dev"

# for ARCH in attentive_lstm lstm pointer_generator_lstm pointer_generator_transformer transformer transducer; do
ARCH=pointer_generator_transformer
RESULTS_PATH="./results/${LANGUAGE}/${ARCH}"

yoyodyne-train \
    --arch "${ARCH}" \
    --encoder_layers 2 \
    --decoder_layers 2 \
    --embedding_size 64 \
    --hidden_size 64 \
    --target_col 2 \
    --features_col 3 \
    --features_encoder_arch transformer \
    --accelerator cpu \
    --experiment "${LANGUAGE}" \
    --train "${TRAIN}" \
    --val "${DEV}" \
    --label_smoothing 0.1 \
    --batch_size 32 \
    --max_epochs 4 \
    --save_top_k 1 \
    --check_val_every_n_epoch 1 \
    --model_dir "${RESULTS_PATH}" ;
```

kylebgorman:
Cool. I'm mostly following along. Let me know when you're ready for review.

yoyodyne/models/pointer_generator.py
Adamits: Is it faster?

Adamits: Or are you suggesting because it is in-place?

kylebgorman: Yeah, I am suggesting in-place because it shouddl be faster (clearer too maybe?). the way you wrote it:

* creates a new tensor with (1 - gen_probs) * ptr_probs in it
* assigns it to `ptr_scores`, which triggers atomic increments/decremetns in the ref counter. since the ref count for the older data that `ptr_probs` pointed to should now be zero, it will be marked for (but not yet undergo) GC on the next cycle
* ultimately the GC will also delete the old ptr_probs data

Only the tensor multiplication is necessary if you use the in-place one.

yoyodyne/models/pointer_generator.py
kylebgorman: Same deal, this is an optimization to avoid thrashing memory/ref counting/GC.

yoyodyne/models/pointer_generator.py
Adamits: Ok. I just need to find the op for a vector * scalar. I get an error with `_mul` because they are not the same shape.

Adamits: I am wrong, I just was calling mul_ on the vector first.

Adamits: These in-place operations seem to break backprop, though I do not understand the error message.

Adamits: I assume its b/c we are overwriting something on those variables that needs to be stored in the CG for backprop, and no longer is?

kylebgorman: FWIW we have used these elsewhere in similar contexts, I believe, and no related bugs yet. If this really does break it I wonder if we're doing "silly scoping" somewhere---i.e., something isn't being passed in or something like that.

Adamits: > something isn't being passed in or something like that.

Can you explain more here? I can look into this a bit harder if it is a symptom of another issue.

kylebgorman: For instance maybe we wrote something that is conceptually similar to:

```python
def foo(bar):
   return baz + bar   # where is baz from? 
```

Barring that I can't imagine how this would influence the computation graph.

kylebgorman: You can always just leave this as a `# TODO: consider in-place operations; using .mul_ here seems to cause issues.` or something like that.

Adamits: Ok, I will do that for now.

yoyodyne/models/__init__.py
```diff
@@ -26,6 +29,7 @@ def get_model_cls(arch: str) -> BaseEncoderDecoder:
         "attentive_lstm": AttentiveLSTMEncoderDecoder,
         "lstm": LSTMEncoderDecoder,
         "pointer_generator_lstm": PointerGeneratorLSTMEncoderDecoder,
+        "pointer_generator_transformer": PointerGeneratorTransformerEncoderDecoder,
```
Adamits, line 32: Flake8 complains about this. If I put the model on a new line, black complains. @kylebgorman do you know the solution here?

Adamits:
Ready for review! Notice my inline comment about why CI is failing.

yoyodyne/models/__init__.py
kylebgorman: Just use a `  # noqa` on the offending lines.

yoyodyne/defaults.py
```diff
@@ -14,6 +14,7 @@
 # Architecture arguments.
 ARCH = "attentive_lstm"
 ATTENTION_HEADS = 4
```
kylebgorman, line 16: I think for clarity I'd prefer `SOURCE_ATTENTION_HEADS` and `FEATURES_ATTENTION_HEADS` (note for consistency I am always using "features" and not "feature"). And same for the flags. We do this with encoder modules too---distinguish between source and features even though sometimes source includes features because concatenation...

yoyodyne/models/modules/transformer.py
```diff
@@ -73,6 +73,42 @@ def forward(
...
+class AttentionOutput:
+    """Tracks an attention output during the forward pass.
+
+    This object can be passed into a hook to the Transformer forward pass
+    in order to modify its behavior such that certain attention weights are
+    returned, and then stored in self.outputs."""
+
+    def __init__(self):
+        """Initializes an AttentionOutput."""
+        self.outputs = []
```
kylebgorman, line 86: maybe declare this type for the class

yoyodyne/models/modules/transformer.py
```diff
@@ -184,6 +220,242 @@ def name(self) -> str:
...
+class TransformerDecoderSeperateFeatures(nn.TransformerDecoder):
+    """A Transformer decoder with seperate features.
+
+    Adding seperate features into the transformer stack is implemented with
+    TransformerDecoderLayerSeperateFeatures layers.
+    """
+
+    def forward(
+        self,
+        tgt: torch.Tensor,
```
kylebgorman, line 232: For consistency with what we do elsewhere I suggest `target_*` rather than `tgt_`.

yoyodyne/ptr_gen_trm_stesp.log
kylebgorman, line 1: remove this

kylebgorman (commented):
Added some comments. A bunch of "seperate" typos (should be "separate"). One of them looked like it might be not-just-no-op actually.

yoyodyne/ptr_gen_trm_stesp.log
Adamits: Oops!

yoyodyne/defaults.py
Adamits: It may be that `SOURCE_ATTENTION_HEADS` is the best name, but wanted to clarify that we reuse this variable for self attention in both the encoder and decoder, as well as for the "cross-attention" (can't remember the canonical name for this) from encoder to decoder. I think using the same number for all of these is standard. SOURCE could imply that it is specific to the encoder, or just the "cross" attention?

kylebgorman: I see and agree. Not a bad idea to comment that somewhere in the source (maybe not here though).

Adamits:
Not quite sure what happened. I had pulled your commit, but still somehow needed to then merge either my next commit or yours back in?

Anyway, I think this is in a good state now given your notes. I did do the `ATTENTION_HEADS -> SOURCE_ATTENTION_HEADS` change. It seems ok for now?

kylebgorman:
I made some ninja style edits so you had to do a merge those back in. Will
take a final look soon.

yoyodyne/models/pointer_generator.py
bonham79: pytorch is annoyingly inconsistent with in place operation and maintaining position in the computation graph (also some weird stuff regarding operation managed by ptl AMP for 16 vs 32 bit precision). Safer is usually to just cast to a new tensor (pytorch is usually reliable with only duplicating data when needed).

bonham79:
> Thought a bit about adding features to this. I think the two clearest options are:
> 
>     1. _Best analog to our LSTM implementation_ Modify the `TransformerDecoderLayer`s to have a second multiheaded attention that attends to the features. Concatenate the result of this to the lemma attention. Note that we then need to choose a number of attention heads for the features attention, and need to make our model bigger to accomodate this concatenation (recall the transformer represents the "context" vector as the concatenation of all attention ehad results).
> 
>     2. Add the features to the input sequence (as in the regular transformer encoder decoder), and mask those positions in the (averaged) multiheaded attention that is used to compute pointer probabilities.
> 

Late on the draw but  simply adding features to input seems the most unobtrusive result. (In my mind it works similar to positional encoding, so don't see why similar can't be done here).

Adamits:
> > Thought a bit about adding features to this. I think the two clearest options are:
> > ```
> > 1. _Best analog to our LSTM implementation_ Modify the `TransformerDecoderLayer`s to have a second multiheaded attention that attends to the features. Concatenate the result of this to the lemma attention. Note that we then need to choose a number of attention heads for the features attention, and need to make our model bigger to accomodate this concatenation (recall the transformer represents the "context" vector as the concatenation of all attention ehad results).
> > 
> > 2. Add the features to the input sequence (as in the regular transformer encoder decoder), and mask those positions in the (averaged) multiheaded attention that is used to compute pointer probabilities.
> > ```
> 
> Late on the draw but simply adding features to input seems the most unobtrusive result. (In my mind it works similar to positional encoding, so don't see why similar can't be done here).

Sure, we should add this option for both Transformer and LSTM. I was hypothesizing that its probably trivial to learn not to 'point' to features. I just talked to a lab mate who uses someone else's implementation of transformer pointer-generator apparently, and indeed thinks concatenating the features works fine, and it doesn't really seem to matter if you mask or not.

kylebgorman:
> Sure, we should add this option for both Transformer and LSTM. I was hypothesizing that its probably trivial to learn not to 'point' to features. I just talked to a lab mate who uses someone else's implementation of transformer pointer-generator apparently, and indeed thinks concatenating the features works fine, and it doesn't really seem to matter if you mask or not.

Probably true with enough (for some value of "enough") data but seems like a waste of parameters.

We don't necessarily need to support concatenation at all since the user can always concatenate TSV columns, but we do need to support separate encoders since there's no UNIX trick for that.

Adamits:
Does this PR require updates from me?

yoyodyne/models/modules/transformer.py
```diff
@@ -184,8 +223,248 @@ def name(self) -> str:
...
+                output,
+                memory,
+                features_memory,
+                target_mask=target_mask,
+                memory_mask=memory_mask,
+                features_memory_mask=features_memory_mask,
+                target_key_padding_mask=target_key_padding_mask,
+                memory_key_padding_mask=memory_key_padding_mask,
+            )
+        if self.norm is not None:
```
kylebgorman, line 279: possibly dumb question, does this need to be nested inside the for-loop that begins at line 268?

yoyodyne/models/modules/transformer.py
```diff
@@ -184,8 +223,248 @@ def name(self) -> str:
...
+            x,
+            mem,
+            mem,
+            attn_mask=attn_mask,
+            key_padding_mask=key_padding_mask,
+            # FIXME: Introduced in torch 2.0
+            # is_causal=is_causal,
+            need_weights=False,
+        )[0]
+        return self.dropout2(x)
```
kylebgorman, line 463: Another dumb question: where is the `dropout2` object or method implemented? Did I just miss it?

yoyodyne/models/modules/transformer.py
```diff
@@ -184,8 +223,248 @@ def name(self) -> str:
...
+                features_memory_mask=features_memory_mask,
+                target_key_padding_mask=target_key_padding_mask,
+                memory_key_padding_mask=memory_key_padding_mask,
+            )
+        if self.norm is not None:
+            output = self.norm(output)
+        return output
+
+
+class TransformerDecoderLayerSeperateFeatures(nn.TransformerDecoderLayer):
```
kylebgorman, line 284: typo: `Seperate` should be `Separate`.

yoyodyne/models/modules/transformer.py
```diff
@@ -184,8 +223,248 @@ def name(self) -> str:
...
+        )
+        self.symbols_linear = nn.Linear(
+            kwargs["d_model"],
+            # FIXME: This will break when used if odd dim_feedforward
+            int(kwargs["dim_feedforward"] / 2),
+            bias=kwargs.get("bias"),
+            **factory_kwargs,
+        )
+        self.features_linear = nn.Linear(
+            kwargs["d_model"],  # TODO: Seperate feature embedding size?
```
kylebgorman, line 315: typo: "Seperate" for "separate"

kylebgorman (commented):
Sorry @Adamits I thought this wasn't working yet. (Do we have to migrate to PyTorch 2 also to get this working?) 

I added two dumb questions, and a few typos. 

I have a smarter question. Right now you're using a divide-by-two trick, which as you note won't work if the dimensionality is odd. Why divide at all? I realize this will give us a bigger actual feedforward layer but I think that's what most people would expect to happen. 

@bonham79 any thoughts on this PR in general?

Adamits:
> Sorry @Adamits I thought this wasn't working yet. (Do we have to migrate to PyTorch 2 also to get this working?)

No it should work. I have comments about flags we might need when we go to torch 2.0. This is because I could only find the references to certain torch things (e.g. _sa_block) in torch 2.0, and some of the args I passed were throwing errors in my <2.0 torch.

> I added two dumb questions, and a few typos.
> 
> I have a smarter question. Right now you're using a divide-by-two trick, which as you note won't work if the dimensionality is odd. Why divide at all? I realize this will give us a bigger actual feedforward layer but I think that's what most people would expect to happen.

This is because I want a separate symbols and features multiheaded attention, so I can query only the symbols. But to run the Transformer it expects representations of size `h` in the feedforward block. Since the torch implementation iirc computes the mh attention sizes from the number of attention heads, and feedforward size, we need to half the size of each since for each attention head, we know have TWO representations (one for symbols and one for features), and then concatenate them before passing them through the feedforward layer. I could instead keep the dimensions the same and average them, but concat seems better to me.

> @bonham79 any thoughts on this PR in general?

yoyodyne/models/modules/transformer.py
Adamits: I believe this is in the torch `TransformerDecoderLayer` that I inherit from.

yoyodyne/models/modules/transformer.py
Adamits: Oh i will double check.

Adamits: Adding a norm after each layer looks right.

Adamits: Actually, I am not sure I see this in the original paper, but it is there in the torch implementation.

Adamits: Ah, it is optional and we do not set it. I don't see a citation in their code for when you'd want to set it, but we should keep it here in case that option is set.

kylebgorman: Okay but maybe we should make a note to try layer norm. I thought that was the standard for transformer encoder layers.

kylebgorman:
> > Sorry @Adamits I thought this wasn't working yet. (Do we have to migrate to PyTorch 2 also to get this working?)
> 
> No it should work. I have comments about flags we might need when we go to torch 2.0. This is because I could only find the references to certain torch things (e.g. _sa_block) in torch 2.0, and some of the args I passed were throwing errors in my <2.0 torch.
> 
> > I added two dumb questions, and a few typos.
> > I have a smarter question. Right now you're using a divide-by-two trick, which as you note won't work if the dimensionality is odd. Why divide at all? I realize this will give us a bigger actual feedforward layer but I think that's what most people would expect to happen.
> 
> This is because I want a separate symbols and features multiheaded attention, so I can query only the symbols. But to run the Transformer it expects representations of size `h` in the feedforward block. Since the torch implementation iirc computes the mh attention sizes from the number of attention heads, and feedforward size, we need to half the size of each since for each attention head, we know have TWO representations (one for symbols and one for features), and then concatenate them before passing them through the feedforward layer. I could instead keep the dimensions the same and average them, but concat seems better to me.

Concat is fine. Why not act as if when people request `h` that it was actually `2h` and use `h` of them for source symbols and `h` of them for features?

Adamits:
> > > Sorry @Adamits I thought this wasn't working yet. (Do we have to migrate to PyTorch 2 also to get this working?)
> > 
> > 
> > No it should work. I have comments about flags we might need when we go to torch 2.0. This is because I could only find the references to certain torch things (e.g. _sa_block) in torch 2.0, and some of the args I passed were throwing errors in my <2.0 torch.
> > > I added two dumb questions, and a few typos.
> > > I have a smarter question. Right now you're using a divide-by-two trick, which as you note won't work if the dimensionality is odd. Why divide at all? I realize this will give us a bigger actual feedforward layer but I think that's what most people would expect to happen.
> > 
> > 
> > This is because I want a separate symbols and features multiheaded attention, so I can query only the symbols. But to run the Transformer it expects representations of size `h` in the feedforward block. Since the torch implementation iirc computes the mh attention sizes from the number of attention heads, and feedforward size, we need to half the size of each since for each attention head, we know have TWO representations (one for symbols and one for features), and then concatenate them before passing them through the feedforward layer. I could instead keep the dimensions the same and average them, but concat seems better to me.
> 
> Concat is fine. Why not act as if when people request `h` that it was actually `2h` and use `h` of them for source symbols and `h` of them for features?

That is probably fine too, but " act as if when people request h" might require some fiddling with some deeper abstractions in torch iirc. I can't remember clearly so I should look into this either way.

Adamits:
Hey Kyle,

Oh good catch. I reproduced the `target_mask` for both cases. Not sure what I overlooked before,but hopefully something simple. I will check it out soon.

Adamits:
> Hey Kyle,
> 
> Oh good catch. I reproduced the `target_mask` for both cases. Not sure what I overlooked before,but hopefully something simple. I will check it out soon.

I think its due to a mismatch between our interface (`target_mask`) and the torch 1.13.1 interface (`tgt_mask`). But I also thought this particular scripts should hit my interface anyway, so need to investigate further.

yoyodyne/models/modules/transformer.py
Adamits: I think I had forgotten why I had `tgt`, and then I must not have tested before changing it here :( sorry! (see comment about this error).

Adamits:
Ok for Error 2. from your comment, I found that I was setting the output size of the feed forward layers I added for compressing feature/symbols before concatenation. I have now fixed it.

I must have had the settings wrong in my tests, thank you for testing!!!

I will push what I have for the fixes now, but this PR still needs some smoothing out (re: `tgt` v.s. `target` at the very least).

kylebgorman:
Would it be helpful for me to test again at this point? Or do you want more time with it?

Adamits:
If you could test again, that would be awesome :)

kylebgorman (commented):
Thanks Adam.

* Features-less (Maori data) works on GPU as of this commit and achieves a decent accuracy
* Features-full (Polish data, I am using the linear feature encoder) also works on GPU, but accuracy is awful: https://wandb.ai/hyperparam-sensitivity/pol/runs/nvtn0q1o?workspace=user-kylebgorman

A note for later. Validation (at least for Polish) takes a rather long time, the data loader visually appears to have freezed if you're watching the progress bars and it accounts for a lot of the wall clock time. I wonder if the batches are just too big; recall we are using the trick where my already big batch sizes--400?--are twice as large during validation, when I'm not collecting gradients, as during training. Maybe that was not a good heuristic after all.

Adamits:
> Thanks Adam.
> 
> * Features-less (Maori data) works on GPU as of this commit and achieves a decent accuracy
> * Features-full (Polish data, I am using the linear feature encoder) also works on GPU, but accuracy is awful: https://wandb.ai/hyperparam-sensitivity/pol/runs/nvtn0q1o?workspace=user-kylebgorman

Thanks so much for the testing.

I wonder if my good results for this were inadvertently calling it without features. I will be debugging today. 

> A note for later. Validation (at least for Polish) takes a rather long time, the data loader visually appears to have freezed if you're watching the progress bars and it accounts for a lot of the wall clock time. I wonder if the batches are just too big; recall we are using the trick where my already big batch sizes--400?--are twice as large during validation, when I'm not collecting gradients, as during training. Maybe that was not a good heuristic after all.

Does it take significantly longer than the vanilla transformer? I wonder how much bigger the model is. I recall the pointer-generator is a lot slower than the vanilla LSTM with the same parameters. I can play with hyperparameters a bit to see if batch size should be smaller anyway. But yeah maybe we want to have an eval batch size option (we removed it right?), and just default to twice the train batch size with a message that we log saying as much.

kylebgorman:
> > A note for later. Validation (at least for Polish) takes a rather long time, the data loader visually appears to have freezed if you're watching the progress bars and it accounts for a lot of the wall clock time. I wonder if the batches are just too big; recall we are using the trick where my already big batch sizes--400?--are twice as large during validation, when I'm not collecting gradients, as during training. Maybe that was not a good heuristic after all.
> 
> Does it take significantly longer than the vanilla transformer? I wonder how much bigger the model is. 

> I recall the pointer-generator is a lot slower than the vanilla LSTM with the same parameters.

I do'nt have this experience: the ptr-generator is about as performant as the vanilla LSTM and converges muchhhh faster.

> I can play with hyperparameters a bit to see if batch size should be smaller anyway. But yeah maybe we want to have an eval batch size option (we removed it right?)...

If I'm right and the batch size is too large, I disagree, we should just use the same one for train and validation and get both from `--batch_size`. We already have too many options.

Adamits:
> I do'nt have this experience: the ptr-generator is about as performant as the vanilla LSTM and converges muchhhh faster.

Sorry I meant that one step of inference takes much longer. Agree that it converges in far fewer steps.

> If I'm right and the batch size is too large, I disagree, we should just use the same one for train and validation and get both from --batch_size. We already have too many options.

Ok, I think that's fine! Tbf I think its pretty normal to give an option for different train/inference batch sizes in these libraries, but I agree we have too many options so your recommendation is the best choice!

bonham79:
@Adamits Do any of the transformer modules use `batch_norm` or is it all `layer_norm`. If the former, we probably should make a flag for `eval_batch_size`

Adamits:
@bonham79 Should just be layer norm.

kylebgorman:
> Hey Kyle,
> 
> I think this may be a hyperparameters issue and not a bug. 

Nice. I can confirm this evening. I mostly borrowed hyperparameters from the Wu et al. paper even though they weren't looking at a pointer-generator model.

kylebgorman:
> @bonham79 Should just be layer norm.

Not directly related to this PR but I would like to confirm that we're using this like I assumed we are.

kylebgorman:
Yes, this smaller model works just fine, thanks Adam. Should I merge now?

Adamits:
> Yes, this smaller model works just fine, thanks Adam. Should I merge now?

Great! I will do a final pass to make sure things make sense here. I will hopefully get to it in a few hours, then I will ping you here.

kylebgorman:
> > Not directly related to this PR but I would like to confirm that we're using this like I assumed we are.
> 
> Ok, should we talk about this? I have never really thought about batch norm, but definitely did not explicitly implement anything to use it. In the transformer, I implemented it to use layer norm, as is typical, and we have the option of doing the norm before or after the FF layer (inside the residual block or after).

This is all I wanted to check, that encoder and decoder have some form of layer norm: 

* [In encoder](https://github.com/Adamits/yoyodyne/blob/807874bc20c51c798e31077b96f2e1debcbfda7d/yoyodyne/models/modules/transformer.py#L208) 
* [In decoder](https://github.com/Adamits/yoyodyne/blob/807874bc20c51c798e31077b96f2e1debcbfda7d/yoyodyne/models/modules/transformer.py#L577) 

LGTM.

bonham79:
Alrighty, not a big deal then. Just wanted to check

Adamits:
Ok made a last pass and added a `TODO`. I think this can be merged now though.

kylebgorman (approved):
LGTM, merging now.

PR #140, opened 2023-10-02
WIP: single embeddings matrix design

### LINKED ISSUE #156: Pointer-generator crashes when source and target are disjoint
Traceback (most recent call last):
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/call.py", line 38, in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 650, in _fit_impl
    self._run(model, ckpt_path=self.ckpt_path)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1112, in _run
    results = self._run_stage()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1191, in _run_stage
    self._run_train()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1204, in _run_train
    self._run_sanity_check()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1276, in _run_sanity_check
    val_loop.run()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/dataloader/evaluation_loop.py", line 152, in advance
    dl_outputs = self.epoch_loop.run(self._data_fetcher, dl_max_batches, kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py", line 137, in advance
    output = self._evaluation_step(**kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py", line 234, in _evaluation_step
    output = self.trainer._call_strategy_hook(hook_name, *kwargs.values())
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1494, in _call_strategy_hook
    output = fn(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/strategies/strategy.py", line 390, in validation_step
    return self.model.validation_step(*args, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/models/base.py", line 286, in validation_step
    greedy_predictions = self(batch)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1194, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/models/pointer_generator.py", line 412, in forward
    predictions = self.decode(
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/models/pointer_generator.py", line 345, in decode
    output, decoder_hiddens = self.decode_step(
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/models/pointer_generator.py", line 286, in decode_step
    gen_probs = self.generation_probability(context, hidden, embedded)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1194, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/models/pointer_generator.py", line 72, in forward
    p_gen += self.W_emb(target_embeddings) + self.bias.expand(
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/module.py", line 1194, in _call_impl
    return forward_call(*input, **kwargs)
  File "/usr/local/lib/python3.10/dist-packages/torch/nn/modules/linear.py", line 114, in forward
    return F.linear(input, self.weight, self.bias)
RuntimeError: CUDA error: CUBLAS_STATUS_EXECUTION_FAILED when calling `cublasGemmEx( handle, opa, opb, m, n, k, &falpha, a, CUDA_R_16F, lda, b, CUDA_R_16F, ldb, &fbeta, c, CUDA_R_16F, ldc, CUDA_R_32F, CUBLAS_GEMM_DFALT_TENSOR_OP)`

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/bin/yoyodyne-train", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/train.py", line 338, in main
    best_checkpoint = train(trainer, model, datamodule, args.train_from)
  File "/usr/local/lib/python3.10/dist-packages/yoyodyne/train.py", line 232, in train
    trainer.fit(model, datamodule, ckpt_path=train_from)
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 608, in fit
    call._call_and_handle_interrupt(
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/call.py", line 63, in _call_and_handle_interrupt
    trainer._teardown()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/trainer/trainer.py", line 1175, in _teardown
    self.strategy.teardown()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/strategies/strategy.py", line 499, in teardown
    self.accelerator.teardown()
  File "/usr/local/lib/python3.10/dist-packages/pytorch_lightning/accelerators/cuda.py", line 75, in teardown
    _clear_cuda_memory()
  File "/usr/local/lib/python3.10/dist-packages/lightning_fabric/accelerators/cuda.py", line 366, in _clear_cuda_memory
    torch.cuda.empty_cache()
  File "/usr/local/lib/python3.10/dist-packages/torch/cuda/memory.py", line 125, in empty_cache
    torch._C._cuda_emptyCache()
RuntimeError: CUDA error: device-side assert triggered
CUDA kernel errors might be asynchronously reported at some other API call,so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.

Adamits:
Throwing this WIP up to store all vocabs in a single embeddings matrix shared between source, target, and features. This will fix the current pointer-generator issues when we have disjoint vocabularies.  I will get back to implementing it later this week.

- Computes a single `vocab_map`
- Retains `target_vocabulary` and `source_vocabulary` for logging.
- Moves embedding initialization onto the model
- Passes a single embedding layer to each module to share.

Closes #156.

**TODO**
- I did not do anything with features. I still need to merge these into the single vocab, being careful not to break anything that requires separate features.
- This does not handle the case that someone may want to have unmerged embeddings, such that the same unicode codepoint on the source and target side should be considered different. We will need to update the index to mark those characters with special symbols to implement this.


kylebgorman (commented):
Looks good so far. Some thoughts:

* I will file a bug describing why we're doing this for you to link to...give me a little time though. This will also describe our discussion of embedding too, though that's a secondary issue.
* I propose we shouldn't store the separate vocabularies in any class object. They should be computed separately, logged in the pretty-print fashion, then thrown out. This requires moving those logging statements, but so be it.
* `vocab_map` would be better just called `index` and we should throw out the `SymbolMap` class.

Will work on this in a day or two.

kylebgorman:
I created issue #142 to document the plan here.

kylebgorman:
I'm not a merge vs. rebase N*zi but the way the PR looks when I review is plenty clear as is.

Adamits:
Yeah sounds good.

> vocab_map would be better just called index and we should throw out the SymbolMap class.

Does this mean we would call it like `index.index` (consider that in dataset.py wew currently have e.g. `self.index.vocab_map`)?

kylebgorman:
> Does this mean we would call it like `index.index` (consider that in dataset.py wew currently have e.g. `self.index.vocab_map`)?

You could make it so `self.index` could so be called, i.e., by perlocating that ability upward...if I'm understanding correctly.

Adamits:
Ok I was working on replacing the indexing with one method as you suggested but there is an issue here:

our output space should be `target_vocab_size`. We *could* set this to the vocab_size (source + target + features), but this seems theoretically odd, since in many cases most of the output space is invalid, but we will still spill probability mass into it.

If we set it to be the a subset of the vocab (`target_vocab_size`), then we need to somehow map these output indices back to our vocabulary. Right now I was thinking about how this impacts turning ints back into symbols, but I think this also impacts decoder modules, which need to lookup the target in the shared embedding matrix.

Adamits:
I think I can make this work with some tricky offsets to map target indices drawn from {0, ..., target_size} to the embedding matrix indices of {0, ..., source+target+features size}. However, this will require providing this info to the model so we can offset predictions when decoding. This seems a bit awkward to me.

Adamits:
Still probably some clunkiness here, and I need to fix formatting. But I wanted to put this up in case you want to take a look:

I have tested for Attentive LSTM and pg LSTM, and it solves the indexing issue Michael has. When I come back to this tomorrow I also want to set `--tie_embeddings` as the default (e.g. defaults to true). Though I guess adding a flag like `--no_tie_embeddings` is a bit awkward?

kylebgorman:
> I have tested for Attentive LSTM and pg LSTM, and it solves the indexing issue Michael has. When I come back to this tomorrow I also want to set `--tie_embeddings` as the default (e.g. defaults to true). Though I guess adding a flag like `--no_tie_embeddings` is a bit awkward?

Look what we did for `--log_wandb` and `--no_log_wandb` [here](https://github.com/CUNY-CL/yoyodyne/blob/a8340043e9834310b51f71bfe2f5470a345ae5f7/yoyodyne/train.py#L317).

kylebgorman:
> Look what we did for `--log_wandb` and `--no_log_wandb` [here](https://github.com/CUNY-CL/yoyodyne/blob/a8340043e9834310b51f71bfe2f5470a345ae5f7/yoyodyne/train.py#L317).

Nevermind, that's not relevant. I was looking for a flag that defaults to `True`. [Here's a better example](https://github.com/CUNY-CL/wikipron/blob/8c17aba81c87be530fbdb75aeb7d706754f2039c/src/wikipron/cli.py#L22).

Adamits:
I have cleaned this up and tested it. Everything works except Transducer right now, which seems to be because of the target_vocab_size manipulation that happens in the model here https://github.com/CUNY-CL/yoyodyne/blob/master/yoyodyne/models/transducer.py#L43. 

I think its because `target_vocab_size` gets updated AFTER we initialize the single embedding matrix. This can cause idnex out of bounds errors. I can't work on this until tonight or tomorrow. @bonham79 in the meantime if you have any intuition about a good fix for this please take a look!

Adamits:
And, as usual, black is failing the test but passing locally. Sorry I can never remember what the culprit is :(

kylebgorman:
> And, as usual, black is failing the test but passing locally. Sorry I can never remember what the culprit is :(

`pip install -r requirements.txt` ought to set you to the exact version to use. (They have been making changes to the generated format...and I recently upgraded it because there was a security issue in a dependency.)

Adamits:
> pip install -r requirements.txt ought to set you to the exact version to use. (They have been making changes to the generated format...and I recently upgraded it because there was a security issue in a dependency.)

Ah thanks, looks like that worked.

yoyodyne/data/indexes.py
```diff
@@ -4,36 +4,80 @@
...
+            tie_embeddings: (bool).
+        """
+        super().__init__()
+        self.tie_embeddings = tie_embeddings
+
+        # We store all separate vocabularies for logging purposes.
+        # If embeddings are tied, so are the vocab items.
+        # Then, the source and target vocabularies are the union.
+        if self.tie_embeddings:
+            vocabulary = list(
```
kylebgorman, line 45: This cast to `list` is no-op because `sorted` only returns a list.

yoyodyne/models/__init__.py
```diff
@@ -78,3 +79,16 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         default="attentive_lstm",
         help="Model architecture. Default: %(default)s.",
     )
+    parser.add_argument(
+        "--tie_embeddings",
+        action="store_true",
+        default=defaults.TIE_EMBEDDINGS,
+        help="Shares embeddings for the source and target vocabularies. "
+        "This should always be set to true for pointer generator models.",
```
kylebgorman, line 87: point of clarification but isn't this just recommended in particular for such models? I think it's valid for the encoder and decoder to have separate embeddings with such models. Putting it another way, when the ptr-gen concept is relevant to your data, so are shared embeddings, but they're not required.

yoyodyne/models/__init__.py
```diff
@@ -78,3 +79,16 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
...
+    )
+    parser.add_argument(
+        "--no-tie_embeddings",
+        action="store_false",
+        dest="tie_embeddings",
+        help="Separates embeddings for the source and target vocabularies.",
```
kylebgorman, line 93: To fit style this should be: `no_tie_embeddings`. And I take the stance that you don't need to put a docstring for the negative ones. argparse is smart and groups the positive and negative pairs with the same destination together, so there's no ambiguity there.

yoyodyne/models/base.py
```diff
@@ -25,9 +25,9 @@ class BaseEncoderDecoder(pl.LightningModule):
     pad_idx: int
     start_idx: int
     # Sizes.
-    source_vocab_size: int
-    features_vocab_size: int
+    vocab_size: int
     target_vocab_size: int
+    features_vocab_size: int
```
kylebgorman, line 30: This order of the three is consistent across a bunch of things so you may want not to change it, even if you're changing the order in which they appear in the symbol table...

yoyodyne/train.py
```diff
@@ -165,6 +166,20 @@ def get_model_from_argparse_args(
         models.BaseEncoderDecoder.
     """
     model_cls = models.get_model_cls(args.arch)
+    if (
```
kylebgorman, line 169: ditto what I said above---I don't think this is required

yoyodyne/train.py
```diff
@@ -165,6 +166,20 @@ def get_model_from_argparse_args(
...
+        args.arch
+        in [
+            "pointer_generator_lstm",
+            "pointer_generator_transformer",
+            "transducer",
+        ]
+        and not args.tie_embeddings
+    ):
+        msg = f"--tie_embeddings set to {args.tie_embeddings}, but "
```
kylebgorman, line 178: if you're doing this though, better to make it a multiple lines of strings with implicit concatenation. String `+=` is really slow and happens at "runtime", but doing it like like:

```python
raise Error(
   "message part 1"
   "message part 2"
)
```

is handled once at "compile" (interpretation) time, not repeatedly at runtime. (And even interpretation can be cached, that's what `.pyc` and `.pyo` objects are.)

kylebgorman (commented):
Some comments below. (I also made two minor style edits.) Global notes:

* README should also be updated to reflect.
* Would you say this simplifies the "symbol math" we have to go through, or is it worse?
* It looks to me like you deleted the two embedding initialization static methods in the base model class, and then didn't replace them anywhere else.

yoyodyne/models/__init__.py
Adamits: Currently it is not technically valid, because our code explicitly combines the source and target dist, and if some tokens are OOV in one or the other (It might only go in one direction, I think the issue is source OOV in target?) we get index errors. We *could* fix this actually by just creating zero indices in the smaller vocabulary I think? But then I guess this breaks things because we need the indices of the output distribution to index the encoded indices.

Maybe we can make a new issue to think this through. This can also be instructive for how to implement an optional mapping for disjoint vocabs.

Adamits: Sorry for the rambling train of though here, but actually I think this PR ensures that "the indices of the output distribution to index the encoded indices" so maybe this is an easy feature update for a different PR.

kylebgorman: So, we could have separate embeddings with tricks {like} [this]. But I don't think it should give an error if it's not enabled.

If it's mandatory downstream, then we should just force it to be true rather than demand it of our CLI users, I think. 

Sure, create another issue for discussion.

yoyodyne/train.py
Adamits: Ah, good to know!!!

Adamits: Also, I had originally tried to handle this in the classes but had some issues. I think when we figure out a PR for class callbacks as you suggested in #111 , this can go there.

yoyodyne/data/indexes.py
```diff
@@ -42,9 +42,7 @@ def __init__(
         # If embeddings are tied, so are the vocab items.
         # Then, the source and target vocabularies are the union.
         if self.tie_embeddings:
-            vocabulary = list(
-                sorted(set(source_vocabulary) | set(target_vocabulary))
-            )
+            vocabulary = sorted(set(source_vocabulary) | set(target_vocabulary))
```
kylebgorman, line 45: Oh, I may have mispoke here earlier. This won't work because you're unioning the sets first. I went off half-cocked and what you had earlier is actually correct. Sorry about that.

Adamits:
> README should also be updated to reflect.

Ok will do. I think just `--tied_embedings/--no_...` right?

> Would you say this simplifies the "symbol math" we have to go through, or is it worse?
Can you be more specific here?

> It looks to me like you deleted the two embedding initialization static methods in the base model class, and then didn't replace them anywhere else.
I move them from the base module to the base model to follow the design pattern we discussed. I think the two `base.py` modeling files made it look confusing :D.

kylebgorman:
> Ok will do. I think just `--tied_embedings/--no_...` right?

Sure. I might create a section before the section on reserved symbols, call it `Tied embeddings`, and then say that by default we share embeddings between the source and target, but one can disable this with `--no_tied_embeddings`. That seems quite pithy.

Then you should also maybe mention the new class of reserved symbols in the following section: `{...}`. 

>> Would you say this simplifies the "symbol math" we have to go through, or is it worse?
> Can you be more specific here?

Is the way we build the index simpler (ignoring the additional complexity from tying the embeddings) or more complicated, nor about the same? 

> I move them from the base module to the base model to follow the design pattern we discussed. I think the two `base.py` modeling files made it look confusing :D.

Roger that.

yoyodyne/data/indexes.py
```diff
@@ -4,36 +4,79 @@
...
+
+        Args:
+            source_vocabulary (List[str]).
+            features_vocabulary (List[str], optional).
+            target_vocabulary (List[str], optional).
+            tie_embeddings: (bool).
+        """
+        super().__init__()
+        self.tie_embeddings = tie_embeddings
+        # We store all separate vocabularies for logging purposes.
```
kylebgorman, line 41: this doesn't look like it's used anywhere

yoyodyne/data/tsv.py
```diff
@@ -118,7 +122,11 @@ def _get_symbols(string: str, sep: str) -> List[str]:
         return list(string) if not sep else string.split(sep)
 
     def source_symbols(self, string: str) -> List[str]:
-        return self._get_symbols(string, self.source_sep)
+        symbols = self._get_symbols(string, self.source_sep)
+        # If not tied, then we distinguish the source vocab with {...}.
+        if not self.tie_embeddings:
```
kylebgorman, line 127: maybe easier to say:

```python
if self.tie_embeddings:
   return symbols
...
```

Less negation -> less cognitive load IMO. Also it's the default and that's another good heuristic for what goes first in a conditional.

yoyodyne/train.py
```diff
@@ -165,6 +166,20 @@ def get_model_from_argparse_args(
         models.BaseEncoderDecoder.
     """
     model_cls = models.get_model_cls(args.arch)
+    if (
+        args.arch
+        in [
```
kylebgorman, line 171: Considering adding the following (or something like this):

`# TODO(#156): add callback interface to check this.`

kylebgorman:
It's not in your PR description but this will close #142 and I think this will close #156 too.

yoyodyne/data/indexes.py
Adamits: This is meant to explain the properties `source_vocabulary, self.target_vocabulary, self.features_vocabulary`, which would not be needed since we use one embedding matrix, but which we store for logging purposes.

kylebgorman (approved):
LGTM. I'm going to test locally, just a bit, just to be sure though.

Adamits:
> It looks like we need to test if there are features, since I'm getting a breakage if there's not. I'll just modify to reflect...

Oh shoot, thank you for testing.

kylebgorman:
Okay, I made a change to `indexes.py` so it no longer crashes if there are no features. This does seem to work on my Polish tests, but validation accuracy performance is quite poor (though non-zero) on Maori, for which we have no features. The best checkpoint gives me only 11% accuracy. Can you take a look at this (after my commit 38311f530438b1154f75e83102ef3a13f134c613) and see if you can replicate? LMK if you need the command.

kylebgorman:
Enabling this may also be resulting in slightly poorer performance on Polish, just from eyeballing...

Adamits:
> Enabling this may also be resulting in slightly poorer performance on Polish with a pointer-generator LSTM, just from eyeballing...

Ok I will do more testing. I also just remembered a potential issue: this removes any mechanism to get a pointer-generator without shared embeddings, which was previously the default. That is because our code requires the same symbol on the source or target side to share the same index. But now, if we want this behavior, then they also must point to the same embedding. To support this, id need to change the implementation of how we combine the pointer and generator distributions.

kylebgorman:
> Ok I will do more testing. I also just remembered a potential issue: this removes any mechanism to get a pointer-generator without shared embeddings, which was previously the default. That is because our code requires the same symbol on the source or target side to share the same index. But now, if we want this behavior, then they also must point to the same embedding. To support this, id need to change the implementation of how we combine the pointer and generator distributions.

Yes, to confirm by default the embedding is shared and that ought to just work with pointer-generator architectures (which require that the embedding be shared, during training). I don't see the problem, since you did that already. That said, it may not be working as intended with pointer-generators. (I get basically the same results with or without tied embeddings on my LSTM experiments, though.)

Adamits:
> > Ok I will do more testing. I also just remembered a potential issue: this removes any mechanism to get a pointer-generator without shared embeddings, which was previously the default. That is because our code requires the same symbol on the source or target side to share the same index. But now, if we want this behavior, then they also must point to the same embedding. To support this, id need to change the implementation of how we combine the pointer and generator distributions.
> 
> Yes, to confirm by default the embedding is shared and that ought to just work with pointer-generator architectures (which require that the embedding be shared, during training). I don't see the problem, since you did that already. That said, it may not be working as intended with pointer-generators. (I get basically the same results with or without tied embeddings on my LSTM experiments, though.)

To be clear, we did not previously default to sharing embeddings, we just ensured characters had the same index, but those indices pointed to separate matrices. I think this is fine, I just mean that we no longer support disjoint *embeddings* for pointer generators since they require the src/tgt vocab index to be the same, but currently disjoint embeddings by necessity entail seperate src/tgt vocab indices.

kylebgorman:
> I trained on the polish abstractness data with features (pointer_generator_lstm, lstm encoder, linear features encoder, hidden size 512, 1 layer, emb 128).
> 
> Here are the first 3 val accs with the master branch:
> 
> ```
> 0.7530120611190796
> 0.8885542154312134
> 0.9036144614219666
> ```
> 
> And with this PR code:
> 
> ```
> 0.8177710771560669
> 0.8885542154312134
> 0.9066265225410461
> ```
> 
> There will be variance since embedding initialization is a function of the embedding matrix size.

I am getting similar results with that setup. Okay, false alarm on my part I guess. I may have installed from the wrong commit. Shall I merge?

kylebgorman:
> To be clear, we did not previously default to sharing embeddings, we just ensured characters had the same index, but those indices pointed to separate matrices. I think this is fine, I just mean that we no longer support disjoint _embeddings_ for pointer generators since they require the src/tgt vocab index to be the same, but currently disjoint embeddings by necessity entail seperate src/tgt vocab indices.

Okay, so we now require shared embeddings for pointer-generators. Do we have to? Could we allow disjoint embeddings if the indexes match up?

kylebgorman:
Sorry to keep jumping around with this, but I let this PR's version keep training on Polish and it went nan around 20 epochs. See if you can replicate?

```readonly DATA=../../abstractness/data
readonly ARCH=pointer_generator_lstm
# TODO: Set according to language.
readonly LANGUAGE=pol
readonly MODEL_DIR="../models/pointer_generator_bilstm"

readonly TRAIN="${DATA}/${LANGUAGE}/${LANGUAGE}_train.tsv"
cat "${DATA}/${LANGUAGE}/${LANGUAGE}_"[0-7]".tsv" > "${TRAIN}"
trap "rm -f ${TRAIN}" EXIT
readonly VAL="${DATA}/${LANGUAGE}/${LANGUAGE}_8.tsv"

yoyodyne-train \
    --experiment "${LANGUAGE}" \
    --train "${TRAIN}" \
    --val "${VAL}" \
    --model_dir "${MODEL_DIR}" \
    --features_col 3 \
    --arch "${ARCH}" \
    --source_encoder lstm \
    --features_encoder linear \
    --embedding_size 128 \
    --hidden_size 512 \
    --learning_rate .001 \
    --optimizer adam \
    --batch_size 32 \
    --max_epochs 60 \
    --dropout .3 \
    --seed 49 \
    --gradient_clip_val 3 \
    --precision 16 \
    --accelerator gpu
```

Adamits:
> Sorry to keep jumping around with this, but I let this PR's version keep training on Polish and it went nan around 20 epochs. See if you can replicate?

On my mac intel CPU, master branch ran for ~37 epochs and did not go nan. I am running this branch now and so far after 25 epochs we have ` loss=0.015, val_loss=0.0968, val_accuracy=0.920`. I am very curious what it could be. Could precision be handled differently on our devices?

Adamits:
> > To be clear, we did not previously default to sharing embeddings, we just ensured characters had the same index, but those indices pointed to separate matrices. I think this is fine, I just mean that we no longer support disjoint _embeddings_ for pointer generators since they require the src/tgt vocab index to be the same, but currently disjoint embeddings by necessity entail seperate src/tgt vocab indices.
> 
> Okay, so we now require shared embeddings for pointer-generators. Do we have to? Could we allow disjoint embeddings if the indexes match up?

Yes, but simulating the indexes matching up will require tricks at runtime somehow, since in a single embedding matrix we need a unique index per symbol.

kylebgorman:
> > Sorry to keep jumping around with this, but I let this PR's version keep training on Polish and it went nan around 20 epochs. See if you can replicate?
> 
> On my mac intel CPU, master branch ran for ~37 epochs and did not go nan. I am running this branch now and so far after 25 epochs we have ` loss=0.015, val_loss=0.0968, val_accuracy=0.920`. I am very curious what it could be. Could precision be handled differently on our devices?

That's my hypothesis, then. Okay. Let's not worry on that issue for now if it's not replicated on the second machine we try...we can wait to see if it is repeatedly reported.

kylebgorman:
> Yes, but simulating the indexes matching up will require tricks at runtime somehow, since in a single embedding matrix we need a unique index per symbol.

So I guess we could permit non-tieing for pointer-generator encoders using said tricks. Is that a feature you want to consider in the future or no?

kylebgorman:
also should I merge this now?

Adamits:
> > Yes, but simulating the indexes matching up will require tricks at runtime somehow, since in a single embedding matrix we need a unique index per symbol.
> 
> So I guess we could permit non-tieing for pointer-generator encoders using said tricks. Is that a feature you want to consider in the future or no?

Given that these results look good, probably not, but it is worth remembering just because one may want to do that for reproducibility reasons.

Adamits:
> also should I merge this now?

Maybe I will take the time today to run a fuller test on a handful more architectures first. Sound ok?

kylebgorman:
> Maybe I will take the time today to run a fuller test on a handful more architectures first. Sound ok?

Yeah do it, then if you don't have any contrary indications at EOB we'll merge.

Adamits:
Another thought as I start running local tests: should we log something about the val for tied_embeddings and what it means? Or is relying on normal arg logging + the README good enough?

kylebgorman:
> Another thought as I start running local tests: should we log something about the val for tied_embeddings and what it means? Or is relying on normal arg logging + the README good enough?

I don't think this is more important than anything else we log by just showing people what args are being passed (including the default ones).

Adamits:
Running the Transducer I get: `IndexError: index out of range in self`, which looks like it points to
```
File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/modules/base.py", line 85, in embed
    embedded = self.embeddings(symbols)
```

kylebgorman:
> Running the Transducer I get: `IndexError: index out of range in self`, which looks like it points to
> 
> ```
> File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/modules/base.py", line 85, in embed
>     embedded = self.embeddings(symbols)
> ```

I can replicate. I am not sure if it was a regression associated with this PR or an earlier regression.

Adamits:
> > Running the Transducer I get: `IndexError: index out of range in self`, which looks like it points to
> > ```
> > File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/modules/base.py", line 85, in embed
> >     embedded = self.embeddings(symbols)
> > ```
> 
> I can replicate. I am not sure if it was a regression associated with this PR or an earlier regression.

Ok I will take a closer look

Adamits:
I think it has to do with how the action vocabulary is created. I recall @bonham79 had to figure out some tricks here that I don't recall, so I probably need to update it to play nice with the new vocabulary.

Adamits:
> @Adamits Sorry, notification for this fell into spam. Do you still need info for the action vocabulary?

Hey, I forgot about this PR actually. Yeah I could use some help. This PR changes the vocab s.t. we store source and target vocab in a single embedding matrix---if we want separate embeddings, then the source vocab items are just wrapped in `{...}`. I was getting an index error when running the transducer that I think is caused by w/e happens after the initial vocabulary is built, where the transducer expands the action vocab.

Right now I cannot remember the details of what I had found so far, so maybe if you recll what the transducer does there and could give a high level overview I could jump back into this?

bonham79:
@Adamits Sure, so technically the transducer never uses the target vocab. The expert preprocesses the target vocab from train and creates a bunch of `edit actions` (like, straight Levenshtein edits) and passes that to the transducer as a prediction vocabulary. The target vocab only comes back into play once the actions are processed. So your fix would either 

a). let the transducer do its own thing with the target actions
or
b). expand the source + target embedding space to include the action embeddings.

The change occurs here: https://github.com/CUNY-CL/yoyodyne/blob/c0640676adc8e25b3e41b5f7c574e7e115381a03/yoyodyne/models/transducer.py#L43 I just pass the expert's list of edit actions to the target vocab then initialize based off that. 

Mind sharing the error you were getting?

Adamits:
@bonham79  Ok thanks that is helpful!! I assume I initialized the embedding matrix before this change to target_vocab size. Anyway, here is the stack:

```
Traceback (most recent call last):
  File "/Users/adamwiemerslage/anaconda3/envs/py3-9/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/Users/adamwiemerslage/anaconda3/envs/py3-9/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/train.py", line 399, in <module>
    main()
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/train.py", line 394, in main
    best_checkpoint = train(trainer, model, datamodule, args.train_from)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/train.py", line 261, in train
    trainer.fit(model, datamodule, ckpt_path=train_from)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 608, in fit
    call._call_and_handle_interrupt(
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/call.py", line 38, in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 650, in _fit_impl
    self._run(model, ckpt_path=self.ckpt_path)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1112, in _run
    results = self._run_stage()
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1191, in _run_stage
    self._run_train()
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1204, in _run_train
    self._run_sanity_check()
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1276, in _run_sanity_check
    val_loop.run()
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/loops/dataloader/evaluation_loop.py", line 152, in advance
    dl_outputs = self.epoch_loop.run(self._data_fetcher, dl_max_batches, kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/loops/loop.py", line 199, in run
    self.advance(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py", line 137, in advance
    output = self._evaluation_step(**kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/loops/epoch/evaluation_epoch_loop.py", line 234, in _evaluation_step
    output = self.trainer._call_strategy_hook(hook_name, *kwargs.values())
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/trainer/trainer.py", line 1494, in _call_strategy_hook
    output = fn(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/strategies/strategy.py", line 390, in validation_step
    return self.model.validation_step(*args, **kwargs)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/transducer.py", line 530, in validation_step
    predictions, loss = self(batch)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/transducer.py", line 119, in forward
    prediction, loss = self.decode(
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/transducer.py", line 196, in decode
    decoder_output = self.decoder(
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/modules/lstm.py", line 118, in forward
    embedded = self.embed(symbol)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/models/modules/base.py", line 85, in embed
    embedded = self.embeddings(symbols)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1518, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/module.py", line 1527, in _call_impl
    return forward_call(*args, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/modules/sparse.py", line 162, in forward
    return F.embedding(
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/torch/nn/functional.py", line 2233, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
IndexError: index out of range in self
```

bonham79:
@Adamits Looking through your changes, it seems you never extend the transducer's vocab size with the additional target vocab size. The target vocab can't be acquired through the vocab map since it's edit actions from the expert. (It's like, every potential target vocab has a copy or insert action, so it's 2x the observed vocabulary.)

Adamits:
> @Adamits Looking through your changes, it seems you never extend the transducer's vocab size with the additional target vocab size. The target vocab can't be acquired through the vocab map since it's edit actions from the expert. (It's like, every potential target vocab has a copy or insert action, so it's 2x the observed vocabulary.)


Right, is there a reason why the action vocabulary loops through the dataset instead of just getting a handle on the vocabulary? I think doing that would just fix this.

Adamits:
> > @Adamits Looking through your changes, it seems you never extend the transducer's vocab size with the additional target vocab size. The target vocab can't be acquired through the vocab map since it's edit actions from the expert. (It's like, every potential target vocab has a copy or insert action, so it's 2x the observed vocabulary.)
> 
> Right, is there a reason why the action vocabulary loops through the dataset instead of just getting a handle on the vocabulary? I think doing that would just fix this.

Sorry nevermind (though this may also be an issue) I am pretty sure my initial assessment is right: I now initialize the embeddings matrix before the transducer modifications. Will see if I can fix this elegantly.

Adamits:
Most recent commit seems to fix the Transducer issue I created. I will try to test more tomorrow night to be sure all of the changes in this most recent commit are needed.

bonham79:
> > @Adamits Looking through your changes, it seems you never extend the transducer's vocab size with the additional target vocab size. The target vocab can't be acquired through the vocab map since it's edit actions from the expert. (It's like, every potential target vocab has a copy or insert action, so it's 2x the observed vocabulary.)
> 
> Right, is there a reason why the action vocabulary loops through the dataset instead of just getting a handle on the vocabulary? I think doing that would just fix this.

Nope, no reason. I just do it because it has to pass an iterable off to the maxwell dependency so can add on the fly. 

Now that you mention it, it would be more consistent to just pass the dataloader vocabulary to the expert. (Or move expert around so it's calling maxwell while building the dataloader. But it's a cheap operation so not a major need.)

bonham79 (changes requested):
Just commenting to annoy myself to review the transducer changes.

kylebgorman:
Does this solve #156?

Adamits:
> Does this solve #156?

Oh yeah I think it does.

Adamits:
This PR is getting very very stale. From what I recall it was ready to go. Shall I run another round of tests on each arch for both train and predict, and if it all works, we merge?

kylebgorman:
> This PR is getting very very stale. From what I recall it was ready to go. Shall I run another round of tests on each arch for both train and predict, and if it all works, we merge?

@Adamits yes please merge this soon. I also thought it was basically ready to go.

Adamits:
Ok I ran these, and everything runs and looks good, except the Transducer. No errors, but I trained it for 5 epochs on english sigmorphon 2017 medium data, and the validation accuracy looks messed up. See the metrics.csv Table in markdown below. We also get an error when running predict.py.

However, when I run the same code from master (i) I get 0 accuracy after 5 epochs and the train loss is converging absurdly slowly and (ii) predict.py still errors.

| lr-Adam | step | val_loss            | val_accuracy       | epoch | train_loss         |
| ------- | ---- | ------------------- | ------------------ | ----- | ------------------ |
| 0.001   | 0    |                     |                    |       |                    |
|         | 15   | 0.9778230786323547  | 0.1899999976158142 | 0     |                    |
|         | 15   |                     |                    | 0     | 3.108109951019287  |
| 0.001   | 16   |                     |                    |       |                    |
|         | 31   | 0.7275158166885376  | 0.1899999976158142 | 1     |                    |
|         | 31   |                     |                    | 1     | 0.9277768731117249 |
| 0.001   | 32   |                     |                    |       |                    |
|         | 47   | 0.581281840801239   | 0.1899999976158142 | 2     |                    |
|         | 47   |                     |                    | 2     | 0.7454320788383484 |
| 0.001   | 48   |                     |                    |       |                    |
|         | 63   | 0.49098914861679077 | 0.1899999976158142 | 3     |                    |
|         | 63   |                     |                    | 3     | 0.6462973356246948 |
| 0.001   | 64   |                     |                    |       |                    |
|         | 79   | 0.4305487275123596  | 0.1899999976158142 | 4     |                    |
|         | 79   |                     |                    | 4     | 0.6198569536209106 |

Adamits:
Update ok the predict.py issue was minor and I fixed it. The transducer seems to be learning to just copy exactly the lemma. Not sure what to make of that yet...

kylebgorman:
I have also seen this pattern with validation accuracy, namely it being, implausibly, a constant. I don't know what the origin of it is.

Re: the copying of lemma, in the printout of the parameters do you see anything that would be doing feature encoding?

kylebgorman:
You might also want to try doing the manual merge against head to see if anything changed there helps...

Adamits:
After I merge master back in, I get the same thing but with a lower acc:

| lr-Adam | step | val_accuracy | val_loss            | epoch | train_loss         |
| ------- | ---- | ------------ | ------------------- | ----- | ------------------ |
| 0.001   | 0    |              |                     |       |                    |
|         | 15   | 0.002        | 0.9778230786323547  | 0     |                    |
|         | 15   |              |                     | 0     | 3.108109951019287  |
| 0.001   | 16   |              |                     |       |                    |
|         | 31   | 0.002        | 0.7275158166885376  | 1     |                    |
|         | 31   |              |                     | 1     | 0.9277768731117249 |
| 0.001   | 32   |              |                     |       |                    |
|         | 47   | 0.002        | 0.581281840801239   | 2     |                    |
|         | 47   |              |                     | 2     | 0.7454320788383484 |
| 0.001   | 48   |              |                     |       |                    |
|         | 63   | 0.002        | 0.49098914861679077 | 3     |                    |
|         | 63   |              |                     | 3     | 0.6462973356246948 |
| 0.001   | 64   |              |                     |       |                    |
|         | 79   | 0.002        | 0.4305487275123596  | 4     |                    |
|         | 79   |              |                     | 4     | 0.6198569536209106 |


Prediction is also now broken:

```
Traceback (most recent call last):
  File "/Users/adamwiemerslage/python-envs/yoyodyne/bin/yoyodyne-predict", line 8, in <module>
    sys.exit(main())
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/predict.py", line 158, in main
    model = get_model_from_argparse_args(args)
  File "/Users/adamwiemerslage/nlp-projects/morphology/yoyodyne/yoyodyne/predict.py", line 70, in get_model_from_argparse_args
    return model_cls.load_from_checkpoint(args.checkpoint)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/core/saving.py", line 139, in load_from_checkpoint
    return _load_from_checkpoint(
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/core/saving.py", line 188, in _load_from_checkpoint
    return _load_state(cls, checkpoint, strict=strict, **kwargs)
  File "/Users/adamwiemerslage/python-envs/yoyodyne/lib/python3.9/site-packages/pytorch_lightning/core/saving.py", line 234, in _load_state
    obj = cls(**_cls_kwargs)
TypeError: __init__() missing 1 required positional argument: 'expert'
```

kylebgorman:
The prediction bug is #197. Fine to submit over the top of that if everything else is working, IMO, since @bonham79 is on that one.

Adamits:
> The prediction bug is #197. Fine to submit over the top of that if everything else is working, IMO, since @bonham79 is on that one.

Sounds good. Let me just train the Transducer longer as a sanity check.I also feel unsure why the (static) accuracies are so much lower once I merge master. I would plan on committing the version with master merged since I resolved a couple conflicts, sound good?

I am also pretty sure you have to merge since this is your org---I dont think I have access to merge.

Adamits:
Ok I got the same result from the transducer. I also wonder if the particular dataset is somehow causing it to learn this copy heuristic. It seems like we should do more explicit testing of that model, and being able to predict would help to debug---would you be ok merging this and I can open a separate ticket?

kylebgorman:
I'm fine with that. I'll deal with the conflicts now.

Adamits:
> I'm fine with that. I'll deal with the conflicts now.

Oh Sorry I blanked on committing those. It is minor changes so I guess I will let you do that to get a 2nd pair of eyeson it?

kylebgorman:
I just dealt with the conflicts. Flake8 noticed that `source_vocab_size` in `train.py` was unused so I removed it. I also fixed the pretty-printing to avoid nested `[` and `]`. I tested locally and things are fine.

PR #147, opened 2023-11-01
Fixes several pointer-generator lstm bugs

Adamits:
I was not able to run the pointer-generator LSTM. It looks like the breaks have been here since #72.

- We now actually pass the encoded features through to the decode_step
- We correctly concatenate the feature attention if provided (previously we erroneously concatenated the source attention * 2)
- No longer adds the extra breaking dimension to the generation probability.

### DIFF
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -72,7 +72,7 @@ class GenerationProbability(nn.Module):
         p_gen += self.W_emb(target_embeddings) + self.bias.expand(
             attention_context.size(0), 1, -1
         )
-        # -> B x sequence_length.
+        # -> B x 1 x sequence_length.
         return torch.sigmoid(p_gen)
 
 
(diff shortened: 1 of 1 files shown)


yoyodyne/models/pointer_generator.py
```diff
@@ -259,7 +259,7 @@ def decode_step(
                 last_h0.transpose(0, 1), features_enc, features_mask
             )
             # -> B x 1 x 4*hidden_size.
-            context = torch.cat([context, context], dim=2)
+            context = torch.cat([context, features_context], dim=2)
```
kylebgorman, line 262: Driveby, but would it make sense to call them `source_context` and `features_context`?

kylebgorman (approved):
LGTM.

yoyodyne/models/pointer_generator.py
Adamits: True. Can do

kylebgorman: Okay, let me know when that's done and I'll merge.

kylebgorman:
I just tested some Maori and Polish (the latter has features) and it seems to be working also. Merging now.

PR #164, opened 2024-02-18
fix: makes message f-string

Adamits:


kylebgorman (approved):
Real easy error to make.

bonham79:
Oh lord, egg on my fave. Thanks for catch @Adamits

PR #166, opened 2024-02-21
W&B no longer "logs" (uploads) checkpoints.

kylebgorman:
I also remove some seemingly pointless logging of model paths; can restore if there's a reason we need this.

Note that there exist alternatives: for instance we can log any checkpoint that's an improvement. From [the docs](https://lightning.ai/docs/pytorch/stable/extensions/generated/lightning.pytorch.loggers.WandbLogger.html):

```python
# log model only if `val_accuracy` increases
wandb_logger = WandbLogger(log_model="all")
checkpoint_callback = ModelCheckpoint(monitor="val_accuracy", mode="max")
trainer = Trainer(logger=wandb_logger, callbacks=[checkpoint_callback])
```

But I think this would still be excessive and I can't imagine why you'd want to log a checkpoint on W&B, tbh.

### DIFF
--- a/yoyodyne/train.py
+++ b/yoyodyne/train.py
@@ -30,6 +30,8 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
         trainer_logger.append(loggers.WandbLogger(project=experiment))
         # Tells PTL to log the best validation accuracy.
         wandb.define_metric("val_accuracy", summary="max")
+        # Logs the path to local artifacts made by PTL.
+        wandb.config["local_run_dir"] = trainer_logger[0].log_dir
     return trainer_logger
 
 
--- a/yoyodyne/train.py
+++ b/yoyodyne/train.py
@@ -27,13 +27,9 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
     """
     trainer_logger = [loggers.CSVLogger(model_dir, name=experiment)]
     if log_wandb:
-        trainer_logger.append(
-            loggers.WandbLogger(project=experiment, log_model="all")
-        )
-        # Tells PTL to log best validation acc
+        trainer_logger.append(loggers.WandbLogger(project=experiment))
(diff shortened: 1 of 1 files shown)


yoyodyne/train.py
```diff
@@ -27,13 +27,9 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
     if log_wandb:
-        trainer_logger.append(
-            loggers.WandbLogger(project=experiment, log_model="all")
-        )
-        # Tells PTL to log best validation acc
+        trainer_logger.append(loggers.WandbLogger(project=experiment))
+        # Tells PTL to log the best validation accuracy.
         wandb.define_metric("val_accuracy", summary="max")
-        # Logs the path to local artifacts made by PTL.
-        wandb.config.update({"local_run_dir": trainer_logger[0].log_dir})
```
Adamits, line 36: I had this here because if you are running a large sweep somewhere, and later want to e.g. get predictions from certain W&B configs, this is easy to do with a script by just getting the CSV, and selecting rows representing runs. Then we just need row["local_run_dir"] to get the checkpoint.

This is especially helpful for sweeps where several runs crashed, and you want to easily differentiate completed from crashed runs. Do you think its worth keeping? Or is it too specific to my needs?

Adamits:
Thanks for catching this. I am sure this was here from copy/pasting from W&B examples. LGTM, pending the one inline comment.

yoyodyne/train.py
kylebgorman: If you think it's useful, I'll restore it. I think I can write it slightly more clearly.

kylebgorman: Restored, but I'm using dict-style syntax. LMK if this looks okay...it's working fine for me locally.

Adamits: Ah I did not know we could do that---yeah it looks better!

Adamits (approved):

PR #167, opened 2024-03-08
Possible cure for W&B memory leak

Othergreengrasses:
Added clean up at the end of the training function to free up the resources as suggested [here ](https://github.com/wandb/wandb/issues/1247#issuecomment-1457737657)

### DIFF
--- a/examples/wandb_sweeps/train_wandb_sweep.py
+++ b/examples/wandb_sweeps/train_wandb_sweep.py
@@ -5,6 +5,7 @@ import argparse
 import functools
 import traceback
 import warnings
+import torch
 
 import pytorch_lightning as pl
 import wandb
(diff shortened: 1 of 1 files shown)


examples/wandb_sweeps/train_wandb_sweep.py
```diff
@@ -42,7 +42,10 @@ def train_sweep(args: argparse.Namespace) -> None:
     best_checkpoint = train.train(trainer, model, datamodule, args.train_from)
     util.log_info(f"Best checkpoint: {best_checkpoint}")
 
-
+   # cleanup
```
kylebgorman, line 45: Please remove the blank line above, and instead make sure there are two blank lines between this and the following function (e.g., around line 48). 

The comment style should be something like: `# Cleans up.`: a sort of declarative sentence, punctuated as such. But that's not very descriptive. I would suggest something like the following:

```python
# Explicitly deallocates model and clears the cache, based on the following suggestion:
# 
#     https://github.com/wandb/wandb/issues/1247#issuecomment-1457737657
```

When you do something based on a "tip" it's good style to link to said tip, to politely give credit but also to provide context.

kylebgorman (commented):
Implementation looks very good to me. Made two style suggestions; the first one is needed to pass the automatic tests (which are currently failing). 

Does it seem to help with the apparent memory leaks you're seeing? It should be possible to just install a version from this branch and try some sweep experiments.

examples/wandb_sweeps/train_wandb_sweep.py
Othergreengrasses: Yes, it helped with the memory leaks without breaking anything.

kylebgorman (commented):
Can you fix the test errors (click on the red x's on the main PR page to see what they are)?

You'll need to `import torch` in the right place, and also run `black -l79` over the file, then stage and push those changes.

kylebgorman (approved):
The `torch` import should be in the second group since it's not a stdlib, but I can take care of that manually myself. LGTM otherwise.

PR #168, opened 2024-04-01
Log model parameters to wandb

Adamits:
Adds number of model parameters to W&B log. 

Note that PTL logs this to stdout by default, but it is convenient to have a column in W&B for this, so that we do not need to retrieve and parse all stdout logs for large analysis.

### DIFF
--- a/yoyodyne/train.py
+++ b/yoyodyne/train.py
@@ -326,7 +326,7 @@ def main() -> None:
     # Logs number of model parameters.
     if args.log_wandb:
         n_params = sum(p.numel() for p in model.parameters())
-        wandb.config.update({"n_model_params": n_params})
+        wandb.config["n_model_params"] = n_params
     # Tuning options. Batch autoscaling is unsupported; LR tuning logs the
     # suggested value and then exits.
     if args.auto_scale_batch_size:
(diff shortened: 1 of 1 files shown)


yoyodyne/train.py
```diff
@@ -323,6 +323,10 @@ def main() -> None:
     trainer = get_trainer_from_argparse_args(args)
     datamodule = get_datamodule_from_argparse_args(args)
     model = get_model_from_argparse_args(args, datamodule)
+    # Logs number of model parameters.
+    if args.log_wandb:
+        n_params = sum(p.numel() for p in model.parameters())
+        wandb.config.update({"n_model_params": n_params})
```
kylebgorman, line 329: Correct me if I'm wrong but didn't we establish that `config` is just a dictionary, or at least has a `__getitem__` like one? Then you could rewrite this:

```python
wandb.config["n_model_params"] = sum(p.numel() for p in model.parameters())
```

or whatever.

kylebgorman (approved):
LGTM but one suggestion?

Adamits:
Good call, that works.

kylebgorman (approved):
LGTM, assuming I'm right and this works. Shall I merge?

Adamits:
> LGTM, assuming I'm right and this works. Shall I merge?

I tested it---yes!

PR #171, opened 2024-04-06
Adds --val_mode.

### LINKED ISSUE #170: Model selection via validation loss
To support experiments on whether our model selection strategy (maximal validation accuracy), we should provide a flag which enables model selection based on minimal validation loss instead.

The interface I would suggest is something like this: a flag `--model_selection_mode` with a default value `accuracy` and an alternative value `loss`. (We already have something similar in place for the reduce-on-plateau scheduling: [see here](https://github.com/CUNY-CL/yoyodyne/blob/b6199700305474a2fbd307d509ea04e42b0341a5/yoyodyne/schedulers.py#L202)).

kylebgorman:
This allows one to either early stop/model select on loss or accuracy using the `--val_metric` flag. The default is as before: accuracy. 

This affects two things: `--save_top_k` model selection (and how the files are named), and the behavior of `--patience`. I could also integrate it with the mode for the reduce-on-plateau scheduler, but I think that'll require a lot more code.

It might seem at first glance that I stripped out the behavior associated with #169. However, as I document, you can simulate it easily just by doing `--save_top_k -1`, so I think that's actually fine. 

Closes #170.


yoyodyne/train.py
```diff
@@ -36,49 +36,65 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
-            save_on_train_epoch_end=True,
-            filename="model-{epoch:03d}",
-        )
     trainer_callbacks = [
-        checkpoint_callback,
         callbacks.LearningRateMonitor(logging_interval="epoch"),
         callbacks.TQDMProgressBar(),
     ]
+    # Parses validation mode.
+    if val_mode == "accuracy":
```
Adamits, line 65: Up to you, but I think these conditionals could contain code just to set `mode`, `monitor`, and `filename`, and then the code block that appends the callback can just be written once, using these 3 variables.

yoyodyne/train.py
```diff
@@ -276,31 +290,34 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
...
-        "--save_best",
-        action="store_true",
-        default=defaults.SAVE_BEST,
-        help=(
-            "Save the best checkpoint(s) according to the validation metric. "
-            "Default: True."
-        ),
+        "--top_k",
+        type=int,
+        default=defaults.TOP_K,
```
Adamits, line 310: Why do we remove the "save" prefix? I feel like top_k could be ambiguous with e.g. sampling, or decoding k hypotheses. "save_" also makes it match the ptl argument.

I would argue `save_top_k` is a better name.

Adamits (changes requested):

Adamits:
Looks good, thanks for thinking this through. I left two comments that may or may not require updates.

michaelpginn:
Also, for reference, HuggingFace calls this flag `metric_for_early_stopping` and `metric_for_best_model`. Since this covers both, maybe something like `--primary_metric` would be a more obvious name?

yoyodyne/train.py
kylebgorman: I tried that in an earlier rendition---and it worked, I just wasn't sure if I liked it. I'll restore it.

yoyodyne/train.py
kylebgorman: I don't really like that name (it sounds like a predicate that takes a boolean yes/no argument and not like one which takes a counting number argument) but this isn't really supposed to be a renaming-flags PR so I'll revert.

kylebgorman:
> be a more obvious name?

Let's take this up in a subsequent PR where we tweak names. I'm okay with "primary metric" I suppose.

kylebgorman:
> Could we combine this flag and the ones from #169? I.e., have a single `--val_mode` flag that can be `accuracy`, `loss`, or `none`, with the latter replacing the `--no_save_best` option.

I actually tried that and it isn't as clear cut as you might imagine, because then you can't control which metric patience uses if you enable that. So I think the solution I've taken here is the best one: namely, documentation. I submit that there's no real reason why you'd want to use different model selection and patience ( and reduce-on-plateau) metrics, so there's no reason we can't combine them.

yoyodyne/train.py
Adamits: > it sounds like a predicate that takes a boolean yes/no argument and not like one which takes a counting number argument

I see, this is a good point.

Issue #60, opened 2023-03-17
Torch and Lightning 2.0.0


kylebgorman:
The library does not work with Lightning (and one suspects that Torch itself is also an issue) > 2.0.0. The first issue I encounter when running `yoyodyne-train` with no arguments is related to a change in how Lightning command-line arguments are handled---I suspect there are at least a few more.

So that the library is not broken at head---which I consider unacceptable---I have pinned as follows:

```
pytorch-lightning>=1.7.0,<2.0.0
torch>=1.11.0,<2.0.0
```

What we need to do is just to migrate to 2.0.0, by fixing Lightning (and Torch, if any) bugs until things work, and then re-pin these two dependencies `>=2.0.0`. I have initially assigned myself, but I would welcome help.

kylebgorman:
I read these docs with some care. There are two steps:

1. Before migrating to 2.0, we should should implement a subclass of `LightningDataModule` for training:

kylebgorman:
I have read these docs and have the basic lay of the land now. The migration should go in two steps:

1. First, we should implement subclasses of `LightningDataModule` (this is available in Lightning pre 2.0 too). The `train_dataloader` method should return the training data dataloader (naturally) and similarly with `dev_dataloader` and `test_dataloader`. (Even though we don't need access to train, dev, and test in any one process, we can probably set it up so we can use the same one for training and inference.) Most of the logic of [`get_datasets`](https://github.com/CUNY-CL/yoyodyne/blob/035f2224c9ef075e1c909b20e4a08face509c1aa/yoyodyne/train.py#L180C1-L181C1) and [`get_loaders`](https://github.com/CUNY-CL/yoyodyne/blob/035f2224c9ef075e1c909b20e4a08face509c1aa/yoyodyne/train.py#L181) would be moved into this too. This is thus mostly just cleanup.
2. Then, we have to move arg parsing to `LightningCli` and migrate Torch and Lightning to >=2.0. This is a bit magical (it introspects to find flags for the dataset and models) but our house is mostly in order so it might just work. At this point there are probably one or two low-level Torch issues we'll hit up against too and can handle those as they arise...

Adamits:
Sounds good, I think I understand. Though I feel a bit annoyed that we have to use `LightningCli` instead of standard python libraries... I will try to prioritize it this week.

kylebgorman:
[...]

PR #173, opened 2024-04-08
Attempts migration to PyTorch >= 2.0.0.

kylebgorman:
See #60 for context, though we are not yet done with it since we are not migrating to PyTorch-Lightning 2.


kylebgorman:
Warnings I'm seeing with the transformer and Torch 2.2:

```
/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/transformer.py:286: UserWarning: enable_nested_tensor is True, but self.use_nested_tensor is False because encoder_layer.norm_first was True
```

```
/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/functional.py:5109: UserWarning: Support for mismatched key_padding_mask and attn_mask is deprecated. Use same type for both instead.
```

Both seem harmless. I am not sure what the fix is for the former; with the latter presumably you make sure you create the tensor types with the same type?

Adamits (commented):
Cool. Off the top of my head, I cannot remember what the `is_causal` flag will do since we already passed a causal mask.

However, if you have tested the pg models here (since I have tested both vanilla enc-dec models). Should we also test the transducer? Otherwise this LGTM.

Adamits:
> Warnings I'm seeing with the transformer and Torch 2.2:

> /home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/modules/transformer.py:286: UserWarning: enable_nested_tensor is True, but self.use_nested_tensor is False because encoder_layer.norm_first was True
> /home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/torch/nn/functional.py:5109: UserWarning: Support for mismatched key_padding_mask and attn_mask is deprecated. Use same type for both instead.
> Both seem harmless. I am not sure what the fix is for the former; with the latter presumably you make sure you create the tensor types with the same type?

Yeah I saw these too. For the mismatched type, I was struggling to actually figure out which types need to change and did not solve this on my end. Though I only tried a couple of things.

kylebgorman:
> Yeah I saw these too. For the mismatched type, I was struggling to actually figure out which types need to change and did not solve this on my end. Though I only tried a couple of things.

We could ask the devs, or we could silence it like we did for other warnings.

Adamits:
> We could ask the devs, or we could silence it like we did for other warnings.

Ok. Just realized this is a draft. Shall we split up any pending subtasks?

Which architectures still need to be tested? (I have already tested `transformer` and `attentive_lstm`)
I can also work on the boolean warning. Then maybe you could double check the is_causal flag behavior?

I think `torch.compile` is outside the scope of what we want here for now.

kylebgorman:
> Ok. Just realized this is a draft. Shall we split up any pending subtasks?
> 
> Which architectures still need to be tested? (I have already tested `transformer` and `attentive_lstm`) I can also work on the boolean warning. Then maybe you could double check the is_causal flag behavior?

Yeah, why don't you check on the warnings? 

I'll be done testing before end of day. 
 
> I think `torch.compile` is outside the scope of what we want here for now.

+1.

Adamits:
> I'll be done testing before end of day.

Nice, I will try to keep up with you then. I need to finish up making my lecture and teach today :P. But should have time in the afternoon to look into the warning more deeply.

kylebgorman:
Hard error on the transducer: `RuntimeError: could not create a primitive descriptor for an LSTM forward propagation primitive`. No idea what this means. (Will check to see if same error is present at Yoyodyne HEAD.)

Adamits:
Note: `is_causal` may have been replaced in more recent torch releases? https://pytorch.org/docs/stable/generated/torch.nn.TransformerDecoder.html

Adamits:
I don't think I can fork this, but I had just confused myself on the `mismatched key_padding_mask and attn_mask is deprecated. Use same type for both instead` warning.

We can solve it in 

by changing `generate_square_subsequent_mask` to

```
@staticmethod
   def generate_square_subsequent_mask(length: int) -> torch.Tensor:
        """Generates the target mask so the model cannot see future states.

        Args:
            length (int): length of the sequence.

        Returns:
            torch.Tensor: mask of shape length x length.
        """
        return torch.triu(torch.ones((length, length), dtype=torch.bool), diagonal=1)
```

kylebgorman:
> Hard error on the transducer: `RuntimeError: could not create a primitive descriptor for an LSTM forward propagation primitive`. No idea what this means. (Will check to see if same error is present at Yoyodyne HEAD.)

So, the breakage of the transducer can be isolated to the Torch migration. (It works still at HEAD.) But the error we get is basically useless so I'm not sure what to do with it. That seems like a major blocker.

kylebgorman:
> We can solve it in
> 
> by changing `generate_square_subsequent_mask` to

This works, thanks.

kylebgorman:
> So, the breakage of the transducer can be isolated to the Torch migration. (It works still at HEAD.) But the error we get is basically useless so I'm not sure what to do with it. That seems like a major blocker.

One more update on this: it only shows up on CPU with certain combinations of precision, so I think it's just a bad error message saying that it doesn't support that combination of accelerators and precisions.

kylebgorman:
All my local tests pass so I am converting to a true PR. @Adamits lmk what you think when you get a chance.

Adamits:
> One more update on this: it only shows up on CPU with certain combinations of precision, so I think it's just a bad error message saying that it doesn't support that combination of accelerators and precisions.

Out of curiosity could you share your script? I ran transducer on my mac with torch 2.2.0 and it ran fine.

Adamits (approved):
LGTM

kylebgorman:
> Out of curiosity could you share your script? I ran transducer on my mac with torch 2.2.0 and it ran fine.

To trigger this, try:

* on CPU, 16-bit precision (so `--precision 16`) or "bf16" precision `--precision bf16`). 
* on GPU, "bf16" precision triggers a different (but conceptually similar) error.

I don't think any of the other options matter re: this.

PR #175, opened 2024-04-09
Metrics helpers

kylebgorman:
This adds a small `metrics` struct which stores the information that `loss` should be minimized, `accuracy` should be maximized, and generates appropriate file names for checkpoints. It is then used in `train.py` to implement the checkpointing callback and the patience callback, and in `schedulers.py` to implement the reduce-on-plateau scheduler.

Whereas #171 used the same flag (`--val_metric`) to do checkpointing and patience, offline discussion concluded that we want these to be separate. Naturally the reduce-on-plateau scheduler was also left separate. 

Some flags have been renamed, and I am happy to discuss this further:

* `--val_metric` is now `--checkpoint_metric` and `--patience_metric`.
* `--reduceonplateau_mode` is now `--reduceonplateau_metric` to match the previous.
* I am still bothered by `--save_top_k` so I have proposed another alternative: `--num_checkpoints`. Hard to get clearer than that.

Since there has been a decent amount of plumbing changes here, I'd appreciate a closer review. However, my local tests seem just fine. 

This PR is a followup to PR #171 and a further response to issue #170.


yoyodyne/train.py
```diff
@@ -36,59 +36,56 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
     """Creates the callbacks.
 
     We will reach into the callback metrics list to picks ckp_callback to find
     the best checkpoint path.
 
     Args:
-        patience (int, optional).
+        num_checkpoints (int, optional): number of checkpoints to save. To
+            save one checkpoint per epoch, use `-1`.
+        checkpoint_metric (string, optional): the
```
Adamits, line 52: Missing comment!

yoyodyne/train.py
```diff
@@ -36,59 +36,56 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
+        patience (int, optional): number of epochs with no
+            progress (according to `patience_metric`) before triggering
+            early stopping.
+
+
         val_metric (str, optional).
-        save_top_k (int, optional)....
+        save_top_k (int, optional).
```
Adamits, line 59: delete

yoyodyne/train.py
```diff
@@ -36,59 +36,56 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
-        patience (int, optional).
+        num_checkpoints (int, optional): number of checkpoints to save. To
+            save one checkpoint per epoch, use `-1`.
+        checkpoint_metric (string, optional): the
+        patience (int, optional): number of epochs with no
+            progress (according to `patience_metric`) before triggering
+            early stopping.
+
+
         val_metric (str, optional).
```
Adamits, line 58: delete

yoyodyne/train.py
```diff
@@ -36,59 +36,56 @@ def _get_logger(experiment: str, model_dir: str, log_wandb: bool) -> List:
...
 
     We will reach into the callback metrics list to picks ckp_callback to find
     the best checkpoint path.
 
     Args:
-        patience (int, optional).
+        num_checkpoints (int, optional): number of checkpoints to save. To
+            save one checkpoint per epoch, use `-1`.
+        checkpoint_metric (string, optional): the
+        patience (int, optional): number of epochs with no
```
Adamits, line 53: 1) I think we need a comment for `patience_metric`
2) Should patience be an attribute of the patience metric object? I think probably no given the design, but just want to double check this is intentional.

Adamits: Actually ignore 2)---this is synonymous to how we have a separate `num_checkpoints` variable. The `Metric` is only concerned about the metric specification, not the specific interval.

Adamits (commented):
Left a few notes, but in general this looks great!

yoyodyne/train.py
```diff
@@ -7,7 +7,7 @@
 import wandb
 from pytorch_lightning import callbacks, loggers
 
-from . import data, defaults, models, schedulers, util
+from . import data, defaults, models, metrics, schedulers, util
```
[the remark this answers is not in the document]
kylebgorman: Good call, fixed.

yoyodyne/train.py
kylebgorman: fixed

yoyodyne/schedulers.py
```diff
@@ -133,22 +133,22 @@ def __init__(
...
                 improvement before reducing LR.
             min_learning_rate (float): lower bound on the learning rate.
             **kwargs: ignored.
         """
+        self.metric = metrics.ValidationMetric(reduceonplateau_metric)
         super().__init__(
             optimizer,
-            mode="min" if reduceonplateau_mode == "loss" else "max",
             factor=reduceonplateau_factor,
-            patience=reduceonplateau_patience,
```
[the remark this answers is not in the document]
kylebgorman: That looks to me like a bug so I have reinstated it.

yoyodyne/train.py
kylebgorman: +1 yeah

kylebgorman:
Just pushed fixes to everything: you want to take a second look before I merge? (Give me the Approve radio button under the review thingy so I can merge without override...)

Adamits (approved):
Gave it a quick scroll through after your updates. LGTM

Issue #183, opened 2024-05-01
Add more evaluation metrics


Adamits:
Currently, we only support evaluating with exact match accuracy. We should support more evaluation metrics and probably require a framework for how they should be added.

kylebgorman:
Closed in #184.

PR #184, opened 2024-05-01
WIP: Additional evaluators

Adamits:
Addresses #183 

I was working on evaluating with character error rate, and went down a small rabbit hole of how to support multiple eval metrics. I am very open to different proposals, especially given that I was not thinking about #175 when writing this (we may need to resolve the semantics of this, at the very least). Basically, the idea is that every eval metric should have some definition of a per-sample score (for accuracy this is binary), and we simply collect a list of these across samples, and again across batches, before reporting a micro-average.

There is some controller code that may be a bit clunky right now that entails reading the requested metrics from the command line, and then initializing those evaluators on the model. Then, during validation, we just loop over the requested evaluators and aggregate their metric.

### CER

In doing this, I also implemented a CEREvaluator. This is currently a bit rough---I just used `torchmetrics` for the actual metric, which I believe is all in python. Additionally, I did not bother decoding gold/predicted tensors back into strings and instead convert the list of integers into a string, where each integer is white-space seperated byt he next integer. I *think* this works fine conceptually, but of course we cannot ensure then that this is *really* character error rate---it is essentially  whatever-you-define-one-symbol-to-be error rate, which in my case happens to be character (or phoneme).

@kylebgorman I was hoping you might have strong opinions on how we need to implement cer (we probably need to convert tensors to strings at each eval call, but if there are tricks s.t. we do not have to that would be great).

Please also feel free to comment on the general framework here. Even just critique with no alternate proposal would be very helpful as I just basically followed how we have been doing eval, but I am happy to rethink the system at a higher level.


kylebgorman (commented):
Thanks for this.

I think what's done here makes sense. 

I think for CER (and accuracy too) you should work with the tensors, or lists of integer labels if the data becomes ragged after you take care of EOS stuff, instead of strings if at all possible. This is less work and ought to save you some plumbing. So I would say the joining ought not to be necessary---you just need to nest the list some more. Is there a technical reason why that's not correct?

My question is yes, how does this interact with #175? I think you're just adding evaluation metrics to the logging, essentially. Then you should be able to specify:

1. as many --eval_metrics as are available and you want (default: just accuracy)
2. one --checkpoint_metric (default: accuracy)
3. one --reduceonplateau_metric (default: loss)
4. one --patience_metric (default: loss)

Any issues there? 

I put some notes below on the implementation.

yoyodyne/evaluators.py
```diff
@@ -13,16 +19,15 @@ class Error(Exception):
...
-    num_predicted: int
+    per_sample_metrics: List[float]
 
     @property
-    def accuracy(self) -> float:
-        """Computes the accuracy."""
-        return self.num_correct / self.num_predicted
+    def average_metric(self) -> float:
+        """Computes the micro-average of the metric."""
+        return sum(self.per_sample_metrics) / len(self.per_sample_metrics)
```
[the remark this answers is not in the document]
Adamits: That's right. Good point about using `statistics.mean`!

> Also should this really be a property?

Ah, I had a property `accuracy` before, and was thinking of `average_metric` as a noun to replace it. I guess this reads as a verb though. Maybe I can make it a method `get_average_metric()`?

yoyodyne/evaluators.py
```diff
@@ -31,27 +36,27 @@ def __add__(self, other_eval: EvalItem) -> EvalItem:
             EvalItem.
         """
         return EvalItem(
-            self.num_correct + other_eval.num_correct,
-            self.num_predicted + other_eval.num_predicted,
+            self.per_sample_metrics + other_eval.per_sample_metrics
         )
 
-    def __radd__(self, start_val: int) -> EvalItem:
+    def __radd__(self, start_int: int) -> EvalItem:
```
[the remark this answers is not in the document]
Adamits: Actually I will have to think harder about this works, but iirc, basically if you want to call `sum([object1, object2])`, python first calls `__radd__` on object one to compute the left most summand, so we need to return something.

yoyodyne/evaluators.py
```diff
@@ -155,3 +221,130 @@ def finalize_predictions(
...
+    """Evaluates character error rate."""
+
+    def get_eval_item(
+        self,
+        predictions: List[List[str]],
+        golds: List[List[str]],
+        pad_idx: int,
+    ) -> EvalItem:
+        cer_calc = CharErrorRate()
+        cers = []
```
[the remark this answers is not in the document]
Adamits: Good point. I think I did it this way b/c I think its what the `torchmetrics.CharErrorRate` seemed to expect. I can check their docs and try to avoid this.

yoyodyne/evaluators.py
```diff
@@ -155,3 +221,130 @@ def finalize_predictions(
...
+def add_argparse_args(parser: argparse.ArgumentParser) -> None:
+    """Adds LSTM configuration options to the argument parser.
+
+    Args:
+        parser (argparse.ArgumentParser).
+    """
+    parser.add_argument(
+        "--eval_metrics",
+        action="append",
+        choices=list(_eval_factory.keys()),
```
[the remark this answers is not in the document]
Adamits: I will test. I know `.keys()` is its own iterable type that I sometimes have issues with so I probably did this reflexively (but maybe unnecessarily).

Adamits:
Thanks for the comments.

I was writing you an email about related things, but might move that to here as it relates to some of your comments. Probably won't get back to addressing this until tomorrow though.

yoyodyne/evaluators.py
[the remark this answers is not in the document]
kylebgorman: `.metric` as a property would work, or maybe `get_averaged_metric` if you think it's verby. (I like "averaged" or "aggregated" or something like that).

yoyodyne/evaluators.py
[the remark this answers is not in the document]
kylebgorman: Okay, actually I realize there is a semantic ambiguity here. I think we could have the following, in theory:

* "symbol error rate" or SER which is done at the tensor level
* actually join the targets using `--target_sep` (which has to be passed in for correctness) and then computing true CER

I suspect when there's a difference most people want the former.

Adamits:
> 1. as many --eval_metrics as are available and you want (default: just accuracy, other option is CER and you can have both)
> 2. one --checkpoint_metric (default: accuracy, but why not add CER to that too in addition to loss?)
> 3. one --reduceonplateau_metric (default: loss, other option accuracy)
> 4. one --patience_metric (default: loss, other option accuracy)
>
> Any issues there? I put some notes below on the implementation.

You're right, it should work fine!

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: Right! This is what I meant with:

> it is essentially whatever-you-define-one-symbol-to-be error rate, which in my case happens to be character (or phoneme).

In the second bullet, the joining entails decoding into strings right? In that case I agree. We **could** have an option for both, though the naming could get confusing... E.g. I might ask for cer because I know what it is, not realizing that ser will be the same, but computed faster.

I do like the idea of getting metrics faster by leaving them as tensors, but I think there could be many such cases where we might want to evaluate against strings, and for a developer, adding a new evaluator is probably easier if you can get strings rather than figuring out the right tensor logic. Maybe this is why e.g. huggingface seems to decode into strings by default.

Another example eval that I need to implement is transliteration, where typically for one word you have several valid targets. Eval can then compute the number of top-k hypothesis that are in the list of targets, or do some fuzzy matching, or ranking, etc. This could all be fine in that a particular evaluator could just decode into strings if need be, but then we have to potentially decoding the same preds into strings many times, and maybe the same golds as well!

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: Ok, what is actually happening here is that when we call `sum(list)`, there is a second argument for the "start" term in the summation, which defaults to 0: `sum(list, start=0)`. `__radd__` is called because `int + EvalItem` is not defined. 

So we could instead use sum like `sum(v[metric_name] for v in validation_step_outputs, EvalItem(np.array()))`---though of course this needs to allocate an empty array, which is kind of a waste.

This is hacky though and makes me think we should reduce the list of per-batch eval metrics with a different function. Maybe I can just unroll the numpy vectors of metrics in the `per_sample_metrics` and call mean there... Though I *did* like combining them in a way that returns a new EvalItem.

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: You were right :D

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: Did you mean to edit my comment haha?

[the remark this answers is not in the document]
kylebgorman: Oh god no sorry I meant to quote-reply to it with that.

yoyodyne/evaluators.py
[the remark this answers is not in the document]
kylebgorman: I think the empty-array solution is what I'd reach for, maybe?

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: Just gonna repost the original down here:

Right! This is what I meant with:

it is essentially whatever-you-define-one-symbol-to-be error rate, which in my case happens to be character (or phoneme).

In the second bullet, the joining entails decoding into strings right? In that case I agree. We could have an option for both, though the naming could get confusing... E.g. I might ask for cer because I know what it is, not realizing that ser will be the same, but computed faster.

I do like the idea of getting metrics faster by leaving them as tensors, but I think there could be many such cases where we might want to evaluate against strings, and for a developer, adding a new evaluator is probably easier if you can get strings rather than figuring out the right tensor logic. Maybe this is why e.g. huggingface seems to decode into strings by default.

Another example eval that I need to implement is transliteration, where typically for one word you have several valid targets. Eval can then compute the number of top-k hypothesis that are in the list of targets, or do some fuzzy matching, or ranking, etc. This could all be fine in that a particular evaluator could just decode into strings if need be, but then we have to potentially decoding the same preds into strings many times, and maybe the same golds as well!

yoyodyne/evaluators.py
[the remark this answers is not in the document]
Adamits: Ok. Before I do that---do you think there is a more intuitive python function than `sum`/`add` for the operation we define here? Maybe `extend`, or perhaps something from `itertools` would be clearer e.g. https://docs.python.org/3/library/itertools.html#itertools.chain.from_iterable?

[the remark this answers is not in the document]
Adamits: Another note here:

Allocating a python List is much faster than allocating a numpy array, but numpy.mean is much faster than statistics.mean. I think it is worth using Lists in this tradeoff as I think we will call mean only once per eval epoch, whereas we will allocate Lists/arrays a lot.

Adamits:
I updated how we compute CER a bit and accidentally left it in an unintuitively named commit.

I just import the torchmetrics edit distance computation now (we could replace it with something faster), and then define a `_compute_cer` that calls it once. This avoids the unnecessary overhead of converting to strings, and always expects a single comparison at once.

kylebgorman:
I think numpy’s mean is fine. Anything to reduce complexity in EvalItem…

yoyodyne/evaluators.py
```diff
@@ -80,10 +77,75 @@ def evaluate(
...
         return self.get_eval_item(predictions, golds, pad_idx)
 
-    @staticmethod
     def get_eval_item(
+        self,
+        predictions: torch.Tensor,
+        golds: torch.Tensor,
+        pad_idx: int,
+    ) -> EvalItem:
+        """Converts finalized predictions and golds into the metric.
```
kylebgorman, line 89: I don't know what people normally do here but I think you could safely omit the docs on the ABC? Any thoughts?

yoyodyne/evaluators.py
```diff
@@ -155,3 +213,136 @@ def finalize_predictions(
...
+
+    def _finalize_tensor(
+        self,
+        tensor: torch.Tensor,
+        end_idx: int,
+        pad_idx: int,
+    ) -> List[List[str]]:
+        # Not necessary if batch size is 1.
+        if tensor.size(0) == 1:
+            return [numpy.char.mod("%d", tensor.cpu().numpy())]
```
kylebgorman, line 261: This might merit a comment explaining what is happening here.

yoyodyne/evaluators.py
```diff
@@ -155,3 +213,136 @@ def finalize_predictions(
...
+        out = []
+        for i, prediction in enumerate(tensor):
```
kylebgorman, line 263: did I miss something or is `i` not being used here?

yoyodyne/evaluators.py
```diff
@@ -155,3 +213,136 @@ def finalize_predictions(
...
+    Args:
+        eval_metric (str).
+
+    Raises:
+        Error.
+
+    Returns:
+        Evaluator.
+    """
+    if eval_metric not in _eval_factory:
```
kylebgorman, line 331: The slightly more Pythonic, and faster, thing to do here is:

```python
try:
    return _eval_factory[eval_metric]
except KeyError(key):
    raise Error(f"No evaluation metric {key}")
```

This is thought better because you're not doing the lookup twice (even if it is constant-time...)

kylebgorman (commented):
This looks pretty good to me (some tiny nits/suggestions) but maybe you want to play with the `__add__` stuff more, up to you? (And if you want move to `numpy.mean` fine with me.)

yoyodyne/evaluators.py
Adamits: Yeah I actually have no idea either. I think ill remove it.

yoyodyne/evaluators.py
Adamits: Yeah I usually do that, for some reason I was thinking I had been told somewhere that the if statement was considered clearer or cleaner, but you are right, I will change this!

Adamits:
Ok I think I have addressed the changes. 

Later I can work on another PR to wire up an option for decoding symbols at validation. This will support e.g. actual CER if you have >character-sized symbols. I think it will also be nice from a developer standpoint: if I want to use yoyodyne for my task and need to implement a new eval, I can just ask for characters and implement it in a straightforward way rather than needing to take the time to figure out how to make pytorch tensors do what I want.

kylebgorman:
> Ok I think I have addressed the changes.
> 
> Later I can work on another PR to wire up an option for decoding symbols at validation. This will support e.g. actual CER if you have >character-sized symbols. I think it will also be nice from a developer standpoint: if I want to use yoyodyne for my task and need to implement a new eval, I can just ask for characters and implement it in a straightforward way rather than needing to take the time to figure out how to make pytorch tensors do what I want.

My suggestion there is to just `"".join(symbols)` since it feels like separators ought not to count for any sensible application of CER?

yoyodyne/evaluators.py
```diff
@@ -1,28 +1,36 @@
...
+import abc
+import argparse
 import dataclasses
+from typing import List
 
+import numpy
 import torch
 from torch.nn import functional
 
+# from torchmetrics.text import CharErrorRate
```
kylebgorman, line 13: Remove?

yoyodyne/evaluators.py
```diff
@@ -1,28 +1,36 @@
...
+from torchmetrics.functional.text.helper import _edit_distance
```
kylebgorman, line 14: It occurs to me that this is protected and we're using it anyways we ought to idk, alias it, or just use our own, or use a different third-party one? I have this one lying around which uses a numpy array instead of Python lists if you want it. 

```python
def _edit_distance(x: Labels, y: Labels) -> int:
    idim = len(x) + 1
    jdim = len(y) + 1
    table = numpy.zeros((idim, jdim), dtype=numpy.uint16)
    table[1:, 0] = 1
    table[0, 1:] = 1
    for i in range(1, idim):
        for j in range(1, jdim):
            if x[i - 1] == y[j - 1]:
                table[i][j] = table[i - 1][j - 1]
            else:
                c1 = table[i - 1][j]
                c2 = table[i][j - 1]
                c3 = table[i - 1][j - 1]
                table[i][j] = min(c1, c2, c3) + 1
    return int(table[-1][-1])
```

but don't hold the PR up for just that.

kylebgorman (approved):
LGTM. One last note. Let me know if you want me to just merge.

yoyodyne/evaluators.py
Adamits: I can test it really quick?

kylebgorman: Sure. You can turn it up to `uint32` if you expect strings longer than 65k.

kylebgorman: That's worrisome if only because i've been using this basic implementation for a very long time and in many places...

kylebgorman: Their lines 340-343 are different than my:

```python
table[1:, 0] = 1
table[0, 1:] = 1
```

And I think theirs might be right...that should be increasing.

Adamits: Sorry, have some meetings now but will return to this in an hour or two.

Adamits: Ok as a sanity check, if I do this instead:

```
table = numpy.zeros((idim, jdim), dtype=numpy.uint16)
        # table[1:, 0] = 1
        # table[0, 1:] = 1
        for i in range(idim):
            table[i][0] = i
        for j in range(jdim):
            table[0][j] = j
```

I reproduce the results. Off the top of my head I think this one is right, and your implementation might be for LCS? I will try to find references...

kylebgorman:
Yea, instead of = 1 it should be range(1, …). I tested and confirmed.

yoyodyne/evaluators.py
kylebgorman: Yes, I think your diagnosis is correct. It also occurred to me that you could use `empty` rather than `zeros` that way. I put it into a gist: 
https://gist.github.com/kylebgorman/14cc4e238bc6d80784df4511c29b3d55.

yoyodyne/evaluators.py
```diff
@@ -203,10 +200,27 @@ def _compute_ser(
...
+        for i in range(1, idim):
+            for j in range(1, jdim):
+                if x[i - 1] == y[j - 1]:
+                    table[i][j] = table[i - 1][j - 1]
+                else:
+                    c1 = table[i - 1][j]
+                    c2 = table[i][j - 1]
+                    c3 = table[i - 1][j - 1]
+                    table[i][j] = min(c1, c2, c3) + 1
+        return table[-1][-1]
```
kylebgorman, line 222: Suggest you cast this to `int` before returning because we don't need the `uint16` precision (or any other numpy weirdness). I don't know if it's actually important...

yoyodyne/evaluators.py
Adamits: Not sure why this was still there, I did not have it locally.

kylebgorman (commented):
One quick note...

yoyodyne/evaluators.py
Adamits: Done. (Also updated README)

kylebgorman (approved):
LGTM.

I made one minor change: I made `_edit_distance` a static method. Could you do a very quick check to make sure that doesn't break anything before I merge?

Adamits:
> LGTM.
> 
> I made one minor change: I made `_edit_distance` a static method. Could you do a very quick check to make sure that doesn't break anything before I merge?

Thanks! Ran my test and it seems to work fine.

Issue #165, opened 2024-02-20
Hard Monotonic Transducer


bonham79:
(Adding on issues board for documentation, this PR will be out over the week.)

Wu and Cotterell's papers on strong alignment seem just up our alley for the library. There should be an implementation of https://aclanthology.org/P19-1148/ , particularly the monotonic cases.

Currently I have going a version that allows the following variants:
1. Hard alignment. Non-monotonic.
2. Hard alignment. Monotonic.
3. Hard alignment. First order monotonic.

2) and 3) should be good for most tasks, 1) should be available for more niche cases.

Things to add on to the PR once it's up (this makes sense to me now, will makes sense with accompanying PR).

1. Original implementation stores all alignment and emission probabilities across the decoding as a cache for loss calculation. This seems unnecessary and should be a running sum. 
2. Ideally, the validation step should give the loss for the gold transcription. However, this requires taking into account decoding passes that produce predictions longer/shorter than the gold. Need to find a good way to penalize this while still making the loss intuitable. 
3. Need to double check, but I believe the prediction pass the decoder can be offloaded to our prexisting attention module.

@kylebgorman  @Adamits  Any additional preferences during development? I've been dancing back and forth with adding in the Ahroni and Goldberg Transducer too, just for completeness. (This and the Swissmen's transducer both supercede that one.)

kylebgorman:
My only thought is I am most excited about variant 2.

I thought the A&G thing was outmoded also, but it's harmless if you want to do it later.

Adamits:
Yeah this sounds good. I have a partial implementation of 3) in a fork somewhere from a year ago that I never finished because I got distracted :D. Will be great to see them in here.

iirc 2) and 3) should be small variations on the implementation of 1), right? I.e. in the paper I think 2) is basically 1), but they just enforce the monotonicity constraint in the mask.

> 1. Original implementation stores all alignment and emission probabilities across the decoding as a cache for loss calculation. This seems unnecessary and should be a running sum.

Agree

> 2. Ideally, the validation step should give the loss for the gold transcription. However, this requires taking into account decoding passes that produce predictions longer/shorter than the gold. Need to find a good way to penalize this while still making the loss intuitable.

Not sure I follow---why is this not an issue with existing architectures?

> 3. Need to double check, but I believe the prediction pass the decoder can be offloaded to our prexisting attention module.

I cannot remember---this would imply that all of the constraints are strictly for training, and at inference a regular old soft attention distribution is used?

[...]

PR #186, opened 2024-05-07
Hard monotonic

bonham79:
Wu and Coterrell's hard monotonic transduce..

Closes https://github.com/CUNY-CL/yoyodyne/issues/165.

It builds off lstm module. To call you pass `hmm_lstm` for `--arch`. Hard monotonic constrain is enforced with `--enable_monotonic` while context window is enabled with `--hmm_context`. 

While y'all are merging, could you take a peek at how attention and log_prob contexts are preserved? I hate having that large a tensor filling up memory but when I try to be clever about it it causes model collapse.


transducer.sh
```diff
@@ -0,0 +1,28 @@
+yoyodyne-train --model_dir=../ \
```
kylebgorman, line 1: you probably didn't mean to check this in?

yoyodyne/models/__init__.py
```diff
@@ -32,6 +33,7 @@ def get_model_cls(arch: str) -> BaseEncoderDecoder:
         "pointer_generator_transformer": PointerGeneratorTransformerEncoderDecoder,  # noqa: 501
         "transducer": TransducerEncoderDecoder,
         "transformer": TransformerEncoderDecoder,
+        "hmm_lstm": HardAttentionHmm,
```
kylebgorman, line 36: Can you think of a better flag name that more closely matches the name of the module? What about like, `--arch hard_attention` or `--arch hard_monotonic_lstm`? While I sort of understand the connection to HMMs I don't think people know it under that name.

yoyodyne/models/__init__.py
```diff
@@ -74,6 +76,7 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             "pointer_generator_transformer",
             "transducer",
             "transformer",
+            "hmm_lstm",
```
kylebgorman, line 79: Whatever name you go with, put this list in lexicographic order like it was beforehand.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,418 @@
...
+        Wu, S. and Cotterell, R. 2019. Exact hard monotonic attention for
+        character-level transduction. In _Proceedings of the 57th Annual
+        Meeting of the Association for Computational Linguistics_, pages
+        1530-1537.
+
+    Original implementation:
+        https://github.com/shijie-wu/neural-transducer
+    """
+
+    enforce_monotonic: bool
```
kylebgorman, line 37: While these are optional in the sense that you don't have to pass them in, they're not nullable, so they shouldn't be labeled `Optional`. I took the liberty of fixing this.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,421 @@
...
+
+        Args:
+                encoder_out (torch.Tensor): batch of encoded input symbols.
+                    Shape: (batch_size x src_len x encoder_hidden*num_dir).
+                encoder_mask (torch.Tensor): mask for the batch of encoded
+                    input symbols. Shape: (batch_size x src_len).
+                target (torch.Tensor): target symbols. Decodes up to
+                    `len(target)` symbols.
+
+        Returns:
```
kylebgorman, line 145: Since Python can't return multiple things we say it returns a single thing, a tuple. I suggested how to restyle this for `forward`: can you do that here too?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,421 @@
...
+                all_log_probs: (torch.Tensor): tensor of emission
+                    probabilities for each transition state (target symb).
+                    Shape: (tgt_len x batch_size x src_len x vocab_size)
+                all_alignements: (torch.Tensor): tensor of transition
```
kylebgorman, line 149: *alignements

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,421 @@
...
+                    probabilities for each transition state (target symb).
+                    Shape: (tgt_len x batch_size x src_len x src_len)
+        """
+        log_probs, alignment_matrix, decoder_hiddens = self.init_decoding(
+            encoder_out, encoder_mask
+        )
+        all_log_probs, all_alignments = [log_probs], [alignment_matrix]
+        for t in range(target.size(1)):
```
kylebgorman, line 157: Can you use a more informative variable name than just `t`? I'm not sure what this is, is it length or something?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,421 @@
...
+        """Performs a single decoding step for current state of decoder.
+
+        Args:
+                tgt_symbol (torch.Tensor): tgt symbol for current state.
+                encoder_out (torch.Tensor): batch of encoded input symbols.
+                    Shape: (batch_size x src_len x encoder_hidden*num_dir).
+                encoder_mask (torch.Tensor): mask for the batch of encoded
+                        input symbols of shape (batch_size x src_len).
+
+        Returns:
```
kylebgorman, line 182: ditto: can't return multiple things, and *alignements.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+            # If finished decoding, pads.
+            pred = torch.where(
+                ~finished, pred, torch.tensor(self.end_idx, device=self.device)
+            )
+            predictions = torch.cat((predictions, pred), dim=-1)
+            # Updates likelihood emissions.
+            likelihood = likelihood + self._gather_at_idx(log_probs, pred)
+        return predictions, likelihood
+
+    def greedy_step(self, log_probs, likelihood):
```
kylebgorman, line 263: Needs typing. Shuld this be a static method too? 

Can I suggest `target` instead of `tgt` since we used that elsewhere?'

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+        # Updates current likelihood with greedy probability prediction.
+        tgt_prob = likelihood + log_probs.transpose(1, 2)
+        tgt_prob = tgt_prob.logsumexp(dim=-1)
+        tgt = torch.max(tgt_prob, dim=-1)[1]
+        return tgt.unsqueeze(-1), likelihood
+
+    def _gather_at_idx(self, prob, tgt):
```
kylebgorman, line 270: ditto from previous note.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+        # Collects probability of tgt index across all states in prob.
+        batch_size, src_seq_len, vocab_size = prob.shape
+        idx = tgt.view(-1, 1).expand(batch_size, src_seq_len).unsqueeze(-1)
+        output = torch.gather(prob, -1, idx).view(batch_size, 1, src_seq_len)
+        idx = idx.view(batch_size, 1, src_seq_len)
+        pad_mask = (idx != self.pad_idx).float()
+        return output * pad_mask
+
+    def _apply_mono_mask(self, alignment):
```
kylebgorman, line 279: ditto

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+                bidirectional=False,
+                embedding_size=self.embedding_size,
+                layers=self.decoder_layers,
+                hidden_size=self.hidden_size,
+            )
+
+    def _get_loss_func(
+        self,
+    ) -> Callable[[torch.Tensor, torch.Tensor, torch.Tensor], torch.Tensor]:
+        # TODO: Make the loss update continuous during training to avoid.
```
kylebgorman, line 372: I don't understand this TODO because I don't know what you're avoiding.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+        # Also enable better reflection of val loss.
+        """Returns the actual function used to compute loss.
+
+        Returns:
+                Callable[[torch.Tensor, torch.Tensor], torch.Tensor]:
+                    configured loss function.
+        """
+        return self.loss
+
+    def loss(self, target, log_probs, alignments):
```
kylebgorman, line 382: needs typing

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,413 @@
...
+        fwd = alignments[0, :, 0].unsqueeze(1) + self._gather_at_idx(
+            log_probs[0], target[:, 0]
+        )
+        for t in range(1, target.shape[1]):
```
kylebgorman, line 386: more informative variable name?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
+"""Hard monotonic neural HMM classes."""
```
kylebgorman, line 1: In at least this file it looks like you're using tabs. Please standardize on 4 spaces instead. `s/\t/    /g` ought to fix things.

kylebgorman (commented):
This is pretty cool! Right now my comments are mostly about naming and style, and all quite minor. I would like Adam to weigh in how it interacts with the other modules/models (like, should it be in `models/modules/lstm.py`? Is it DRY?) next, then we can all test.

Before this is done we'll want to update the README too.

yoyodyne/models/__init__.py
bonham79: Yeah I'm just trying not to confuse people since we also have `attentive lstm`. Honestly their architecture makes more sense just saying it's an HMM. Couldn't we just clarify in the readme?

yoyodyne/models/hmm.py
bonham79: It can be static. I prefer the abbreviated variable name.

yoyodyne/models/hmm.py
bonham79: I feel like this is pretty conventional. It's a decode function for LSTM. We're pretty consistent with using 1 indexing for length.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+                all_alignements: (torch.Tensor): tensor of transition
+                    probabilities for each transition state (target symb).
+        """
+        decoded = self.decoder(
+            tgt_symbol,
+            decoder_hiddens,
+            encoder_out,
+            encoder_mask,
+        )
+        output, alignment_matrix, decoder_hiddens = (
```
Adamits, line 194: Why are we calling the embeddings the alignment matrix?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+        idx = tgt.view(-1, 1).expand(batch_size, src_seq_len).unsqueeze(-1)
+        output = torch.gather(prob, -1, idx).view(batch_size, 1, src_seq_len)
+        idx = idx.view(batch_size, 1, src_seq_len)
+        pad_mask = (idx != self.pad_idx).float()
+        return output * pad_mask
+
+    def _apply_mono_mask(self, alignment):
+        # Applies mask to enforce monotonic alignment.
+        align_mask = torch.ones_like(alignment[0]).triu().unsqueeze(0)
+        align_mask = (align_mask - 1) * -numpy.log(1e-7)
```
Adamits, line 282: For my own understanding, can you explain why we want 0, -1 here rather than 0, 1? Otherwise I understand whats going on and it is really really nicely implemented :).

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+                all_alignements: (torch.Tensor): tensor of transition
+                    probabilities for each transition state (target symb).
+                    Shape: (tgt_len x batch_size x src_len x src_len)
+        """
+        log_probs, alignment_matrix, decoder_hiddens = self.init_decoding(
+            encoder_out, encoder_mask
+        )
+        all_log_probs, all_alignments = [log_probs], [alignment_matrix]
+        for t in range(target.size(1)):
+            tgt_symbol = target[:, t]
```
Adamits, line 158: I think we need to address the `teacher_forcing` flag somehow. If it does not make sense to turn off teacher forcing in this architecture, maybe we should log a warning in case it is turned off?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+            if finished.all().item():
+                break
+            # If finished decoding, pads.
+            pred = torch.where(
+                ~finished, pred, torch.tensor(self.end_idx, device=self.device)
+            )
+            predictions = torch.cat((predictions, pred), dim=-1)
+            # Updates likelihood emissions.
+            likelihood = likelihood + self._gather_at_idx(log_probs, pred)
+        return predictions, likelihood
```
Adamits, line 261: Do we actually use the predictions later? I guess it could be convenient to return them since we already computed them, but iirc they are not needed after this in the current state of this codebase?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+
+    def greedy_step(self, log_probs, likelihood):
+        # Updates current likelihood with greedy probability prediction.
+        tgt_prob = likelihood + log_probs.transpose(1, 2)
```
Adamits, line 265: Is likelihood actually a log likelihood representing alignments? I wonder if we can name these more clearly. Is `alignment_log_likelihoods` right? Maybe too verbose though even if right...

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+        )
+        return loss
+
+    def validation_step(self, batch: data.PaddedBatch, batch_idx: int) -> Dict:
+        predictions, likelihood = self(batch)
+        # Processes for accuracy calculation.
+        predictions = self.evaluator.finalize_predictions(
+            predictions, self.end_idx, self.pad_idx
+        )
+        val_eval_item = self.evaluator.get_eval_item(
```
Adamits, line 317: There are some updates here in master. I can fix this up later though.

yoyodyne/models/hmm.py
Adamits: This method is here to keep with the base model interface, it should only be called once when the model is initialized and just tells the model what to use as `self.loss_func`. Since you override all that stuff I guess you don't need it, but I don't see why removing it would benefit you?

yoyodyne/models/hmm.py
Adamits: Also, A comment in the code, whether in the method signature or elsewhere, about why we need this would be really helpful!

yoyodyne/models/hmm.py
Adamits: Edit: I see you do make use of them in overridden methods, which seems good.

Adamits (changes requested):
Reviewed up until the `modules` changes. I can hopefully get back to it today.

yoyodyne/models/hmm.py
kylebgorman: (If you keep this) just move explanation into TODO to increase bus factor

yoyodyne/models/hmm.py
kylebgorman: Well, what is `t`? "time"?

yoyodyne/models/__init__.py
kylebgorman: We could, but I think it would be better to call it what the previous literature calls it instead.

kylebgorman:
Broader question: given that Wu & Cotterell didn't find an improvement for going from zeroth to first order, would we want to consider getting rid of the contextual form? I think Adam has had the philosophy that it's nice to have stuff that doesn't work around. I am not sure (it adds a maintainence burden) but I haven't actively disagreed. I don't feel strongly either way but I wanted to start the conversation...

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
         return "attentive LSTM"
+
+
+class HMMLSTMDecoder(LSTMDecoder):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.output_proj = nn.Sequential(
+            nn.Linear(self.output_size, self.output_size), nn.Tanh()
+        )
+        self.scale_encoded = nn.Linear(
```
Adamits, line 234: Can you add a comment about what it is for?

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
     @property
     def name(self) -> str:
         return "attentive LSTM"
+
+
+class HMMLSTMDecoder(LSTMDecoder):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.output_proj = nn.Sequential(
```
Adamits, line 231: Was going to comment that we should try to make this named the same as in other architectures (I probably was not consistent either, but it'd be good to start :)), however since this does something slightly different---adds an activation, and then is combined with emission probs---maybe having a different name is good. Just wanted to check if you thought about this at all?

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
+            nn.Linear(self.output_size, self.output_size), nn.Tanh()
+        )
+        self.scale_encoded = nn.Linear(
+            self.decoder_input_size, self.hidden_size
+        )
+
+    def _alignment_step(self, decoded, encoder_out, encoder_mask):
```
Adamits, line 238: Needs typing + comment.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
+        # Multiplies encoding and decoding for alignment representations.
+        # -> B x seq_len.
+        alignment_scores = torch.bmm(
+            self.scale_encoded(encoder_out), decoded.transpose(1, 2)
+        ).squeeze(-1)
+        # Gets probability of alignments.
+        alignment_probs = nn.functional.softmax(alignment_scores, dim=-1)
+        # Mask padding.
+        alignment_probs = alignment_probs * (~encoder_mask) + 1e-7
```
Adamits, line 247: Why can't we do normal masking here? That is, I understood the smoothing in hmm.py to be for enforcing monotonicity in a differentiable way, but I don't think that's what were doing here?

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
+
+class ContextHMMLSTMDecoder(HMMLSTMDecoder):
+    def __init__(self, *args, attention_context, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.delta = attention_context
+        self.alignment_proj = nn.Linear(
+            self.hidden_size * 2, (attention_context * 2) + 1
+        )  # Window size must include center and both sides.
+
+    def _alignment_step(self, decoded, encoder_out, encoder_mask):
```
Adamits, line 321: Needs typing + comment.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
+    def name(self) -> str:
+        return "HMM LSTM"
+
+
+class ContextHMMLSTMDecoder(HMMLSTMDecoder):
+    def __init__(self, *args, attention_context, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.delta = attention_context
+        self.alignment_proj = nn.Linear(
+            self.hidden_size * 2, (attention_context * 2) + 1
```
Adamits, line 318: Does this mean this implements strictly a 1st order HMM? Could be good to note that.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,167 @@ def forward(
...
+        )
+        # Get probability of alignments.
+        # Mask padding.
+        alignment_probs = alignment_probs * (~encoder_mask).unsqueeze(1) + 1e-7
+        alignment_probs = alignment_probs / alignment_probs.sum(
+            dim=-1, keepdim=True
+        )
+        return alignment_probs.log()  # Log probs for quicker computation
+
+    def forward(
```
Adamits, line 351: I think this can just inherit from the other HMM class no?

Adamits (changes requested):
Ok left more comments

Regarding

> like, should it be in models/modules/lstm.py? Is it DRY?

...

> It shouldn't be another module (we have enough). It's just an expansion of LSTM behavior. If we're interested in expanding it to general transformer behavior then we can generalize it as a mixin. But for first PR we can keep in LSTM.

I guess since Travis implemented the refactor for distinguishing models and modules I can defer to where he thinks things should go. That being said, I do not agree that "we have enough" is a good argument to keep things out of modules. It feels a bit weird to have a new file for HMM models, but no file for HMM modules given that some new HMM modules have been added. I suppose the thing that is a bit weird is that most models really refer to a specific decoder, and are thus coupled to a decoder Module, which then are not actually very general right?

My opinion is that this is a matter of what would be most intuitive for a user of the library when they want to implement a new model? 

I also do not feel very clear as to what

> If we're interested in expanding it to general transformer behavior then we can generalize it as a mixin

means. It sounds like it could mean introducing yet another abstraction pattern, which I am skeptical of. I guess there are enough inherent differences in the decoders that I would want to approach it by just writing a new model class entirely.

transducer.sh
bonham79: no. no i did not. 

minor rant: have you noticed that

yoyodyne/models/__init__.py
bonham79: alrighty. though as minor note will mention this is what ryan and shinje call it.

bonham79: Gonna go with `hard_attention_lstm` and then revisit if we want to add a transformer head in its place.

yoyodyne/models/hmm.py
bonham79: t for each t step in target.

yoyodyne/models/hmm.py
bonham79: Which forward?

yoyodyne/models/hmm.py
bonham79: target is used for the full string, tgt implies a simpler version. don't want to confuse the two. can do tgt_char if preferred?

yoyodyne/models/hmm.py
bonham79: Oh, it's just a sophisticated form of gather since it needs to aggregate across all states for the pseudo hmm calculation. I'm kinda not a fan of it and just cribbed from original code. need to figure out a less painful one, so skimped on documentation.

yoyodyne/models/hmm.py
bonham79: @Adamits Oh, I still need to keep the _get_loss_func. My point is that at the moment the model builds a psssh, len x len x vocab_size x num_steps size tensor to track emissions and transitions. I need to simplify this.

@kylebgorman ???

yoyodyne/models/hmm.py
bonham79: Why are we using the four space convention?

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+        decoder_hiddens = self.init_hiddens(batch_size, self.decoder_layers)
+        bos = (
+            torch.tensor(
+                [self.start_idx], device=self.device, dtype=torch.long
+            )
+            .repeat(batch_size)
+            .unsqueeze(-1)
+        )
+        return self.decode_step(
+            bos, decoder_hiddens, encoder_out, encoder_mask
```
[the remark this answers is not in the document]
bonham79: I'm just trying to modularize and make things easier to read. When I had this in the block the extra comments to explain the first free step just made things messy.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor]: tensosr of emission
+                 and transition probabilities for each transition
+                 state (i.e., target symbol).
+        """
+        encoder_out = self.source_encoder(batch.source).output
+        if self.has_features_encoder:
+            encoder_features_out = self.features_encoder(batch.features).output
+            # Averages to flatten embedding.
```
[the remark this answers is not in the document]
bonham79: Ah, good catch. Yeah, they assumed a constant amount of features so I wanted to avoid the projection to avoid some extra code. I've toyed around with this in the past and just adding them seems to do the trick anyhow.

yoyodyne/models/hmm.py
```diff
@@ -0,0 +1,412 @@
...
+        all_log_probs, all_alignments = [log_probs], [alignment_matrix]
+        for t in range(target.size(1)):
+            tgt_symbol = target[:, t]
+            log_probs, alignment_matrix, decoder_hiddens = self.decode_step(
+                tgt_symbol.unsqueeze(-1),
+                decoder_hiddens,
+                encoder_out,
+                encoder_mask,
+            )
+            all_log_probs.append(log_probs), all_alignments.append(
```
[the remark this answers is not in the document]
bonham79: trying to keep it parallel to the initial list befoire the loop. just personal aesthetics.

yoyodyne/models/hmm.py
bonham79: yeah transition probs works better.

yoyodyne/models/hmm.py
bonham79: Lemme revisit this, I was cribbing from Shinje and Ryan.

kylebgorman: Not directly related but maybe you want to do this at the top:

```python
# Note I am using math here because there's no reason to use numpy.
ALIGNMENT_EPSILON = -math.log(1e-7)  
```

and then use that as a constant later.

yoyodyne/models/hmm.py
bonham79: gotcha, i moved it to reduce confusion. wasn't getting the bus factor reference.

yoyodyne/models/hmm.py
bonham79: That makes me sad. Oh well, looks like my code editor auto-converted.

yoyodyne/models/hmm.py
bonham79: @Adamits aight so `triu` sets bottom triangular to 0, and `ones_like` would need to be multiplied by a negative factor to allow 0's and 1's. It would work just the same but I want to say that multiplying the entire matrix by negative one instead of only half is more expensive an operation (I imagine backend pytorch short circuits 0 multiplication). That's only reason I can imagine.

What annoys me is that you need to do the numpy.log thing. When I try a one line `tensor.where(tensor, -float("inf")), it messes up gradient calculations. Any thoughts on that?

yoyodyne/models/hmm.py
bonham79: Do you want a warning or straight assertion failure?

yoyodyne/models/hmm.py
bonham79: It's a bit verbose and not really that informative. The `hmm` parallel really starts to break down at a certain point, especially on decoding. Going too verbose may get misleading.

yoyodyne/models/hmm.py
bonham79: Any major ones?

bonham79:
> Broader question: given that Wu & Cotterell didn't find an improvement for going from zeroth to first order, would we want to consider getting rid of the contextual form? I think Adam has had the philosophy that it's nice to have stuff that doesn't work around. I am not sure (it adds a maintainence burden) but I haven't actively disagreed. I don't feel strongly either way but I wanted to start the conversation...

We should keep. They see results everywhere but G2P iirc the paper correctly. It's also capable of more expressive graphs, which can come in handy for other experiments.

bonham79:
> 
> I guess since Travis implemented the refactor for distinguishing models and modules I can defer to where he thinks things should go. That being said, I do not agree that "we have enough" is a good argument to keep things out of modules. It feels a bit weird to have a new file for HMM models, but no file for HMM modules given that some new HMM modules have been added. I suppose the thing that is a bit weird is that most models really refer to a specific decoder, and are thus coupled to a decoder Module, which then are not actually very general right?
> 

I want to limit the addition of new modules and file names to conservative, 'we have to basis.' I've lost too much time having to scan ESPNet and fairseq libraries to figure out which parts are going where. Modularity is good, but too much modularity is a problem for me. 

My general approach for models is they contain everything necessary for that model. For modules, those should just contain the lego block pieces to customize models. If I have an lstm file, the design idea is that I can use LSTMs wherever I want. If we create a new file for hard_attention, that's implying, to me, that I can apply hard_attention modules wherever I want. Since that is not the case, I'd rather stick them in my lstm box. 

> 
> means. It sounds like it could mean introducing yet another abstraction pattern, which I am skeptical of. I guess there are enough inherent differences in the decoders that I would want to approach it by just writing a new model class entirely.
>

Yeah, that's what I mean. I'd be okay with separating into another file IF we want to generalize hard_attention module decoding for variations on encoders and decoders. (e.g. swap out lstm, transformer, gru, ssm).

yoyodyne/models/modules/lstm.py
bonham79: I'm trying to keep it general to not confuse people on a read through. I commented elsewhere, but the hmm parallel starts to break down at a certain point. If we can just simplify to general (output proj) and the like, it's easier on my brain when debugging.

yoyodyne/models/modules/lstm.py
bonham79: It doesn't work here. I don't know why, but this is what they had to do in the original code. Everytime I replace with normal masking it breaks. I'm just sticking to this for the first commit and cleaning it up once I have a moment to stare at tensor values.

yoyodyne/models/modules/lstm.py
bonham79: not quite, you need the expanded decoded for two operations. lemme see if i can reduce boilerplate.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,148 @@ def forward(
     @property
     def name(self) -> str:
         return "attentive LSTM"
+
+
+class HardAttentionLSTMDecoder(LSTMDecoder):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Activattes emission probs.
```
kylebgorman, line 231: typo: "Activattes"

kylebgorman (commented):
Just following along at this point. One typo I noted below.

yoyodyne/train.py
```diff
@@ -211,6 +212,7 @@ def get_model_from_argparse_args(
         eval_metrics.add(args.patience_metric)
     # Please pass all arguments by keyword and keep in lexicographic order.
     return model_cls(
+        attention_context=args.hard_attention_context,
```
kylebgorman, line 215: Can you lexicographically order this (see comment on line 213) so `attention-context` comes after `arch`? 

(The reason is that it makes it easier to tell if all the necessary arguments are being passed. We have had bugs like that earlier.)

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,148 @@ def forward(
...
+            bidirectional=self.bidirectional,
+        )
+
+    @property
+    def output_size(self) -> int:
+        return self.decoder_input_size + self.hidden_size
+
+    @property
+    def name(self) -> str:
+        return "Hard Attention LSTM"
```
kylebgorman, line 318: These are lowercase elsewhere so for consistency, "hard attention LSTM"

kylebgorman (commented):
This all basically looks fine to me, but I'll defer to Adam for final approval.

I would recommend defining `1e7` as a constant and explaining what it does when you use it, since future-us will have no context for why that's done. It's okay if the comment just says it is following practices elsewhere.

kylebgorman:
> This all basically looks fine to me, but I'll defer to Adam for final approval.
> 
> I would recommend defining `1e7` as a constant and explaining what it does when you use it, since future-us will have no context for why that's done. It's okay if the comment just says it is following practices elsewhere.

Modulo method-level docs that is.

bonham79:
> This all basically looks fine to me, but I'll defer to Adam for final approval.
> 
> I would recommend defining `1e7` as a constant and explaining what it does when you use it, since future-us will have no context for why that's done. It's okay if the comment just says it is following practices elsewhere.

Gonna define the log variant since it's technically an expensive operation and only need value once. Else can do.

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,461 @@
...
+            **kwargs: passed to superclass.
+        """
+        self.enforce_monotonic = enforce_monotonic
+        self.attention_context = attention_context
+        super().__init__(*args, **kwargs)
+        self.classifier = nn.Linear(
+            self.decoder.output_size, self.target_vocab_size
+        )
+        if not self.teacher_forcing:
+            util.log_info(
```
kylebgorman, line 63: I think this should throw an `Error`.

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,461 @@
...
+from typing import Callable, Dict, Optional, Tuple
+
+import numpy
+import torch
+from torch import nn
+
+from .. import data, defaults, util
+from . import lstm, modules
+
+NEG_LOG_EPSILON = -numpy.log(1e-7)
```
kylebgorman, line 13: why not just use `math.log` for this? There's no real need for `numpy` is there?

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,461 @@
...
+    @staticmethod
+    def greedy_step(log_probs, likelihood):
+        # Updates current likelihood with greedy probability prediction.
+        tgt_prob = likelihood + log_probs.transpose(1, 2)
+        tgt_prob = tgt_prob.logsumexp(dim=-1)
+        tgt_char = torch.max(tgt_prob, dim=-1)[1]
+        return tgt_char.unsqueeze(-1), likelihood
+
+    @staticmethod
+    def _gather_at_idx(prob, tgt, pad_idx=None):
```
kylebgorman, line 297: Needs typing and args, etc. in docstring.

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,461 @@
...
+        batch_size, src_seq_len, _ = prob.shape
+        idx = tgt.view(-1, 1).expand(batch_size, src_seq_len).unsqueeze(-1)
+        output = torch.gather(prob, -1, idx).view(batch_size, 1, src_seq_len)
+        idx = idx.view(batch_size, 1, src_seq_len)
+        if pad_idx:
+            pad_mask = (idx != pad_idx).float()
+        return output * pad_mask
+
+    @staticmethod
+    def _apply_mono_mask(transition_prob):
```
kylebgorman, line 313: Needs typing and docstring.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,149 @@ def forward(
...
+                        encoder_mask.shape[1] - (self.delta + 1) - i,
+                    ),
+                )
+                for i, t in enumerate(alignment_probs)
+            ],
+            dim=1,
+        )
+        # Get probability of alignments.
+        # Mask padding.
+        alignment_probs = alignment_probs * (~encoder_mask).unsqueeze(1) + 1e-7
```
kylebgorman, line 363: Can you make `1e-7` another module constant?

kylebgorman (commented):
Just a couple nits. Can't wait to try it.

yoyodyne/models/hard_attention.py
bonham79: I only like throwing Errors if it's going to do behavior that you really shouldn't be doing. If it's something that's just not going to happen I prefer a warning flag.

yoyodyne/models/hard_attention.py
bonham79: Isn't numpy more stable than `math`? Especially since it's being cast for a tensor operation

kylebgorman: I don't have any objection to using `numpy` in general, but here they produce exactly the same value, a `float` scalar which are exactly the same value: 

```python
assert numpy.log(1e7) == math.log(1e7)
```

I think for scalars they both are just calling into the same `libc` function anyways.

yoyodyne/models/modules/lstm.py
bonham79: Are we going to use it elsewhere?

yoyodyne/models/hard_attention.py
bonham79: I'd rather just use numpy for consistency with other libraries. (I see everyone use the numpy log). It's the same number of imports and we need to keep support with the dependency anyhow.

yoyodyne/models/hard_attention.py
bonham79: "really shouldn't be doing" -> you're fundamentally changing your architecture or going to have a long term effect that we should end here. That should throw an error since the alternative is building an architecture that's going to break anyhow.

Here, not doing teaching forcing doesn't change architecture or training. There's no long term effects of just changing the flag.

yoyodyne/models/hard_attention.py
kylebgorman: Internal consistency is something I care about, and we've used `math` [elsewhere in the library](https://github.com/CUNY-CL/yoyodyne/blob/c0640676adc8e25b3e41b5f7c574e7e115381a03/yoyodyne/models/pointer_generator.py#L20), so I will either fix that, or this, at some later date. Go ahead and I'll resolve issue. 

(I don't care about "external consistency" at all.)

yoyodyne/models/modules/lstm.py
kylebgorman: Maybe. (It is used in the model itself, sorta) But that's just one of the two concerns I have for whether something should be a constant. The other one is "is it inscrutable?" If so, putting it at the top and giving it a variable can help. A comment would also maybe help if you'd rather.

yoyodyne/models/hard_attention.py
bonham79: | (I don't care about "external consistency" at all.)
I do, it's easier when adapting codebases and sanity checking stability issues. So i'd recommend just changing occurrences of `math.log` to `numpy.log`

yoyodyne/models/modules/lstm.py
bonham79: It's moved to its own variable and added comment.  Let's keep here until we're seeing frequent use of epsilons in the library, then just move them all to default.

yoyodyne/models/hard_attention.py
kylebgorman: okay do you want to do that in this PR or should I later?

bonham79: Sure, lemme search and find all instances.

yoyodyne/models/hmm.py
Adamits: Yeah https://github.com/CUNY-CL/yoyodyne/pull/184

yoyodyne/models/modules/lstm.py
Adamits: Right I meant that I do not think its consistent with the rest of the repo iirc.

yoyodyne/models/hmm.py
Adamits: Is `alignment` part of the hmm parallel? I just think log_probs is not that informative here. Maybe a method signature will clear it up.

Adamits:
> > I guess since Travis implemented the refactor for distinguishing models and modules I can defer to where he thinks things should go. That being said, I do not agree that "we have enough" is a good argument to keep things out of modules. It feels a bit weird to have a new file for HMM models, but no file for HMM modules given that some new HMM modules have been added. I suppose the thing that is a bit weird is that most models really refer to a specific decoder, and are thus coupled to a decoder Module, which then are not actually very general right?
> 
> I want to limit the addition of new modules and file names to conservative, 'we have to basis.' I've lost too much time having to scan ESPNet and fairseq libraries to figure out which parts are going where. Modularity is good, but too much modularity is a problem for me.
> 
> My general approach for models is they contain everything necessary for that model. For modules, those should just contain the lego block pieces to customize models. If I have an lstm file, the design idea is that I can use LSTMs wherever I want. If we create a new file for hard_attention, that's implying, to me, that I can apply hard_attention modules wherever I want. Since that is not the case, I'd rather stick them in my lstm box.
> 
> > means. It sounds like it could mean introducing yet another abstraction pattern, which I am skeptical of. I guess there are enough inherent differences in the decoders that I would want to approach it by just writing a new model class entirely.
> 
> Yeah, that's what I mean. I'd be okay with separating into another file IF we want to generalize hard_attention module decoding for variations on encoders and decoders. (e.g. swap out lstm, transformer, gru, ssm).

Right agreed. And tbh we already have more modularity than I like :D, I think it is an issue in almost every modern NLP library/toolkit I have used..

yoyodyne/models/modules/lstm.py
Adamits: Hmmm I would be interested to know why it doesn't work (though I would need to dig back through the code to have any opinion anyway, so dont let me hold up the PR here.)

yoyodyne/models/hmm.py
bonham79: Yeah alignment works since it's at least similar to how we think about attention. It is effectively log_probs, so I'd rather keep it in.

yoyodyne/models/hmm.py
bonham79: Ah yeah, I updated it that in a push. Check that in current file.

yoyodyne/models/modules/lstm.py
bonham79: Yeah it's HardAttentionLSTMDecoder now. More consistent?

yoyodyne/models/modules/lstm.py
bonham79: No it's weird. It's next up on the debug because it's annoying that it doesn't work.

bonham79:
> 
> Right agreed. And tbh we already have more modularity than I like :D, I think it is an issue in almost every modern NLP library/toolkit I have used..

yeahhhh, i just accept feature creep as an invetability and try to plan around it. sera sera

bonham79:
@kylebgorman Replaced the `math` with `log`. Found a few redundant uses of `inf` so just created a `defaults.INF` and `defaults.NEG_INF` set of arguments. While at it I also offloaded epsilon to defaults anyhow.

kylebgorman:
> @kylebgorman Replaced the `math` with `log`. Found a few redundant uses of `inf` so just created a `defaults.INF` and `defaults.NEG_INF` set of arguments. While at it I also offloaded epsilon to defaults anyhow.

What you did there looks good. I'll let you and Adam work out the remaining details---ping me when you're ready to merge.

bonham79:
@Adamits Anything major that I haven't commented on yet?

yoyodyne/models/modules/lstm.py
Adamits: I have been commenting on `self.output_proj` this whole time. In other places we have `self.classifier` but it looks like this does something slightly different so ignore me sorry.

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,481 @@
...
+        # 0 log-probability value for masking.
+        # Implemented same as in original repo.
+        mask = (mask - 1) * defaults.NEG_LOG_EPSILON
+        transition_prob = transition_prob + mask
+        transition_prob = transition_prob - transition_prob.logsumexp(
+            -1, keepdim=True
+        )
+        return transition_prob
+
+    def training_step(self, batch: data.PaddedBatch, batch_idx: int) -> Dict:
```
Adamits, line 345: Is this return type correct? Does it not return a tensor (scalar) loss (sorry if this error is just copied from an earlier error))

yoyodyne/models/hard_attention.py
```diff
@@ -0,0 +1,481 @@
...
+    ) -> Callable[[torch.Tensor, torch.Tensor, torch.Tensor], torch.Tensor]:
+        """Returns the actual function used to compute loss.
+
+        Returns:
+                Callable[[torch.Tensor, torch.Tensor], torch.Tensor]:
+                    configured loss function.
+        """
+        return self.loss
+
+    def loss(
```
Adamits, line 440: I think `_loss` might be better here. 1) would we ever want to call it publicly? And 2) self.loss sounds like a property.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,149 @@ def forward(
...
+        # Activates emission probs.
+        self.output_proj = nn.Sequential(
+            nn.Linear(self.output_size, self.output_size), nn.Tanh()
+        )
+        # Project transition probabilities to depth of module.
+        self.scale_encoded = nn.Linear(
+            self.decoder_input_size, self.hidden_size
+        )
+
+    def _alignment_step(
```
Adamits, line 240: I think this and the overridden version of this method would benefit from a method signature.

yoyodyne/models/modules/lstm.py
```diff
@@ -223,3 +223,149 @@ def forward(
...
+class ContextHardAttentionLSTMDecoder(HardAttentionLSTMDecoder):
+    # First order HMM variant of Hard Attention LSTM.
+    def __init__(self, *args, attention_context, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.delta = attention_context
+        self.alignment_proj = nn.Linear(
+            self.hidden_size * 2, (attention_context * 2) + 1
+        )  # Window size must include center and both sides.
+
+    def _alignment_step(
```
Adamits, line 330: Same comment as above

Adamits (approved):
Sorry I left a couple more nits---otherwise LGTM

yoyodyne/models/modules/lstm.py
bonham79: oh my apologies. Oh well we have a better namespace.

mm, do you have a better name? it really is just activating the `output` tensor so output_proj at least lets us know where it should go.

yoyodyne/models/hard_attention.py
bonham79: no you're right. probably an artifact from when i was trying something else out.

yoyodyne/models/hard_attention.py
bonham79: that's a better call.

yoyodyne/models/modules/lstm.py
bonham79: could you clarify? do you mean beyond typing?

kylebgorman (approved):
LGTM. I'm going to merge now.

bonham79:
@kylebgorman revert fly by comment to width. Size should only refer.to tensor.dim.

kylebgorman:
> @kylebgorman revert fly by comment to width. Size should only refer.to tensor.dim.

done

PR #187, opened 2024-05-07
Puts val loss on cpu for np.mean()

Adamits:
I guess I had not tested the final evaluators update on gpu. Since we now get the mean val loss with numpy.mean, we need to be sure the loss tensors are on cpu.

### DIFF
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -304,7 +304,7 @@ class BaseEncoderDecoder(pl.LightningModule):
             greedy_predictions, 2, 0, target_padded.size(1)
         )
         loss = self.loss_func(greedy_predictions, target_padded)
-        val_eval_items_dict.update({"val_loss": loss})
+        val_eval_items_dict.update({"val_loss": loss.cpu()})
         return val_eval_items_dict
 
     def validation_epoch_end(self, validation_step_outputs: Dict) -> Dict:


kylebgorman:
What if we just use PyTorch's mean? Does that solve the problem? 

Two interfaces for it:

https://pytorch.org/docs/stable/generated/torch.mean.html
https://pytorch.org/docs/stable/generated/torch.Tensor.mean.html

Adamits:
> What if we just use PyTorch's mean? Does that solve the problem?
> 
> Two interfaces for it:
> 
> https://pytorch.org/docs/stable/generated/torch.mean.html https://pytorch.org/docs/stable/generated/torch.Tensor.mean.html

Good point, thats a better solution.

Adamits:
Actually, the issue is that we end up with a list, not a tensor, and I think the overhead of turning the list into a tensor first is probably > the overhead of moving the values to CPU. I can test though.

kylebgorman:
> Actually, the issue is that we end up with a list, not a tensor, and I think the overhead of turning the list into a tensor first is probably > the overhead of moving the values to CPU. I can test though.

yeah whatever's faster (or if they're the same pick the one that's more elegant...)

kylebgorman (approved):
LGTM.

kylebgorman:
Shall I merge?

Adamits:
Yes!

PR #195, opened 2024-06-09
Improves docs throughout the models and modules

kylebgorman:
Most of this is just a matter of spacing and punctuation. However:

* I resolved the disagreement between new flag names and internal names. I don't care what exactly they're called and can go the other way if you like, but the names really ought to be consistent as it was hard for me to track otherwise.
* I fixed a few apparent typing errors.
* The TODOs I added are ones where I need help from you (@adamits or @bonham79) to figure out what the types ought to be. I'll resolve those before submitting.
* I made use of constants in a few places where they weren't being used after #186.


yoyodyne/models/hard_attention.py
```diff
@@ -279,8 +285,12 @@ def greedy_decode(
             )
         return predictions, likelihood
 
+    # FIXME(kbg,bonham79): Needs input types and docs.
```
bonham79, line 288: Which is fixme, above or below?

yoyodyne/models/modules/lstm.py
```diff
@@ -226,25 +232,29 @@ def name(self) -> str:
...
         self.output_proj = nn.Sequential(
             nn.Linear(self.output_size, self.output_size), nn.Tanh()
         )
-        # Project transition probabilities to depth of module.
+        # Projects transition probabilities to depth of module.
         self.scale_encoded = nn.Linear(
             self.decoder_input_size, self.hidden_size
         )
 
+    # TODO: needs docs.
```
bonham79, line 248: Note for myself.

yoyodyne/models/modules/transformer.py
```diff
@@ -336,21 +336,21 @@ def forward(
                 encoder.
             features_memory (torch.Tensor): the mask for the features.
             target_mask (Optional[torch.Tensor], optional): the mask for the
-                target sequence. Defaults to None.
+                target sequence.
```
bonham79, line 339: Should include defaults in docstrings.

bonham79 (changes requested):

yoyodyne/defaults.py
```diff
@@ -28,8 +28,8 @@
 DECODER_LAYERS = 1
 EMBEDDING_SIZE = 128
 ENCODER_LAYERS = 1
-HARD_ATTENTION_CONTEXT = 0
-HARD_ATTENTION_MONOTONIC = False
+ATTENTION_CONTEXT = 0
```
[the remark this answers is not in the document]
kylebgorman: This is already documented in the help string, it doesn't even have an obvious interpretation in the soft attention case, and it's a standard we hold no other default or flag to (the flag for the number of attention heads doesn't work with LSTMs for instance), and it's the only remaining case where a default and flag have a different name than what they're called in the model class.

yoyodyne/models/hard_attention.py
```diff
@@ -56,15 +60,25 @@ def __init__(
...
-        Simply feeds a BOS string to decoder to provide initial probability.
+        This feeds the BOS string to the decoder to provide an initial
+        probability.
+
+        Args:
+            encoder_out (torch.Tensor).
+            encoder_mask (torch.Tensor).
+
+        Returns:
+            Tuple[torch.Tensor, torch.Tensor, Tuple[torch.Tensor,
```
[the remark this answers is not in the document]
kylebgorman: Not a bad idea---I think @Adamits did this for module outputs---but if it's okay with you I'd like to put that in a different PR. Implementationally it may not be as simple as just making a named tuple if we need to keep it connected with the compute graph. (See [here](https://github.com/CUNY-CL/yoyodyne/blob/a96100f49f6c619329019a8f4083e68040d63a1a/yoyodyne/data/batches.py#L49) for an example of what one does if that's the case.)

yoyodyne/models/hard_attention.py
kylebgorman: Below. Can you provide the types for that? And maybe some text for describing inputs and outputs?

yoyodyne/models/modules/transformer.py
kylebgorman: That's already documented on the previous line, that's what "optional" means in this context.

yoyodyne/models/hard_attention.py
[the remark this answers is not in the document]
bonham79: Creating minor issue to tag to PR.

yoyodyne/models/modules/lstm.py
```diff
@@ -105,6 +107,8 @@ def name(self) -> str:
 
 
 class LSTMDecoder(LSTMModule):
+    """LSTM encoder."""
```
Adamits, line 110: encoder --> decoder

yoyodyne/models/modules/transformer.py
```diff
@@ -336,21 +336,21 @@ def forward(
...
             memory_mask (Optional[torch.Tensor], optional): the mask for the
-                memory sequence. Defaults to None.
+                memory sequence.
             features_memory_mask (torch.Tensor, optional): the mask
```
Adamits, line 342: Should this be `Optional[torch.Tensor]` ?

yoyodyne/models/lstm.py
```diff
@@ -149,6 +149,8 @@ def decode(
         predictions = torch.stack(predictions)
         return predictions
 
+    # TODO(adamits,kbg): out-of-date documentation.
```
Adamits, line 152: Is this for the README or the method below? Or perhaps just a general note for me to check all of it out?

Adamits (approved):
Two minor nits and a question. Otherwise looks great!

yoyodyne/models/modules/transformer.py
kylebgorman: Fixed

yoyodyne/models/lstm.py
kylebgorman: I would like help explaining the return type, which is currently just labeled "_description_". I *think* the union is because you can request confidences in addition to just the prediction tensor but I'm not sure. Could you confirm and then I can 1) fill this in and 2) remove the TODO?

I also want to get @bonham79's suggestion on that one undocumented static function into this PR.

Adamits: > I would like help from you in explaining the return type, which is currently just labeled "_description_". I _think_ the union is because you can request confidences in addition to just the prediction tensor but I'm not sure. Could you confirm and then I can 1) fill this in and 2) remove the TODO?
> 
> I also want to get @bonham79's suggestion on that one undocumented static function into this PR.

Ah I see. I think it should be a List of ints (the predictions) and a optionally a List of floats (the likelihoods). I can rewrite this so its easier to follow, but I think I have a long-standing TODO to implement this to handle batches for a bunch of the architectures. Right the Union was because you can optionally `return_confidences`, but I don't think thats the right usage of that type hint given we can just mark an optional return.

I also think we should just remove `return_confidences` and always return them. It is easy to `preds, _ = model.beam_decode(...)`

kylebgorman: Okay if I just add a TODO to the effect "Modify to eliminate polymorphic return type; always return confidences." ?

Adamits: > Okay if I just add a TODO to the effect "Modify to eliminate polymorphic return type; always return confidences." ?

Yeah sure

yoyodyne/models/hard_attention.py
bonham79: I'm not a fan of pushing to other's repos, so this would be the doc string:

```
    @staticmethod
    def greedy_step(
        log_probs: torch.Tensor, likelihood: torch.Tensor
    ) -> Tuple[torch.Tensor, torch.Tensor]:
        """Greedy decoding of current timestep.

        Args:
            log_probs (torch.Tensor): vocabulary probabilities at current time
                step.
            likelihood (torch.Tensor): accumulative likelihood of decoded character
                sequence.

        Returns:
            Tuple[torch.Tensor, torch.Tensors]: Greedy decoded character for current
                time step and current likelihood of decoded character sequence. 
        """
        tgt_prob = likelihood + log_probs.transpose(1, 2)
        tgt_prob = tgt_prob.logsumexp(dim=-1)
        tgt_char = torch.max(tgt_prob, dim=-1)[1]
        return tgt_char.unsqueeze(-1), likelihood
```
Also realize this isn't necessary as a tuple. Will remove with qol update.

yoyodyne/models/modules/lstm.py
bonham79: def _alignment_step(
        self,
        decoded: torch.Tensor,
        encoder_out: torch.Tensor,
        encoder_mask: torch.Tensor,
    ) -> torch.Tensor:
        """Creates alignment matrix for current time-step.

        Given current encoder representation and the decoder
            representations at current time step, calculates alignment
            scores between all potential source sequence pairings.
            These alignments are used to predict likelihood of state
            transition for current decoder output.
            
             See: https://aclanthology.org/P19-1148/.

        Args:
            decoded (torch.Tensor): Output from decoder for
                current time step of shape B x 1 x decoder_dim
            encoder_out (torch.Tensor): encoded input sequence of shape
                B x seq_len x encoder_dim.
            encoder_mask (torch.Tensor): mask for the encoded input batch of
                shape B x seq_len.
        
                
        Returns:
            torch.Tensor of alignment scores across source sequence
                of shape B x seq_len x seq_len.
        """

        alignment_scores = torch.bmm(
            self.scale_encoded(encoder_out), decoded.transpose(1, 2)
        ).squeeze(-1)
        # Gets probability of alignments.
        alignment_probs = nn.functional.softmax(alignment_scores, dim=-1)
        # Mask padding.
        alignment_probs = alignment_probs * (~encoder_mask) + 1e-7
        alignment_probs = alignment_probs / alignment_probs.sum(
            dim=-1, keepdim=True
        )
        # Expands over all time steps. Log probs for quicker computation.
        return (
            alignment_probs.log()
            .unsqueeze(1)
            .expand(-1, encoder_out.shape[1], -1)
        )

yoyodyne/models/hard_attention.py
kylebgorman: That's true, you're not changing the likelihood.

Issue #192, opened 2024-06-08
TQDM Error with multi GPU Transducer


bonham79:
Issue when running multi-gpu training with edit action transducer: 
```
Traceback (most recent call last):                                                                                                   
  File "/home/salamander/anaconda3/envs/sigmorphon2024/bin/yoyodyne-train", line 8, in <module>
    sys.exit(main())
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/yoyodyne/train.py", line 390, in main
    model = get_model_from_argparse_args(args, datamodule)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/yoyodyne/train.py", line 214, in get_model_from_argparse_args
    return model_cls(
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/yoyodyne/models/transducer.py", line 43, in __init__
    super().__init__(*args, **kwargs)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/yoyodyne/models/lstm.py", line 36, in __init__
    super().__init__(*args, **kwargs)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/yoyodyne/models/base.py", line 155, in __init__
    self.save_hyperparameters(
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/pytorch_lightning/core/mixins/hparams_mixin.py", line 110, in save_hyperparameters
    save_hyperparameters(self, *args, ignore=ignore, frame=frame)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/site-packages/pytorch_lightning/utilities/parsing.py", line 275, in save_hyperparameters
    obj._hparams_initial = copy.deepcopy(obj._hparams)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/copy.py", line 297, in _reconstruct
    value = deepcopy(value, memo)
  File "/home/salamander/anaconda3/envs/sigmorphon2024/lib/python3.10/copy.py", line 172, in deepcopy
    y = _reconstruct(x, memo, *rv)
[...]

PR #197, opened 2024-06-10
adding expert to hp ignore due to pickeling issue

bonham79:
Fixes: https://github.com/CUNY-CL/yoyodyne/issues/192

TQDM in the transducer's expert module doesn't pickle currently. So when doing multi-gpu training, pytorch gets annoyed. Since it's not really needed beyond training, I'm dropping it from the model.

Note, when running prediction with the model, the lack of the hyperparameter means a new expert will have to be created. This is dumb, so need to add a dummy class to the transducer model to avoid this. This portion I'm leaving in draft atm.

### DIFF
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -153,7 +153,7 @@ class BaseEncoderDecoder(pl.LightningModule):
         self.decoder = self.get_decoder()
         # Saves hyperparameters for PL checkpointing.
         self.save_hyperparameters(
-            ignore=["source_encoder", "decoder", "features_encoder"]
+            ignore=["source_encoder", "decoder", "expert", "features_encoder"]
         )
         # Logs the module names.
         util.log_info(f"Model: {self.name}")


Adamits:
Multiprocess + pickle has given me so many issues recently across interdependent libraries...

One edge case: how about continued training from a checkpoint? Then we need the expert right, which I assume has some relevant state?

kylebgorman:
@Adamits I think this would pose exactly the same issue as the multi-GPU case.

Adamits:
@kylebgorman Right, I meant an edge case where this solution would break something. If I understand correctly, Travis wants to just not save the expert state to avoid the pickle issue with TQDM. But if we want to continue form a checkpoint, i think we need the expert state.

tbartley94:
> One edge case: how about continued training from a checkpoint? Then we need the expert right, which I assume has some relevant state?

you'd have to train the expert again. thankfully your np fix makes it less annoying. but yeah, that's why i was curious if there were better ideas. 

fwiw, the entire point behind the expert transducer is for single gpu setup. (Actually it works better on CPU for most cases due to the expert queries). You don't get really parallelism speed ups since there's so many gpu-> cpu transfers.

kylebgorman:
Yeah I also find the transducer is better on CPU. (The same was true with the original code in DyNet.)

bonham79:
> Yeah I also find the transducer is better on CPU. (The same was true with the original code in DyNet.)

Yeah it's the main motivation for its use so not surprised.

kylebgorman:
Is this made redundant by #247? @bonham79 @Adamits

bonham79:
@kylebgorman not just redundant, unnecessary and detrimental

PR #202, opened 2024-06-16
Pin NumPy < 2.0.0

kylebgorman:
NumPy 2 is the first major release in almost 2 decades and has some breaking changes. If we run with NumPy 2 we see the following test failure:

```
AttributeError: `np.float_` was removed in the NumPy 2.0 release. Use `np.float64` instead.. Did you mean: 'float16'?
```

This PR pins the versions to NumPy < 2.0.0 until PyTorch, W&B, etc. support NumPy 2. A TODO is added. In a minor drive-by, I also update the effected test to take the hard monotonic models into account.


bonham79 (approved):
LGTM

PR #204, opened 2024-07-04
Batch autosizing

### LINKED ISSUE #148: Automatic batch sizing
PyTorch Lightning supports a mode where it automatically computes the maximum batch size on your accelerator (by collecting gradients over a few batches and then using binary search to find a max that doesn't give OOM errors). Documentation is [here](https://lightning.ai/docs/pytorch/stable/advanced/training_tricks.html#batch-size-finder). It appears straightforward to implement (see "Note" on docs.) 

Having this enabled would be very useful for computing the maximum batch size possible. From this one could factor the desired batch size and accumulate gradients across multiple batches per optimizer step, so this is related to #132 in the obvious way.

This will also combine nicely with a move to LightningCLI, IMO, since that uses a subcommand interface and we could do something like `yoyodyne scale_batch_size --arg1 --arg2` to enable this.

kylebgorman:
Closes #148. (Sorry to jump ahead---I think you were working on this @Adamits but it's become a major blocker for various other things I'm trying to do with my students.)

If requested with `--find_batch_size`, this finds the max batch size, then computes the "optimal" batch size to use such that it, plus multiple gradient accumulation steps, is as large as possible but does not exceed the max. The major features of the sizer are also exposed as flags though I think the defaults are what most people will want. (The `binsearch` method gives more precise results but it will probably not find more optimal numbers since they won't be perfectly divisible. However this is what you want if you just want to find the absolute max and train with it.) 

The values are all logged and set, so that one can combine this with training, or one can simply do this without training at all using a trick like `--max_epochs 0` as documented in the README.

I tested this by taking a large transformer model (many encoder and decoder layers), requesting an enormous batch size of 2048. In this case (and following my GPU memory utilization bears this out) it succeeds at batches of size 1024 but fails at batches of 2048, so it sets batch size = 1024 and uses two gradient accumulation steps per update---it proceeds to update, so this would be compatible with a hyperparameter search. I also repeated this with `--max_epochs 0`, in which case it simply logs the same results from the finder, and then halts.

Other changes:

* I have undone the hack whereby I double batch size when not training, since that seems to conflict with this.
* I have removed the initial LR finder since that also seems to conflict. 

Sample output #1 (no OOMs at risk):

```
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 2 succeeded, trying batch size 4
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 4 succeeded, trying batch size 8
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 8 succeeded, trying batch size 16
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 16 succeeded, trying batch size 32
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 32 succeeded, trying batch size 64
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 64 succeeded, trying batch size 128
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 128 succeeded, trying batch size 256
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 256 succeeded, trying batch size 512
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 512 succeeded, trying batch size 1024
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 1024 succeeded, trying batch size 2048
`Trainer.fit` stopped: `max_steps=3` reached.
Batch size 2048 succeeded, trying batch size 4096
Batch size 4096 failed, trying batch size 2048
Finished batch size finder, will continue with full run using batch size 2048
Restoring states from the checkpoint path at ../models/transformer/.scale_batch_size_84895d35-c4e3-4f2d-82c9-31945a024da8.ckpt
Restored all states from the checkpoint file at ../models/transformer/.scale_batch_size_84895d35-c4e3-4f2d-82c9-31945a024da8.ckpt
Max batch size: 2048
Using batch size: 1024
Using gradient accumulation steps: 1
```

Sample output #2 (avoiding an OOM with gradient accumulation):

```
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 2 succeeded, trying batch size 4
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 4 succeeded, trying batch size 8
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 8 succeeded, trying batch size 16
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 16 succeeded, trying batch size 32
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 32 succeeded, trying batch size 64
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 64 succeeded, trying batch size 128
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 128 succeeded, trying batch size 256
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 256 succeeded, trying batch size 512
`Trainer.fit` stopped: `max_steps=10` reached.
Batch size 512 succeeded, trying batch size 1024
Batch size 1024 failed, trying batch size 512
Finished batch size finder, will continue with full run using batch size 512
Restoring states from the checkpoint path at ../models/transformer/.scale_batch_size_22730f17-efd1-44f8-b5f0-bc4efda67e89.ckpt
Restored all states from the checkpoint file at ../models/transformer/.scale_batch_size_22730f17-efd1-44f8-b5f0-bc4efda67e89.ckpt
Max batch size: 512
Using batch size: 512
Using gradient accumulation steps: 4
```


Adamits:
Awesome!

This had become the thing that I put in my calendar to do after work like 5
different times and never got around to, so I appreciate you doing it!

I will give it a review tomorrow

Adam

README.md
```diff
@@ -1,5 +1,4 @@
-Yoyodyne 🪀
-==========
+# Yoyodyne 🪀
```
Adamits, line 1: Are these formatting changes intentional? Have I or someone else been reverting them and you keep having to change them back? I feel like I often see your PRs update the README markdown format in this same way.

yoyodyne/sizing.py
```diff
@@ -0,0 +1,96 @@
...
+from pytorch_lightning.tuner import tuning
+
+from . import data, defaults, models
+
+
+def max_batch_size(
+    trainer: pl.Trainer,
+    model: models.BaseEncoderDecoder,
+    datamodule: data.DataModule,
+    *,
```
Adamits, line 16: I think we have been following the `*args` convention. Unless this is doing something else that I do not understand...

yoyodyne/sizing.py
```diff
@@ -0,0 +1,96 @@
...
+    mode: str = defaults.AUTO_BATCH_SIZE_MODE,
+    steps_per_trial: int = defaults.AUTO_BATCH_SIZE_STEPS_PER_TRIAL,
+    max_trials: int = defaults.AUTO_BATCH_SIZE_MAX_TRIALS,
+) -> int:
+    """Computes the maximum batch size that will fit in memory."""
```
Adamits, line 21: Args/Returns spec?

yoyodyne/sizing.py
```diff
@@ -0,0 +1,96 @@
...
+    )
+
+
+def optimal_batch_size(
+    desired_batch_size: int, max_batch_size: int
+) -> Tuple[int, int]:
+    r"""Computes optimal batch size and number of gradient accumulation steps.
+
+    Given the desired batch size $b$ and a max batch size $n_{\textrm{max}}$,
+    this returns a tuple $a \in \mathcal{N}_{+}, n \in \mathcal{N}_{\le n}$
```
Adamits, line 38: Should `n \in \mathcal{N}_{\le n}` --> `n \in \mathcal{N}_{\le n_\textrm{max}}`?

yoyodyne/sizing.py
```diff
@@ -0,0 +1,96 @@
...
+    """Adds batch sizing configuration options to the argument parser.
+
+    Args:
+        parser (argparse.ArgumentParser).
+    """
+    parser.add_argument(
+        "--auto_batch_size_find",
+        action="store_true",
+        default=defaults.AUTO_BATCH_SIZE_FIND,
+        help="Automatically find the maximum batch size? "
```
Adamits, line 68: I think the question mark should be a period. Or is this a convention for boolean flags?

yoyodyne/defaults.py
```diff
@@ -35,6 +35,12 @@
 MAX_TARGET_LENGTH = 128
 EVAL_METRICS = set()
 
+# Tuning arguments.
+AUTO_BATCH_SIZE_FIND = False
+AUTO_BATCH_SIZE_MODE = "power"
+AUTO_BATCH_SIZE_STEPS_PER_TRIAL = 3
```
Adamits, line 41: Is there justification for 3 rather than 1? Furthermore for any number < number of steps per epoch, we probably can't ensure you won't run OOM right?

Adamits (changes requested):
This looks clean! I left a few nits.

One high-level thought though is:

Suppose I don't care what effective batch size I train with, I strictly want the maximum possible batch size with no gradient accumulation. Can I do that in one go? Or do I need to run this and log the max_batch size, kill the job, then rerun setting that as `--batch_size`?

yoyodyne/defaults.py
kylebgorman: This is just shadowing the default values from the lower-level PTL library. 

The `"power"` method then just starts with 2^1, 2^2, etc. until it hits an OOM or batch size exceeds the training set size. If you use `"binsearch"` it starts like that using the OOM size as an upper bound. The extra work in the binary search method is usually not worth it if you're going to use gradient accumulation since the extra work it does will move it away from a nice power-of-two value with good divisors. 

Yes, this is heuristic when less than the number of batches per epoch. If you get an OOM anyways, try doubling it. (I actually hit that case once; doubling fixed it.)

README.md
kylebgorman: I use Pandoc (from markdown to markdown) to reflow the README when I make major changes to the README.

kylebgorman:
> Suppose I don't care what effective batch size I train with, I strictly want the maximum possible batch size with no gradient accumulation. Can I do that in one go? 

You can in the common case where the requested `--batch_size` is smaller than the max size it finds. Then it'll use the requested batch size and no gradient accumulation (it'll set that flag to `1`).

yoyodyne/defaults.py
Adamits: Ahhh ok makes sense!

Adamits:
> > Suppose I don't care what effective batch size I train with, I strictly want the maximum possible batch size with no gradient accumulation. Can I do that in one go?
> 
> You can in the common case where the requested `--batch_size` is smaller than the max size it finds. Then it'll use the requested batch size and no gradient accumulation (it'll set that flag to `1`).

What I meant is: can I just request it to find max_batch_size and then train with that whatever it is. But I think just calling this for 0 epochs, and then requestin w/e max it found is fine so I don't think it'sworth worrying about.

kylebgorman:
> What I meant is: can I just request it to find max_batch_size and then train with that whatever it is. But I think just calling this for 0 epochs, and then requestin w/e max it found is fine so I don't think it'sworth worrying about.

Oh, that's not supported but I can imagine how you'd support that (I think you'd have a convention like when the batch size is non-positive and finding is requested, we just use max batch size).

yoyodyne/sizing.py
kylebgorman: This is unrelated to what you're mentioning. It is a convention in the language for requiring certain args to be passed as kwargs, and we use it elsewhere in the library.

yoyodyne/sizing.py
kylebgorman: I actually do use this convention but not in this library so I'll fix as you request.

yoyodyne/sizing.py
kylebgorman: +1

kylebgorman:
> Oh, that's not supported but I can imagine how you'd support that (I think you'd have a convention like when the batch size is non-positive and finding is requested, we just use max batch size).

Added and documented this. The trick I am using is you pass `--batch_size 0 --auto_batch_size_find` and it'll just use the max batch size it finds. (In that case I'd recommend also using `--auto_batch_size_mode binsearch` since it gets you slightly higher saturation of the accelerator.)

kylebgorman:
> Added and documented this. The trick I am using is you pass `--batch_size 0 --auto_batch_size_find` and it'll just use the max batch size it finds. (In that case I'd recommend also using `--auto_batch_size_mode binsearch` since it gets you slightly higher saturation of the accelerator.)

Nevermind, this breaks something else.

Issue #194, opened 2024-06-09
Index error with hard monotonic attention and features


kylebgorman:
(It is always possible that I am just doing something wrong here.)

Invocation:

```
yoyodyne-train \
    --experiment "${LANGUAGE}" \
    --train "${TRAIN}" \
    --val "${VAL}" \
    --model_dir "${MODEL_DIR}" \
    --features_col 3 \
    --arch hard_attention_lstm \
    --enforce_monotonic \
    --hard_attention_context 1 \
    --hidden_size 400 \
    --batch_size 128 \
    --embedding_size 200 \
    --accelerator gpu \
    --max_epochs 60 \
    --log_every_n_step 20 \
    --learning_rate .001 \
    --dropout .4 \
    --seed 49 \
    --gradient_clip_val 3
```

The trace below is the CPU one because it's more informative---you get a less informative error from CUDA if you're on GPU.

```
  File "/home/kbg/.miniconda3/envs/py310/bin/yoyodyne-train", line 8, in <module>
    sys.exit(main())
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/yoyodyne/train.py", line 406, in main
    best_checkpoint = train(trainer, model, datamodule, args.train_from)
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/yoyodyne/train.py", line 267, in train
    trainer.fit(model, datamodule, ckpt_path=train_from)
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 608, in fit
    call._call_and_handle_interrupt(
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/pytorch_lightning/trainer/call.py", line 38, in _call_and_handle_interrupt
    return trainer_fn(*args, **kwargs)
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 650, in _fit_impl
    self._run(model, ckpt_path=self.ckpt_path)
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 1112, in _run
    results = self._run_stage()
  File "/home/kbg/.miniconda3/envs/py310/lib/python3.10/site-packages/pytorch_lightning/trainer/trainer.py", line 1191, in _run_stage
[...]

PR #206, opened 2024-07-05
Autosizing modes

kylebgorman:
In this follow-up to #194, I make it so that the batch size finder has two modes:

* `opt`imal mode is the same as before: it uses the gradient accumulation trick if necessary
* `max` mode just finds the max batch size and uses it, ignoring `--batch_size`

I theorize the latter would be really useful for when you have reason to believe the model does well with large batches (this might be all models?) and rather than tuning it, want to tune the other hyperparameters instead.

I am testing this with a fancy Bayesian sweep [here](https://wandb.ai/hyperparam-sensitivity/pol-max_batch-hard_monotonic_lstm?nw=nwuserkylebgorman) (the link may not be accessible to just anyone).


Adamits (approved):
LGTM

PR #209, opened 2024-07-10
Implements #208

### LINKED ISSUE #208: Quicker automatic batch sizing in `opt` mode
I propose to hide some of the complexity of the automatic batch sizing library to work faster with `--find_batch_size opt`.

If one is using the `opt` strategy, chances are they are simply using this inside of a sweep and trying to forestall OOMs when both model size and batch size are substantial. So one likely outcome is that the max batch size found by the automatic system (greatly, perhaps) exceeds the desired `--batch_size`. However, there is no easy way to tell the batch scaling library to stop at a certain value. Without this you spend a lot of time searching for big batch sizes (and each trial takes twice as long...or worse...than the previous one) knowing that you don't need this. 

However, there is an indirect way to do so. What I propose is that in the `opt` strategy, we set the `max_trials` argument [here](https://github.com/CUNY-CL/yoyodyne/blob/3de37d9dd0e5bc52e527098004975951b586f4a5/yoyodyne/sizing.py#L44) to terminate search once it exceeds the desired batch size. That is, I'd say `max_trials = int(log2(desired_batch_size - 1))`. **Note that this seems like it would be off by 1, but the way they wrote the search strategy seems, to me, to be off by 1 in the opposite direction so it cancels out, lol.**

I would add the two new `max_*` strategies to the `--find_batch_size` flag, and I would remove the `--find_batch_size_mode` flag, which is no longer useful (you always want "power"). Continuining the flag changes, I would eliminate `--find_batch_max_trials` since it would be set automatically with the `opt` strategy, and we can just use the default 25 for the `max` case, which is more than enough.

kylebgorman:
Closes #208. See #208 for context.

### DIFF
--- a/yoyodyne/sizing.py
+++ b/yoyodyne/sizing.py
@@ -61,8 +61,8 @@ def find_batch_size(
     This sets the batch size on the datamodule, and if necessary, the
     gradient accumulation steps on the trainer.
 
-    With the max method, binary search is used to find the largest batch size
-    that fits in memory.
+    With the max method, batch size is doubled until a memory error, thus
+    finding the largest batch size that fits in memory.
 
     With the optimal method, batch size is doubled until a memory error or
     once we confirm that the desired batch size fits in memory (whichever
@@ -83,7 +83,6 @@ def find_batch_size(
         max_batch_size = tuner.scale_batch_size(
             model,
             datamodule=datamodule,
-            mode="binsearch",
             steps_per_trial=steps_per_trial,
         )
         # This automatically overwrites the datamodule batch size argument.
@@ -102,7 +101,6 @@ def find_batch_size(
             # by one" but the way it's defined in the batch size finder is
             # itself "off by one" in the opposite direction, so it cancels out.
             max_trials=int(numpy.log2(desired_batch_size - 1)),
-            mode="power",
             steps_per_trial=steps_per_trial,
         )
         steps, batch_size = _optimal_batch_size(
--- a/yoyodyne/sizing.py
+++ b/yoyodyne/sizing.py
@@ -61,11 +61,12 @@ def find_batch_size(
     This sets the batch size on the datamodule, and if necessary, the
     gradient accumulation steps on the trainer.
 
-    With the max strategy, the "binsearch" method is used, and gives the
-    largest batch size that fits in memory.
+    With the max method, binary search is used to find the largest batch size
+    that fits in memory.
 
-    With the optimal strategy, the "power" strategy" is used, and the search
-    halts if the desired batch size fits in memory.
+    With the optimal method, batch size is doubled until a memory error or
+    once we confirm that the desired batch size fits in memory (whichever
+    comes first).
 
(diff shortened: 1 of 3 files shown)


yoyodyne/sizing.py
```diff
@@ -87,53 +54,68 @@ def find_batch_size(
...
     Args:
         method (str): one of "max" (find the maximum batch size) or "opt" (find
             the "optimal" batch size, using the gradient accumulation trick if
             necessary.)
         trainer (pl.Trainer).
         model (models.BaseEncoderDecoder).
         datamodule (data.DataModule).
-        max_trials (int, optional): maximum number of increases in batch size
-            before terminating.
-        mode (str, optional): one of "binsearch" or "power".
```
Adamits, line 108: Why are we removing this?

yoyodyne/sizing.py
```diff
@@ -87,53 +54,68 @@ def find_batch_size(
...
+    comes first).
+
     Args:
         method (str): one of "max" (find the maximum batch size) or "opt" (find
             the "optimal" batch size, using the gradient accumulation trick if
             necessary.)
         trainer (pl.Trainer).
         model (models.BaseEncoderDecoder).
         datamodule (data.DataModule).
-        max_trials (int, optional): maximum number of increases in batch size
```
Adamits, line 106: Shouldn't we still support this? I thought your change enables a specific use-case, not that it fully replaces what you had before

Adamits (changes requested):
Great catch on this. I have a couple in-line questions about the options you are removing.

yoyodyne/sizing.py
kylebgorman: So I have tested this relativley extensively and you absolutely do not want to use binsearch. It takes orders of magnitude more time and it then immediately crashes with an OOM anyways because our batch lengths vary.

yoyodyne/sizing.py
kylebgorman: This is meaningless in `opt` mode and in `max` mode the default is fine (had I designed this I probably would have made this run forever but it is going to hit an OOM or get bigger than the training set long before 2^25...)

Adamits: Ok, so we basically think it is not a useful feature and therefore don't expose it.

Adamits (approved):
Thanks for the responses. LGTM.

yoyodyne/sizing.py
kylebgorman: Yeah, I thought it would be useful, but there are two factors:

1. The time to run trials is linaer (or maybe a little worse than linear, IMO) with the size of the batches, so binsearch spends ln(n) time with the biggest, slowest batches.
2. When it settles on a max size it is going to have gotten lucky with having drawn three smaller batches.

I am still exposing the number of batches to draw per trial so if you want something more precise you can do it...

Part of me thinks I should just write my own scaler from scratch---it's not terribly complex as can be seen from the PTL impl.

PR #210, opened 2024-07-11
Adds a script to retrieve hyperparameters

kylebgorman:
This script usees the W&B API to find the best run and then retrieve its hyperparameters. Some hacks are used to suppress redundant hyperparameters (or things that are not really hyperparameters) and to print them out as flags so that this could be used like:

        yoyodyne-train ... $(./best_hyperparameters ...)


kylebgorman:
Gentle ping @Adamits

Adamits (approved):
Cool idea! LGTM

PR #211, opened 2024-07-11
Cleanup of some flag help strings

kylebgorman:
* If something has a null default, don't list a default at all ("not enabled" isn't always precise enough).
* For a boolean flag (e.g., --foo/--no_foo) defaulting to true, give its default as "enabled".
* For a boolean flag defaulting to false, give its default as "not enabled".
* Mention that transducer architectures go wtih tied embeddings.
* Mention that patience_metric goes with patience.

### DIFF
--- a/yoyodyne/evaluators.py
+++ b/yoyodyne/evaluators.py
@@ -331,5 +331,5 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         action=util.UniqueAddAction,
         choices=_eval_factory.keys(),
         default=defaults.EVAL_METRICS,
-        help="Additional metrics to compute. Default: none.",
+        help="Additional metrics to compute.",
     )
--- a/yoyodyne/models/lstm.py
+++ b/yoyodyne/models/lstm.py
@@ -342,8 +342,8 @@ class LSTMEncoderDecoder(base.BaseEncoderDecoder):
             "--bidirectional",
             action="store_true",
             default=defaults.BIDIRECTIONAL,
-            help="Uses a bidirectional encoder "
-            "(LSTM-backed architectures only. Default: %(default)s.",
+            help="Uses a bidirectional encoder (LSTM-backed architectures "
+            "only. Default: enabled.",
         )
         parser.add_argument(
             "--no_bidirectional",
--- a/yoyodyne/train.py
+++ b/yoyodyne/train.py
@@ -317,7 +317,8 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         help="Path to input validation data TSV.",
     )
     parser.add_argument(
-        "--train_from", help="Path to ckpt checkpoint to resume training from."
+        "--train_from",
+        help="Path to checkpoint used to resume training.",
     )
     # Other training arguments.
     parser.add_argument(
@@ -346,8 +347,8 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         choices=["accuracy", "loss", "ser"],
         default=defaults.PATIENCE_METRIC,
         help="Stops early when validation `accuracy` does not increase or "
-        "when validation `loss` or `ser` does not decrease for `--patience` "
-        "epochs. Default: %(default)s.",
+        "when validation `loss` or `ser` does not decrease. "
+        "Default: %(default)s.",
     )
     parser.add_argument("--seed", type=int, help="Random seed.")
     parser.add_argument(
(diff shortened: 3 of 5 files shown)


kylebgorman:
Gentle ping @Adamits

yoyodyne/evaluators.py
```diff
@@ -331,5 +331,5 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         action=util.UniqueAddAction,
         choices=_eval_factory.keys(),
         default=defaults.EVAL_METRICS,
-        help="Additional metrics to compute. Default: none.",
+        help="Additional metrics to compute.",
```
Adamits, line 334: To be sure, does this always add to the default, or if we set only one eval_metric, does it override the default. The help language makes it sound like the former

yoyodyne/models/lstm.py
```diff
@@ -342,8 +342,8 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             "--bidirectional",
             action="store_true",
             default=defaults.BIDIRECTIONAL,
-            help="Uses a bidirectional encoder "
-            "(LSTM-backed architectures only. Default: %(default)s.",
+            help="Uses a bidirectional encoder (LSTM-backed architectures "
+            "only. Default: enabled.",
```
Adamits, line 346: So if one was to change `defaults.BIDIRECTIONAL,` then one would need to remember to update this text. Feels slightly dangerous. Then again, parsing defaults feels maybe a bit excessive?

yoyodyne/train.py
```diff
@@ -317,9 +317,7 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         help="Path to input validation data TSV.",
     )
     parser.add_argument(
-        "--train_from",
-        help="Path to ckpt checkpoint to resume training from. "
-        "Default: not enabled.",
+        "--train_from", help="Path to ckpt checkpoint to resume training from."
```
Adamits, line 320: Maybe remove "ckpt"?

yoyodyne/train.py
```diff
@@ -341,23 +339,23 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
...
     parser.add_argument(
         "--patience_metric",
         choices=["accuracy", "loss", "ser"],
         default=defaults.PATIENCE_METRIC,
-        help="Stops early when validation `accuracy` stops increasing or "
-        "when validation `loss` or `ser` stops decreasing. "
-        "Default: %(default)s.",
+        help="Stops early when validation `accuracy` does not increase or "
+        "when validation `loss` or `ser` does not decrease for `--patience` "
+        "epochs. Default: %(default)s.",
```
Adamits, line 350: I am confused about the final "s" in the string here.

yoyodyne/train.py
```diff
@@ -341,23 +339,23 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
...
     )
     parser.add_argument("--seed", type=int, help="Random seed.")
     parser.add_argument(
         "--log_wandb",
         action="store_true",
         default=defaults.LOG_WANDB,
-        help="Use Weights & Biases logging (log-in required). Default: True.",
+        help="Use Weights & Biases logging (log-in required). "
+        "Default: not enabled.",
```
Adamits, line 358: Same Q as before about if defaults change

Adamits (changes requested):

yoyodyne/evaluators.py
kylebgorman: It's the former. The "default metric" as you're calling it is set by `--checkpoint_metric` in my design and can actually include all three metrics we support (loss, accuracy, SER).

yoyodyne/models/lstm.py
kylebgorman: That's true but I find the additional information (that this is the default value of a boolean flag) useful enough to justify the small potential for debt. 

There is a tradition I learned about at Google of writing programmatic comments which express the notion "if you change this you have to change that"; it's then enforced via some kind of software nudge (looks like your PR changes this, are you sure you don't want to change that?). But I don't know if anything like that exists outside of that context.

yoyodyne/train.py
kylebgorman: Patience is a number, so after 5 (say) epochs...

yoyodyne/train.py
kylebgorman: Done.

kylebgorman: Also rewrote slightly to avoid stigmatized particle stranding.

yoyodyne/train.py
kylebgorman: I removed it though, I decided it is awkward as you say.

kylebgorman:
@Adamits I made some minor changes here; is this ready to go?

Adamits (approved):
LGTM

PR #217, opened 2024-07-29
Adds support for additional Python versions

kylebgorman:
Enables support for Python > 3.10.

At some earlier point we were stuck on PyTorch <2, and there was no support for Python 3.10 or later. #173 freed us from this restriction, but left in the restriction to PyTorch-Lightning <2. However, we were pinned on Python <= 3.10 by PyTorch, not PyTorch-Lightning, so we can now add support for additional Pythons.

Python 3.12 is the actively developed branch; Python 3.8-3.11 are are on long-term (i.e., security-fix) support; Python 3.8 is just about to go out of date and Python 3.13 is approaching prerelease: https://devguide.python.org/versions/. So this adds support for all supported Python versions (except 3.8, which was already old when we started this project and which we never supported in the first place).

I have updated the patch-level since this is worth putting on PyPI; I also have enabled testing on additional versions.

This is closely related to, but does not close, #60.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
       command-run:
         type: string
     docker:
-      - image: cimg/python:3.10
+      - image: cimg/python:3.12
     steps:
       - checkout
       - run:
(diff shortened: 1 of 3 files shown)


kylebgorman:
I've tested this on Python 3.11 and Python 3.12 and all the architectures I believed to be in good condition are WAI so I'm merging this. (My mental stack can't handle more than a few of my PRs open at a time.)

PR #219, opened 2024-07-29
Moves embedding initialization to its own module

### LINKED ISSUE #215: Missing init_embeddings in Transformer module
There's a call to `init_embeddings` in the Transformer invariant feature encoder. That only exists in the model. This should be moved

kylebgorman:
These were previously stored as static methods in the base model. But they are also needed at the module level, so we separate them. This should fix the break in the feature invariant transformer reported in #215. (Closes #215.)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,7 +7,7 @@ jobs:
       command-run:
         type: string
     docker:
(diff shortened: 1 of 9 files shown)


kylebgorman:
Holding off on this until I have the bandwidth to test it.

PR #220, opened 2024-07-30
Fixes bug with optional features encoder

kylebgorman:
The logic in `train.py` was to use a separate features encoder module if:

- features are present in the datamodule
- the model requires features to be separate from source (e.g., pointer-generators)
- and a separate features encoder module is requested.

This is incorrect; the logic should read:

- features are present in the datamodule, and
      - a separate features encoder module is requested, or
      - the model requires features to be separate from source.

In the case where the model requires it but a specific features encoder module is not specified, we use the same module as the source encoder; e.g., a pointer-generator LSTM will have an LSTM source encoder and an LSTM features encoder.

The README was also out of date: the flags are `--source_encoder_arch` and `--features_encoder_arch` (was missing the `_arch` part).

### DIFF
--- a/README.md
+++ b/README.md
@@ -221,7 +221,7 @@ additional flags. Supported values for `--arch` are:
     `--source_attention_heads`; default: `4`).
 
 The user can override the default encoder architectures. One can override the
-source encoder using the `--source_encoder` flag:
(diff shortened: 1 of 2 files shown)


kylebgorman:
This is now thoroughly tested so I'm submitting.

PR #224, opened 2024-07-30
Disables `norm_first` in the transformer encoder

### LINKED ISSUE #214: Nested tensor warning
The transformer logs this on creation:

```UserWarning: enable_nested_tensor is True, but self.use_nested_tensor is False because encoder_layer.norm_first was True```

There are two things we can do about it:

1. To silence the error message but nothing else, we can add [`enable_nested_tensor=False` here](https://github.com/CUNY-CL/yoyodyne/blob/e04fedb940e94132ae729aa9df92251a3003dee4/yoyodyne/models/modules/transformer.py#L194).
2. Or we can set [`norm_first=False` here](https://github.com/CUNY-CL/yoyodyne/blob/e04fedb940e94132ae729aa9df92251a3003dee4/yoyodyne/models/modules/transformer.py#L191).

I don't know what the consequences of the latter are, really (I have no intuitions about when to norm) so I piloted it and it doesn't seem to affect performance much.

kylebgorman:
This allows us to use the experimental, but supposedly faster, nested tensor API:

https://pytorch.org/docs/stable/nested.html

As the documentation indicates, this apparently is particularly helpful with padding. 

Closes #214. Mutually exclusive with #225.

### DIFF
--- a/yoyodyne/models/modules/transformer.py
+++ b/yoyodyne/models/modules/transformer.py
@@ -189,7 +189,7 @@ class TransformerEncoder(TransformerModule):
             nhead=self.source_attention_heads,
             dropout=self.dropout,
             activation="relu",
-            norm_first=True,
+            norm_first=False,
             batch_first=True,
         )
         return nn.TransformerEncoder(


kylebgorman:
Having browsed this, it seems like we do want pre-norm. So in #239 I lightly modify the TODO to say we should enable nested tensors if they ever support pre-norm. I will close this in the meantime, but it'll stay in the records.

PR #225, opened 2024-07-30
Disables the use of nested tensors

### LINKED ISSUE #214: Nested tensor warning

kylebgorman:
We can just silence the warning by agreeing not to use the nested tensors backend.

Closes #214. Mutually exclusive with #224.

### DIFF
--- a/yoyodyne/models/modules/transformer.py
+++ b/yoyodyne/models/modules/transformer.py
@@ -196,6 +196,7 @@ class TransformerEncoder(TransformerModule):
             encoder_layer=encoder_layer,
             num_layers=self.layers,
             norm=nn.LayerNorm(self.embedding_size),
+            enable_nested_tensor=False,
         )
 
     @property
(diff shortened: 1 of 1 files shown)


kylebgorman:
@Adamits I propose we do this now (to silence the warning) and then keep #224 around in case we decide to use this feature. Any objection?

PR #226, opened 2024-07-30
Cleans up documentation on the separate feature transformer decoder

kylebgorman:
The feature invariant decoder reserves half of its capacity for features. This requires that its size be divisible by 2. I raise an explicit error in the unlikely case this is hit, and remove the FIXME. I am also using the more idiomatic // 2 rather than int(_ / 2).


kylebgorman:
This is tested and is doing fine (the FIT is actually a pretty great model on the Polish data; it learns fast) so I'm merging now.

PR #228, opened 2024-08-01
Unifies tensor finalizing during eval

### LINKED ISSUE #207: Transducer over-copying
In #140 I observed the transducer val loss converging but the accuracy staying the same for the english-medium dataset in the sigmorphn-2017 shared task dataset. When predicting, it seemed to learn to primarily copy the lemma.

We should test this on some other datasets, and try to reproduce results with proper hyperparameters, etc, to determine if there is a bug.

In the master branch, predict.py is broken for transducer so we may be blocked by that (#197) for debugging this.

Adamits:
Currently, the transducer predictions are right-padded with EOS (and indeed it seems to learn to make predictions in this way). Our gold tensors are right padded with PADs, and other models have their predictions 'finalized' so that everything after the first EOS symbol is replaced with a PAD.

This ensures that the transducer is finalized in the same way, which fixes some accuracty errors I was seeing. 

Note that in my tests for english-medium sigmorphon-2017 data, the transducer still seems to learn to copy the lemma and not correctly inflect, so there still may be further issues that I have been investigating in #207 

EDIT: closes #207

### DIFF
--- a/yoyodyne/models/transducer.py
+++ b/yoyodyne/models/transducer.py
@@ -7,7 +7,7 @@ import numpy
 import torch
 from torch import nn
 
-from .. import data, defaults
+from .. import data, defaults, util
 from . import expert, lstm, modules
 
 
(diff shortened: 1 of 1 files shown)


Adamits:
Nevermind on my note about copying. This did, in fact, learn to inflect (~71% acc after 20 epochs). I will update #207.

kylebgorman (approved):
LGTM

PR #230, opened 2024-08-01
Adds explicit encoding arg for text files

### LINKED ISSUE #229: Fix encoding when opening text files
"Normal" platforms more or less all default into UTF-8 encoding these days, but apparently not Windows. Our sole (that I know of) Windows user was able to get everything working simply by adding `encoding="utf8"` when opening files for reading or writing (I think this be limited to `data/tsv.py` and maybe `prediction.py`). This of course is harmless on other platforms.

kylebgorman:
Closes #229.

### DIFF
--- a/yoyodyne/data/tsv.py
+++ b/yoyodyne/data/tsv.py
@@ -61,7 +61,7 @@ class TsvParser:
 
     @staticmethod
     def _tsv_reader(path: str) -> Iterator[str]:
-        with open(path, "r") as tsv:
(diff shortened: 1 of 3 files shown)


Adamits (approved):
LGTM

Can you think of any reason to expose this as a flag in the CLI?

kylebgorman:
> Can you think of any reason to expose this as a flag in the CLI?

Good question but no. This code can't and won't ever run on any computer old enough to not work with UTF-8.

PR #231, opened 2024-08-02
Clears the cache after an OOM during sweeps

### LINKED ISSUE #227: Cascading OOMs in sweep script
I believe this was reported before, though maybe not so concisely: if you have a OOM in the inner loop of the sweep script, the memory is not freed and all subsequent runs will also OOM immediately. It's not clear to me whether there is a race condition (i.e., Python is _going_ to free the memory but the next sweep starts before it gets a chance to) or whether it just leaks if there's an interior crash in the `function` argument to the W&B agent; probably the latter. 

Either way, I think the issue is that the in the case of OOM, [these lines](https://github.com/CUNY-CL/yoyodyne/blob/46c79b4d20414870190c78c1e1d561cfd49419ac/examples/wandb_sweeps/train_wandb_sweep.py#L52) are never reached. It seems to me that we need to wrap `train_sweep` with a `try:` block, and move the deallocation/cache eviction code to a `finally:` block. If that works, I'll know immediately.

I am questioning whether the `try:` block [here](https://github.com/CUNY-CL/yoyodyne/blob/46c79b4d20414870190c78c1e1d561cfd49419ac/examples/wandb_sweeps/train_wandb_sweep.py#L83) is actually necessary.

kylebgorman:
As reported in #227, memory is not freed after an OOM during the sweep, meaning that, usually, all subsequent sweeps also fail. We solve this by:

1. Cleaning up the `train.train` API with a more friendly endpoint. This has the desirable effect that there is no longer any need to keep `train_wandb_sweep.py` "up to date" with trainer interface: it just takes `args` and handles the rest.
2. Wrapping the call to `train.train` in a `try`/`except` with a `finally:` block that clears the CUDA cache.

Closes #227.

Sample test:

```
# This has deliberately huge batch sizes---kbg.
$ cat config/bad_config.yaml
method: random
metric:
  name: val_accuracy
  goal: maximize
parameters:
  # Constants.
  arch:
    value: attentive_lstm
  # Hyperparameters.
  encoder_layers:
    values: [1, 2]
  embedding_size:
    distribution: q_uniform
    q: 16
    min: 16
    max: 512
  hidden_size:
    distribution: q_uniform
    q: 64
    min: 64
    max: 1024
  dropout:
    distribution: uniform
    min: 0
    max: 0.5
  label_smoothing:
    distribution: uniform
    min: 0.0
    max: 0.2
  batch_size:
    distribution: q_uniform
    q: 1024
    min: 1024
    max: 4096
  beta1:
    distribution: uniform
    min: 0.8
    max: 0.999
  beta2:
    distribution: uniform
    min: 0.98
    max: 0.999
  learning_rate:
    distribution: log_uniform_values
    min: 0.00001
    max: 0.01
  scheduler:
    values: [null, reduceonplateau, warmupinvsqrt]
  reduceonplateau_factor:
    distribution: uniform
    min: 0.1
    max: 0.9
  reduceonplateau_patience:
    distribution: q_uniform
    q: 1
    min: 1
    max: 5
  min_learning_rate:
    distribution: log_uniform_values
    min: 0.000001
    max: 0.001
  warmup_steps:
    distribution: q_uniform
    q: 50
    min: 50
    max: 5000
$ export ENTITY=cuny-cl
$ export PROJECT=OOM
$ wandb sweep --entity "${ENTITY}" --project "${PROJECT}" config/b
ad_config.yaml
wandb: Creating sweep from: config/bad_config.yaml
wandb: Creating sweep with ID: vjt25un4
wandb: View sweep at: https://wandb.ai/cuny-cl/OOM/sweeps/vjt25un4
wandb: Run sweep agent with: wandb agent cuny-cl/OOM/vjt25un4
$ ../../yoyodyne-wrap/examples/wandb_sweeps/./train_wandb_sweep.py --entity "${ENTITY}" --project "${PROJECT}" --sweep_id vjt25un4 --count 50 --experiment "${PROJECT}" --model_dir models --train ../tsv/eng/polyg2p-train.tsv --val ../tsv/eng/polyg2p-dev.tsv --source_sep " " --target_sep " " --features_col 3 --features_encoder linear --accelerator gpu --gradient_clip_val 3 --max_epochs 100 --max_time 01:00:00:00 --patience 10 --seed 1985
...
```

### DIFF
--- a/examples/wandb_sweeps/train_wandb_sweep.py
+++ b/examples/wandb_sweeps/train_wandb_sweep.py
@@ -32,10 +32,10 @@ def train_sweep(args: argparse.Namespace) -> None:
         setattr(args, key, value)
     try:
         train.train(args)  # Ignoring return value.
-    except RuntimeError:
+    except RuntimeError as error:
         # TODO: consider specializing this further if a solution to
         # https://github.com/pytorch/pytorch/issues/48365 is accepted.
-        pass
+        util.log_info(f"Runtime error: {error!s}")
     finally:
         # Clears the CUDA cache.
         torch.cuda.empty_cache()
--- a/examples/wandb_sweeps/train_wandb_sweep.py
+++ b/examples/wandb_sweeps/train_wandb_sweep.py
@@ -10,14 +10,10 @@ import pytorch_lightning as pl
 import torch
 import wandb
 
-from yoyodyne import sizing, train, util
+from yoyodyne import train, util
 
 
 warnings.filterwarnings("ignore", ".*is a wandb run already in progress.*")
-
-
-class Error(Exception):
-    pass
 
 
 def train_sweep(args: argparse.Namespace) -> None:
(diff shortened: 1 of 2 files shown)


kylebgorman:
I haven't heard but I'm going to go ahead and merge this, having tested it locally.

PR #240, opened 2024-08-27
Moves initializer args to class-level docstrings.

kylebgorman:
This is what Python tooling (and LightingCLI) expect. Cf. #239.


kylebgorman:
This ought to be no change in behavior to the actual runtime code and I've tested locally, so I will now merge.

PR #233, opened 2024-08-04
Fixes things I broke in the Transducer with single-embeddings design …

Adamits:
…and when updating inference.

PR #244, opened 2024-08-30
Cleanup to schedulers

### LINKED ISSUE #243: Epoch-based warmup
Standard procedure for LR schedulers is to step once per epoch unless otherwise specified. We have one exception to this: the warmup inverse-square-root scheduler. 

This is sort of important in the context of #60 aka the LightingCLI migration. If we use the inherited optimization routine (and we otherwise can) there's no way to pass the information that the warmup inverse-square-root scheduler is a step-based one, unless we override the optimization stepping and hardcode a check (is this a `WarmupInverseSquareRootSchedule` object?) into the base model. Rather than doing this, I propose we just change the scheduler to work with epochs rather than steps.

kylebgorman:
Tags schedulers with their configuration metadata.

Simplifies the `metrics` helper library somewhat to support.

Note that the LightningCLI migration will make it possible for users to use basically any scheduler in PyTorch, Lightning, or specified by the user, so this is a temporary simplification to enable that migration.

Closes #243.


yoyodyne/defaults.py
```diff
@@ -60,11 +60,8 @@
 LOG_WANDB = False
 TEACHER_FORCING = True
 
-# LR Scheduler arguments.
-WARMUP_STEPS = 0
-START_FACTOR = 1 / 3
-END_FACTOR = 1.0
-TOTAL_DECAY_STEPS = 5
+# LR scheduler arguments.
+WARMUP_EPOCHS = 5
```
Adamits, line 64: Any reason for 5?

Adamits (approved):
> We redefine schedulers to work at the epoch level. This seems to be standard practice with a few exceptions.

Can you support this? I thought warmup was standard and is always done by number of steps, no? Did you run a test to train a transformer with warmup at the epoch level? I usually use 4k steps for inflection from the Shije Wu transformer for character seq2seq tasks paper. Unsure how this actually translates to epochs since you'd be making fewer updates for the same number of steps. Do you have a suggestion or reference for how we should think of the warmup_epochs hyperparameter?


In general I am not in favor of removing features that one might want to use for experimentation, even if *we* aren't using them now. But given
> Note that the LightningCLI migration will make it possible for users to use basically any scheduler in PyTorch, Lightning, or specified by the user, so this is a temporary simplification to enable that migration.

then it seems fine to me if it makes this easier. I am going to approve, but am not 100% sure if the warmup steps -->epochs makes sense to me.

yoyodyne/defaults.py
kylebgorman: Seemed like a good guess as any. This is giving me SOTA (?) performance in another project without any tuning...

kylebgorman:
> > We redefine schedulers to work at the epoch level. This seems to be standard practice with a few exceptions.
> 
> Can you support this? I thought warmup was standard and is always done by number of steps, no? 

I would not say that this is a well-established principle that one has to do it one way or the other. The Wu paper is somewhat unique in that they talk in steps and not epochs, not just for the scheduler but also use a "max_steps" rather than "max_epochs". Certainly they're not the only people to do that but it's uncommon, I think. 

> Did you run a test to train a transformer with warmup at the epoch level?

I have and it works great. 

>  I usually use 4k steps for inflection from the Shije Wu transformer for character seq2seq tasks paper. Unsure how this actually translates to epochs since you'd be making fewer updates for the same number of steps. Do you have a suggestion or reference for how we should think of the warmup_epochs hyperparameter?

I think I'd just take 4k and divide by the number of batches per epoch; so if there are 400 batches per epoch then you'd set `--warmup_epochs 10`. I am using W&B to eyeball this and hoping to see it hit the warmup. 

Supporting both step and epoch-based schedulers requires some way to indicate whether it's a step or epoch-based scheduler, and this isn't pretty. It is near impossible, as far as I can tell, in Lightning; it actually just has special casing that a certain scheduler is step-based and assumes all others are epoch-based. We are sort of overriding that now but it requires us to basically register every option in every scheduler we support as a flag. 

The vision in LightningCLI is that all this configuration stuff lives in configuration files, not duplicated messily in the base model. But it doesn't have any way to know whether a scheduler is step- or epoch-based because PyTorch doesn't record that kind of information about its schedulers, and some schedulers are imaginable either step- or epoch-based. (However others only make sense as epoch-based. What I assert is that **all** of them can be made epoch-based.)

What do you think?

Adamits:
Hey Kyle, I had written a long response to this last week and I do not see it anywhere. Let's see if I can remember:


> > > We redefine schedulers to work at the epoch level. This seems to be standard practice with a few exceptions.
> > 
> > 
> > Can you support this? I thought warmup was standard and is always done by number of steps, no?
> 
> I would not say that this is a well-established principle that one has to do it one way or the other. The Wu paper is somewhat unique in that they talk in steps and not epochs, not just for the scheduler but also use a "max_steps" rather than "max_epochs". Certainly they're not the only people to do that but it's uncommon, I think.

I am not sure it is unique in that way---they got that stuff from the attention is all you need paper, and generally as datasets get bigger I think most discussion is in terms of steps rather than epochs. to me: how many gradient updates did I make? seems more important than: how many times did I see all of the data? Of course, idk that this means schedulers need to be directed this way---I have never experimented much here---I am just making the point that I think configuring things wrt steps rather than epochs seems more natural for most use-cases and reproducibility to me.

> > Did you run a test to train a transformer with warmup at the epoch level?
> 
> I have and it works great.
> 
> > I usually use 4k steps for inflection from the Shije Wu transformer for character seq2seq tasks paper. Unsure how this actually translates to epochs since you'd be making fewer updates for the same number of steps. Do you have a suggestion or reference for how we should think of the warmup_epochs hyperparameter?
> 
> I think I'd just take 4k and divide by the number of batches per epoch; so if there are 400 batches per epoch then you'd set `--warmup_epochs 10`. I am using W&B to eyeball this and hoping to see it hit the warmup.
> 
> Supporting both step and epoch-based schedulers requires some way to indicate whether it's a step or epoch-based scheduler, and this isn't pretty. It is near impossible, as far as I can tell, in Lightning; it actually just has special casing that a certain scheduler is step-based and assumes all others are epoch-based. We are sort of overriding that now but it requires us to basically register every option in every scheduler we support as a flag.

Not sure why its impossible---this should be configurable no? Though maybe we need some conditions in a child class of the actual torch scheduler? Either way, it is messy so it seems fine to remove for a bit while we update things that will enable this more cleanly if I understand correctly.

> The vision in LightningCLI is that all this configuration stuff lives in configuration files, not duplicated messily in the base model. But it doesn't have any way to know whether a scheduler is step- or epoch-based because PyTorch doesn't record that kind of information about its schedulers, and some schedulers are imaginable either step- or epoch-based. (However others only make sense as epoch-based. What I assert is that **all** of them can be made epoch-based.)
> 
> What do you think?

mechanically, all of them can be epoch based without raising an error sure, but im not sure you can reproduce the exact optimization of a step based scheduler if you do not update the scheduler per-step... I suppose we can default it to work this way, and write a not eon how to override things in python to make your scheduler step-based if you choose?

kylebgorman:
> I am not sure it is unique in that way---they got that stuff from the attention is all you need paper, and generally as datasets get bigger I think most discussion is in terms of steps rather than epochs. to me: how many gradient updates did I make? seems more important than: how many times did I see all of the data? Of course, idk that this means schedulers need to be directed this way---I have never experimented much here---I am just making the point that I think configuring things wrt steps rather than epochs seems more natural for most use-cases and reproducibility to me.

Not exactly since a step may be less than a gradient update (in fact often is in my settings). 

Do you know of any other scheduler you'd like to support that is (necessarily, usually, always) step-based? 

Lightning treats epoch-based as the default. It seems to me before the transformers paper, people either used no scheduler or they reduced on plateau (which is going to inherently be epoch-based). After this paper at least some people working with transformers use warmup. The paper defines this in terms of steps but it is not hard to redefine in terms of epochs, and other than the loss of precision it's not obvious you lose anything from doing so. 

> Not sure why its impossible---this should be configurable no? Though maybe we need some conditions in a child class of the actual torch scheduler? Either way, it is messy so it seems fine to remove for a bit while we update things that will enable this more cleanly if I understand correctly.

There are three logical possibilities, as I see it:

1. a scheduler is indifferent and we can interpret "step" either to be a batch or an epoch; we just need to pass "which". Sadly this is a pretty big lift in LightningCLI. The way I'd do it is to subclass every scheduler and put a place to put this information. I then need to override "configure_optimizers" to use this information (in LightningCLI you otherwise don't need to override this).
3. a scheduler only works in epoch mode; this is the case for reduce-on-plateau; I am not sure how this should interact with (1); does the scheduler have a slot for the metadata but it throws a configuration-time error? 
4. ditto but the scheduler only works in step mode; I don't know if this exists but it poses the same issues as (2) if so.

Thoughts on this alternative design? This would be for the future.

I could also make separate subclasses for (1) and then tag them (i.e., add a property) as to whether they're epoch- or step-based.
 
> mechanically, all of them can be epoch based without raising an error sure, but im not sure you can reproduce the exact optimization of a step based scheduler if you do not update the scheduler per-step... I suppose we can default it to work this way, and write a not eon how to override things in python to make your scheduler step-based if you choose?

It's true you can't run 3500 steps of warmup if the batches don't divide that way but I actually think working this way is uncommon. Maybe I'm wrong. 

I appreciate the discussion here---it's helpful. I think I may want to push this through at the last minute before I do the CLI migration (for which I am waiting for you and Travis to close the other PRs), and we can consider my more elaborate design at a later date.

Adamits:
> > I am not sure it is unique in that way---they got that stuff from the attention is all you need paper, and generally as datasets get bigger I think most discussion is in terms of steps rather than epochs. to me: how many gradient updates did I make? seems more important than: how many times did I see all of the data? Of course, idk that this means schedulers need to be directed this way---I have never experimented much here---I am just making the point that I think configuring things wrt steps rather than epochs seems more natural for most use-cases and reproducibility to me.
> 
> Not exactly since a step may be less than a gradient update (in fact often is in my settings).

I assume you mean the case of gradient accumulation (don't know how else 1 step would not = one gradient update since to me thats the definition of a step)---and there I think that's an abuse of definition. In gradient accumulation we are redefining one step to include `k` forward passes before 1 gradient update in order to simulate a larger batch size. I think thats the only case where number of forward and backward passes differs no?

> Do you know of any other scheduler you'd like to support that is (necessarily, usually, always) step-based?
> 
> Lightning treats epoch-based as the default. It seems to me before the transformers paper, people either used no scheduler or they reduced on plateau (which is going to inherently be epoch-based). After this paper at least some people working with transformers use warmup. The paper defines this in terms of steps but it is not hard to redefine in terms of epochs, and other than the loss of precision it's not obvious you lose anything from doing so.

Yeah and pytorch also seems to default to epoch, which I think is a good argument to do it that way.

> > Not sure why its impossible---this should be configurable no? Though maybe we need some conditions in a child class of the actual torch scheduler? Either way, it is messy so it seems fine to remove for a bit while we update things that will enable this more cleanly if I understand correctly.
> 
> There are three logical possibilities, as I see it:
> 
> 1. a scheduler is indifferent and we can interpret "step" either to be a batch or an epoch; we just need to pass "which". Sadly this is a pretty big lift in LightningCLI. The way I'd do it is to subclass every scheduler and put a place to put this information. I then need to override "configure_optimizers" to use this information (in LightningCLI you otherwise don't need to override this).
> 2. a scheduler only works in epoch mode; this is the case for reduce-on-plateau; I am not sure how this should interact with (1); does the scheduler have a slot for the metadata but it throws a configuration-time error?
> 3. ditto but the scheduler only works in step mode; I don't know if this exists but it poses the same issues as (2) if so.
> 
> Thoughts on this alternative design? This would be for the future.

Right I see. This is similarly a good argument that we should default to just work in epoch mode (1).

> I could also make separate subclasses for (1) and then tag them (i.e., add a property) as to whether they're epoch- or step-based.
> 
> > mechanically, all of them can be epoch based without raising an error sure, but im not sure you can reproduce the exact optimization of a step based scheduler if you do not update the scheduler per-step... I suppose we can default it to work this way, and write a not eon how to override things in python to make your scheduler step-based if you choose?
> 
> It's true you can't run 3500 steps of warmup if the batches don't divide that way but I actually think working this way is uncommon. Maybe I'm wrong.

I don't follow. The fundamental difference I am describing, and the thing I don;t think that yoyodyne should be opinionated on (unless we need to be due to constraints in the libraries we use -- which it seems we do need this constraint for smooth support) is that if a scheduler is epoch-based, then the learning rate only changes once per epch, right?

So we make `k` gradient updates all with the same LR, where `k` is the number of steps in an epoch. For step-based, each of the `k` updates has a slightly different learning rate. This is an inherent difference which I believe you are saying does not matter much in your experiments. However, what I am saying is I would rather not make one of these impossible, since AFAIK step-based schedulers *are* common. 

It seems to me that despite my argument here, the best course of action is to make your suggested update, which will make step-based learning-rate schedulers not possible out of box, but add a note somewhere about this, and suggesting where a developer should customize the code to write such a scheduler.

> I appreciate the discussion here---it's helpful. I think I may want to push this through at the last minute before I do the CLI migration (for which I am waiting for you and Travis to close the other PRs), and we can consider my more elaborate design at a later date.

Sure. FYI #233 looks like its pasted into and updated in #247. I suggest we keep #233 open until #247 is merged, but we should not need both.

Adamits (approved):
Shall we discuss where to put a note acknowledging that we don;t support step-based scheduling and suggestions for how to write one? This doesn;t ahve to go into this PR.

LGTM

kylebgorman:
I do acknowledge the distinction between changing LR mid-epoch. I just don't see any reason to suppose it matters yet, which is why I am slightly nonchalant about this change.

> I assume you mean the case of gradient accumulation (don't know how else 1 step would not = one gradient update since to me thats the definition of a step)---and there I think that's an abuse of definition. In gradient accumulation we are redefining one step to include `k` forward passes before 1 gradient update in order to simulate a larger batch size. I think thats the only case where number of forward and backward passes differs no?

That's what I meant, yeah. 
 
> Shall we discuss where to put a note acknowledging that we don;t support step-based scheduling and suggestions for how to write one? This doesn't have to go into this PR.

Just added some information of this form to the module docstring.

I wanted to draw your attention to one other late-breaking change here; I fixed some typing stuff for `configure_optimizers` and its helper methods.

Adamits:
Those notes and the typing changes LGTM.

kylebgorman:
I took this in a totally different direction: I now tag the supported schedulers with the config data they need so we can support step-based stuff. (I still maintain that the step/epochs distinction is not an important one here.) So nothing is lost, this is just cleanup that moves scheduler-specific logic from the base model to the `schedulers.py` module. 

We can actually keep this up in LightningCLI if desired.

@Adamits please take one last look to see if you like this.

kylebgorman:
gentle ping on this @Adamits, I redesigned as described in previous comment.

Adamits (commented):
This is nice! I left two very small things.

This might be more appropriate for an email but while thinking about this, I was wondering if we could simplify the metric abstraction by porting the evaluator code into it, so that a metric both specifies how it should be used for logging AND the implementation for measuring that metric.

When you see `metric.get_metric()` it kind of sounds like it would implement the metric you want to measure, but in practice this is held by a completely different interface.

I can think about what this would require more, but do you see any problems with this?

yoyodyne/defaults.py
```diff
@@ -61,7 +61,7 @@
 TEACHER_FORCING = True
 
 # LR scheduler arguments.
-WARMUP_EPOCHS = 5
+WARMUP_STEPS = 5
```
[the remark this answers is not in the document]
kylebgorman: Have reverted back to 0.

kylebgorman:
> This might be more appropriate for an email but while thinking about this, I was wondering if we could simplify the metric abstraction by porting the evaluator code into it, so that a metric both specifies how it should be used for logging AND the implementation for measuring that metric.
> 
> When you see `metric.get_metric()` it kind of sounds like it would implement the metric you want to measure, but in practice this is held by a completely different interface.
> 
> I can think about what this would require more, but do you see any problems with this?

No sounds good, take a stab at it sometime?

PR #250, opened 2024-10-07
Moves parameter counting into base model

kylebgorman:
We have special logic inside the trainer which logs, to W&B (and only if enabled), the number of parameters. This makes two changes:

1. the actual logic into summing the parameters across the elements is moved into the base model.
2. the name of the logged quantity is given a more informative name.

Nothing else changed.


Adamits (approved):
Perfect!

PR #251, opened 2024-10-08
Fixes init_hiddens calling convention

kylebgorman:
`init_hiddens` is always called with the batch size as the first argument and the number of decoder layers as the second. Since it is an instance method and the latter is constant, it should not be passed in directly but rather read from `self`.


yoyodyne/models/lstm.py
```diff
@@ -196,11 +195,9 @@ def beam_decode(
...
             )
         # Initializes hidden states for decoder LSTM.
-        decoder_hiddens = self.init_hiddens(
-            encoder_out.size(0), self.decoder_layers
-        )
-        # Log likelihood, last decoded idx, all likelihoods, hiddens tensor.
-        histories = [[0.0, [special.START_IDX], [0.0], decoder_hiddens]]
+        decoder_hiddens = self.init_hiddens(batch_size)
+        # Log likelihood, last decoded idx, all likelihoods,  hiddens tensor.
+        histories = [[0.0, [self.start_idx], [0.0], decoder_hiddens]]
```
Adamits, line 200: Careful merging this with #253 :)

Adamits (approved):
Good catch on this, LGTM

PR #252, opened 2024-10-08
Generates mask as needed.

### LINKED ISSUE #245: Generate masks as needed?
Right now padded tensors are actually two tensors: the data and a mask tensor indicating the padding. However, the padding is sometimes not needed at all and some times it's only needed briefly. (As far as I know there's no place where we need to collect gradients over it or anything like that.) We should consider whether we can generate the padding tensor as needed in those locations, which might serve to shrink the memory requirements for batches and thus help avoid OOMs by generating the mask only where it's needed.

I took a quick informal study of how these padding masks (not to be confused with causal masks) are used in a few of our models:

* In the LSTM module, the encoder does not make use of a mask as far as I can see; the inattention decoder uses it to figure out where the padding ends but that could easily be rewritten using `torch.where`, `torch.nonzero`, or `torch.index`; the soft attentive decoder uses the mask for the attention computation but we could probably generate it at that site; the hard attentive decoder does something similar to the attentive decoder. 
* In the transformer module, I'm a little confused but it looks like the positional encoding regenerates the mask; the transformer encoders and decoders use the provided masks though they could be regenerated there.
* In our pointer generator model, things are roughly similar to the LSTM module, but I note we already regenerate the padding mask for the smooth NLLLoss implementation, showing this is feasible.

kylebgorman:
Closes #245.

### DIFF
--- a/yoyodyne/data/batches.py
+++ b/yoyodyne/data/batches.py
@@ -11,7 +11,7 @@ from torch import nn
 
 
 class PaddedTensor(nn.Module):
-    """A tensor and its mask.
+    """A tensor.
 
     This is ordinarily used for padding a tensor list, so it represents
     one of (source, target, features) for a batch.
@@ -29,8 +29,8 @@ class PaddedTensor(nn.Module):
         pad_len (int, optional): desired length for padding.
     """
 
+    pad_idx: int
     padded: torch.Tensor
-    mask: torch.Tensor
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


Adamits:
Just to confirm, I expect that `self.padded == self.pad_idx` runs in O(N) -- here batch_size * seq_len -- and is heavily optimized by torch right? In that case this change seems like a huge upgrade. I guess this just runs once per forward pass for any model that requires it?

PR #253, opened 2024-10-08
The four special indices are now constants.

kylebgorman:
This saves a huge amount of boilerplate: net decrease is over 100 lines.


Adamits:
If I understand correctly, this removes passing the 4 special indices through each method, and just calls them from the global defaults.

The only issue I can foresee is if (i) these are exposed to the CLI and (ii) someone chooses to change them (suppose for some reason <S> is a valid vocab item). Thoughts?

Also, this seems like a change where its easy to make a small mistake that fully breaks a rogue line of code somewhere. Did you run several architectures a for a couple steps to test for regression?

kylebgorman:
> If I understand correctly, this removes passing the 4 special indices through each method, and just calls them from the global defaults.

That's right.
 
> The only issue I can foresee is if (i) these are exposed to the CLI and (ii) someone chooses to change them (suppose for some reason `<S>` is a valid vocab item). Thoughts?

There is nothing stopping people from violating the "reserved symbols" we document, that's true, but that would break things both before and after this change. 

> Also, this seems like a change where its easy to make a small mistake that fully breaks a rogue line of code somewhere. Did you run several architectures a for a couple steps to test for regression?

Yep.

kylebgorman:
> > There is nothing stopping people from violating the "reserved symbols" we document, that's true, but that would break things both before and after this change.
> 
> What else would it break? Can't you say: actually the SOS token should be `$S$` and we use `<S>` as a vocab token? This is an edge case id like to consider, but IDK if this would have **actually** worked before this change so dont let me hold it up.

No, there's no configurable way to change that what SOS or EOS is.

Were you to use `<S>` as an input symbol in your TSV files, it would map to the wrong thing. But that's clearly documented.

PR #254, opened 2024-10-08
GRU support

### LINKED ISSUE #180: GRU support
It would be nice if our RNN encoders and decoders, which are currently LSTMs, could be replaced with GRUs.

A simple CLI option would be something like this. Rename the architectures `Rnn` and `AttentiveRnn` and modify the architecture options to reflect the change. By default, make the RNN type default to LSTM, but add a flag (`--rnn_type`? just `--rnn`) to select a GRU instead.

Alternatively, but only if necessary, one could project this choice down into the architecture level so that one can choose between an `AttentiveLstm` and an `AttentiveGru` etc. 

I am assigning this to @bonham79 since I know it's an interest of his.

kylebgorman:
This adds GRU support; everywhere there is an LSTM model, there is now a GRU model too.

I initially tried to make RNN type a general flag but because LSTMs return the cell state in addition to the hidden state, and because various models need to reshape, average, or otherwise manipulate that cell state, this was really not feasible. Thefore I just create, for each model that was previously "LSTM-backed", an abstract class called `FooRNN{Encoder,Decoder,Model}`. `FooLSTM` subclasses this and returns a LSTM module (it may also have special logic in the forward method, or decode method, or whatever), as does `FooGRU`.

I experimented with traditional Elman RNNs (they have the same simpler interface as GRUs) but performance was absymal so I'm not going to bother.

All models have been tested on CPU and GPU.

Other changes:

* #251 is also implied here, but I separated it out for review.
* The names got confusing so I also went ahead and replaced `EncoderDecoder` in our naming convention with simply just `Model`.

Closes #180. (Note however there's still plenty to do to study the effects this has.)


kylebgorman:
There is at least one case where both $h$ and $c$ are manipulated: [see here](https://github.com/CUNY-CL/yoyodyne/blob/2f733ce1459b4bc20368f1d13e114fca6702a7e6/yoyodyne/models/pointer_generator.py#L447). This seemed to necessitate the design I pursue here.

I suspect there is some abstraction that would reduce the amount of boilerplate here, but I'm not even sure if it's worth tracing down.

kylebgorman:
> I didnt look everywhere, but iirc when we manipulate the state, we manipulate the `h` rather than the cell right? so if we just have a property for "the thing that should be manipulated" and GRU returns the hidden tensor that is returned from its forward call, and LSTM just returns the first tensor in the tuple, then I think we should be good.

There is at least one case where both $h$ and $c$ are manipulated: [see here](https://github.com/CUNY-CL/yoyodyne/blob/2f733ce1459b4bc20368f1d13e114fca6702a7e6/yoyodyne/models/pointer_generator.py#L447). This seemed to necessitate the design I pursue here.

I suspect there is some abstraction that would reduce the amount of boilerplate here, but I'm not even sure if it's worth tracing down. If you can think of anything though please go ahead and suggest!

README.md
```diff
@@ -185,54 +184,65 @@ not an attention mechanism is present. This flag also specifies a default
...
+    an attention mechanism. The initial hidden state is treated as a learned
     parameter.
--   `hard_attention_lstm`: This is an LSTM encoder/decoder modeling generation
-    as a Markov process. By default, it assumes a non-monotonic progression over
+-   `attentive_lstm`: This is similar to the `attentive_gru` but instead uses an
+    LSTM decoder and encoder (by default).
+-   `gru`: This is an GRU decoder with GRU encoders (by default); in lieu of an
+    attention mechanism, the last non-padding hidden state of the encoder is
+    concatenated with the decoder hidden state.
+-   `hard_attention_gru`: This is an GRU encoder/decoder modeling generation as
```
bonham79, line 195: oh, this is exciting

yoyodyne/models/hard_attention.py
```diff
@@ -503,3 +426,189 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             help="Width of attention context "
             "(hard attention architectures only). Default: %(default)s.",
         )
+
+
+class HardAttentionGRUModel(HardAttentionRNNModel):
```
bonham79, line 431: Hmm, this redundancy of boilertplate strikes me as unnecessary. Is there any way we can just keep the underlying decoder structure as abstracted and just pass a flag that allows the attention decoder to specify architecture in module construction?

yoyodyne/models/modules/attention.py
```diff
@@ -90,6 +90,6 @@ def score(
         # -> B x seq_len x hidden_size
         m = self.M(concat)
         # -> B x seq_len x 1.
-        scores = self.V(torch.tanh((m)))
+        scores = self.V(torch.tanh(m))
```
bonham79, line 93: san check, tanh doesn't have any weirdness where everything as to be a tuple? (kinda like how cat needs an iterable?

yoyodyne/models/modules/base.py
```diff
@@ -13,10 +12,12 @@
 
 @dataclasses.dataclass
 class ModuleOutput:
-    """For tracking outputs of forward passes over varying architectures."""
+    """Output for forward passes."""
 
     output: torch.Tensor
-    hiddens: Optional[Tuple[torch.Tensor, torch.Tensor]] = None
+    hiddens: Optional[
```
bonham79, line 18: This overloading of the hiddens property strikes me as a case for just creating a new module output class.

yoyodyne/models/modules/rnn.py
```diff
@@ -304,33 +404,53 @@ def forward(
...
-        return self.decoder_input_size + self.hidden_size
-
     @property
     def name(self) -> str:
         return "hard attention LSTM"
 
 
-class ContextHardAttentionLSTMDecoder(HardAttentionLSTMDecoder):
-    """First-order HMM hard attention LSTM."""
+class ContextualHardAttentionRNNDecoder(HardAttentionRNNDecoder):
```
bonham79, line 450: Leave as Context. (the ual can create a spelling error on a late coding debug)

bonham79 (changes requested):
Generally looks good. Saving check on Transducer for the morning. Definitely need to finish that PR to avoid making life too difficult.

Two things to consider for future work:

1). May want to operate on cell level in LSTM and GRU constructions to make disrepencies easier to manage.
2). I'd like to do what we did we encoding as we did to decoders. Just create general paradigms for decoding that allows easier swapping out of the decoder architecture. Avoids some of the boilerplate here.

yoyodyne/models/modules/attention.py
kylebgorman: ‘(X)’ isn’t a singleton tuple, that’s written ‘(X,)’. Thus these parentheses are no-op.

yoyodyne/models/hard_attention.py
kylebgorman: I tried and it’s nearly impossible because GRUs don’t return cell state and LSTMs do, and you can’t just ignore it.

tbartley94: Hmm, aight. I think this is a case for a future PR where we just use GRU/LSTM cells classes then.

kylebgorman:
I agree. I absolutely couldn’t figure out how to do it though.KOn Oct 20, 2024, at 11:52 AM, tbartley94 ***@***.***> wrote:﻿
@tbartley94 commented on this pull request.



In yoyodyne/models/hard_attention.py:
> @@ -503,3 +426,189 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             help="Width of attention context "
             "(hard attention architectures only). Default: %(default)s.",
         )
+
+
+class HardAttentionGRUModel(HardAttentionRNNModel):

Hmm, aight. I think this is a case for a future PR where we just use GRU/LSTM cells classes then.

yoyodyne/defaults.py
```diff
@@ -36,6 +36,7 @@
 HIDDEN_SIZE = 512
 MAX_SOURCE_LENGTH = 128
 MAX_TARGET_LENGTH = 128
+RNN_TYPE = "lstm"
```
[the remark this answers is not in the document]
kylebgorman: good call, removed

yoyodyne/models/base.py
```diff
@@ -63,32 +63,32 @@ class BaseEncoderDecoder(lightning.LightningModule):
...
-        start_idx,
         end_idx,
-        vocab_size,
         features_vocab_size,
-        target_vocab_size,
+        pad_idx,
         source_encoder_cls,
-        eval_metrics=defaults.EVAL_METRICS,
-        features_encoder_cls=None,
+        start_idx,
```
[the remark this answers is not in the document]
kylebgorman: the former

examples/wandb_sweeps/best_hyperparameters.py
```diff
@@ -54,7 +54,9 @@ def main(args: argparse.Namespace) -> None:
     logging.basicConfig(format="%(levelname)s: %(message)s", level="INFO")
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument(
-        "--entity", required=True, help="The entity scope for the project."
+        "--entity",
```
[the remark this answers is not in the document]
kylebgorman: I don't love it but that's what W&B calls it: entity project and sweep.

yoyodyne/models/expert.py
```diff
@@ -10,7 +10,16 @@
 import abc
 import argparse
 import dataclasses
-from typing import Any, Dict, Iterable, Iterator, List, Sequence, Set, Tuple
+from typing import (
```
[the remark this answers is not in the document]
kylebgorman: I have manually merged now.

yoyodyne/models/hard_attention.py
```diff
@@ -179,51 +175,6 @@ def decode(
             all_transition_probs.append(transition_probs)
         return torch.stack(all_log_probs), torch.stack(all_transition_probs)
 
-    def decode_step(
```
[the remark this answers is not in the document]
kylebgorman: In most classes there is an abstract model (`..RNNModel`) and concrete implementations like `...GRUModel`). However you do bring up a good point: this is one case where `decode_step` has the same implementation for both, so I will recombine them.

yoyodyne/models/modules/rnn.py
kylebgorman: done

yoyodyne/models/modules/base.py
kylebgorman: I think a class called `Hiddens` is the right solution in the future, yes.

yoyodyne/models/modules/rnn.py
```diff
@@ -37,9 +34,14 @@ def __init__(
     def num_directions(self) -> int:
         return 2 if self.bidirectional else 1
 
+    def get_module(self): ...
```
bonham79, line 37: add `NotImplementedError`

yoyodyne/models/modules/rnn.py
```diff
@@ -37,9 +34,14 @@ def __init__(
...
+
+    @property
+    def name(self) -> str: ...
```
bonham79, line 40: ditto

yoyodyne/models/modules/rnn.py
```diff
@@ -49,46 +51,71 @@ def forward(self, source: data.PaddedTensor) -> base.ModuleOutput:
...
+        return base.ModuleOutput(encoded, hiddens)
 
-    def get_module(self) -> nn.LSTM:
-        return nn.LSTM(
+    @property
+    def output_size(self) -> int:
+        return self.hidden_size * self.num_directions
+
+
+class GRUEncoder(RNNEncoder):
```
bonham79, line 79: Very minor, but if we wanted to simplify, the GRU and LSTM Encoder can just be an RNN encoder with a data factory call to nn.X. Bit on the fence if that's good for abstraction or just trying to be annoyingly compact.

bonham79 (changes requested):
Minor comments. Rest looks good.

One thing: I noticed in your module implementation that (unless I missed somehting), all the LSTM/GRU boiler plate there is the same. Could condense it by definiting a data-factory in the RNN module and then track the rnn type with a property flag. Not a major change atm but something to note.

yoyodyne/models/hard_attention.py
```diff
@@ -466,6 +384,11 @@ def _loss(
         loss = -torch.logsumexp(fwd, dim=-1).mean() / target.shape[1]
         return loss
 
+    def get_decoder(self): ...
```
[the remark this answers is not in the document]
kylebgorman: I did some research and `...` and raising `NotImplementedError` do slightly different things, and in this case I slightly prefer the latter (and it's consistent with what we did in the beam search implementation, for instance) so I will take your advice.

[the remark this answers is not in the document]
bonham79: Yeah I looked through the forums for different views and apparently the ellipsis is viewed as a bit 'cute'  (their words, not mine) among the Python community. I think silent fall throughs are counter to the other cases throughout the library as well.

kylebgorman:
I wanna take the pulse on this: what do you say I submit this and then I can figure out DRYer simplifications once I have something working in place?

bonham79:
> I wanna take the pulse on this: what do you say I submit this and then I can figure out DRYer simplifications once I have something working in place?

All my comments are minor or points of improvement. So fine if you want to just merge it now to dry run some stuff. We'll condence over future prs

kylebgorman:
For the record, switching to `NotImplementedError` uncovered a minor bug, which I then fixed. Otherwise the hidden parameters  would have gone uninitialized. So let that be a lesson to us.

PR #257, opened 2024-10-15
Added beam search for LSTM

CarlosGolvano:
This adds the beam search algorithm for the LSTM model.

- The following arguments have been added with default values of 1: `beam_width` and `n`.
- The LSTM beam search now operates with the log-likelihood of the logits.
- The parameter `return_confidences` is no longer necessary as the log-likelihoods are always returned, eliminating the polymorphic return type.
- BaseModel now includes a function interface `beam_decode` that raises a `NotImplementedError`.
- Transformer, pointer generator, and transducer now call `beam_decode`, which will raise the previous exception.
- `predict.py` now displays the beam search results. It prints all the predictions and scores for an input in one line, separating each prediction by tabs. The format is as follows:
```
prediction_00/score_00\tprediction_01/score_01\t...\tprediction_0n/score_0n
prediction_10/score_10\tprediction_11/score_11\t...\tprediction_1n/score_1n
```

### Issue
- The return of `BaseEncoderDecoder.predict_step()` is now polymorphic. Beam search returns two tensors (predictions and scores), while greedy search only returns one (predictions). The decision was made to retain it as polymorphic because the trainer and validation functions expect just one tensor and not a tuple of two tensors.


README.md
```diff
@@ -115,8 +115,14 @@ One must specify the following required arguments:
...
 
 The `--predict` file can either be a TSV file or an ordinary TXT file with one
-source string per line; in the latter case, specify `--target_col 0`. Run
-[`yoyodyne-predict --help`](yoyodyne/predict.py) for more information.
+source string per line; in the latter case, specify `--target_col 0`.
+
+Beam search is implemented (currently only in LSTM) and can be used via the following arguments:
+
+-   `--beam_width`: beam width. Allows beam search for beam_width > 1.
+-   `--n`: number of returned hypotheses.
```
kylebgorman, line 123: What would you think about making this a more informative flag name?

Alternatively, I for one would be okay if whatever beam width you selected is how much it returned also, given how easy it would be to say, set `--beam_width 5` and then ignore the latter 4 values if I don't care about them. This would mean one less flag, which seems like a win to me.

README.md
```diff
@@ -115,8 +115,14 @@ One must specify the following required arguments:
...
+
+The log-likelihood for each hypothese is always returned. Run [`yoyodyne-predict --help`](yoyodyne/predict.py) for more information.
```
kylebgorman, line 125: typo: should be `hypothesis`.

yoyodyne/models/base.py
```diff
@@ -295,12 +323,21 @@ def predict_step(
...
+            the argmax at each timestep. Position 1 are the scores for each 
+            history in beam search. It will be None when using greedy.
+
         """
         predictions = self(batch)
-        # -> B x seq_len x 1.
-        greedy_predictions = self._get_predicted(predictions)
-        return greedy_predictions
+
+        if self.beam_width is not None and self.beam_width > 1:
```
kylebgorman, line 333: How would we get a null (`None`) beam width value in the first place? It has a sensible default (!) so it makes sense to keep it integral. Then you can get rid of the `is not None and` part here.

yoyodyne/models/base.py
```diff
@@ -295,12 +323,21 @@ def predict_step(
...
+            Tuple(torch.Tensor, torch.Tensor): position 0 are the indices of 
+            the argmax at each timestep. Position 1 are the scores for each 
+            history in beam search. It will be None when using greedy.
+
         """
         predictions = self(batch)
-        # -> B x seq_len x 1.
-        greedy_predictions = self._get_predicted(predictions)
-        return greedy_predictions
+
```
kylebgorman, line 332: Pet peeve of mine but I hate blank lines within functions; they make it hard for me to tell where a function starts or ends. Add a comment or break it into smaller functions if you feel the need. A bunch were added here, can you remove them?

yoyodyne/models/base.py
```diff
@@ -295,12 +323,21 @@ def predict_step(
...
+        if self.beam_width is not None and self.beam_width > 1:
+            # For beam seach the output of the model is
+            # Tuple(predictions, scores).
+            return predictions[0], predictions[1]
```
kylebgorman, line 336: The typing at (your) line 316 has to change to accomodate this.

yoyodyne/models/base.py
```diff
@@ -295,12 +323,21 @@ def predict_step(
...
+        else:
+            # -> B x seq_len x 1.
+            greedy_predictions = self._get_predicted(predictions)
+            return greedy_predictions, None
```
kylebgorman, line 340: ditto

yoyodyne/models/base.py
```diff
@@ -476,3 +513,17 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
             help="Dimensionality of the hidden layer(s). "
             "Default: %(default)s.",
         )
+
```
kylebgorman, line 516: I assume you don't intend this to be commented out?

kylebgorman (commented):
Thank you so much for this, this is great.

* I have one style complaint; I admit it's a personal preference...
* I see a few doc typos
* I see some dead code commented out; please remove
* I have some suggestions about the interface in `yoyodyne-predict`

@Adamits may also want to weigh in on this. 

I have one other general question: would it be helpful to build a class called `Beam` that encapsulates the the heapq calls and the list that is acting as the heap?

yoyodyne/models/pointer_generator.py
```diff
@@ -427,16 +433,28 @@ def forward(
...
-                batch.source.mask,
-                batch.source.padded,
-                last_hiddens,
-                self.teacher_forcing if self.training else False,
-                features_enc=features_encoded,
-                features_mask=batch.features.mask,
-                target=batch.target.padded if batch.target else None,
-            )
+            if self.beam_width is not None and self.beam_width > 1:
+                # predictions, scores = self.beam_decode(
```
kylebgorman, line 437: Could we just let them call it and get the `NotImplementedError` from the base model? It seems harmless to do so rather than repeat ourselves on lines 444-446.

Ditto for the transducer and transformer.

Adamits:
Awesome!! I am going to take a look in about 30 mins.

yoyodyne/models/base.py
```diff
@@ -42,6 +42,7 @@ class BaseEncoderDecoder(lightning.LightningModule):
     teacher_forcing: bool
     # Decoding arguments.
     beam_width: int
+    n: int
```
Adamits, line 45: If we *are* going to keep this arg, we should at least change it to something like `beam_n` as `n` could refer to other decoding options otherwise.

yoyodyne/models/lstm.py
```diff
@@ -166,38 +166,24 @@ def beam_decode(
...
-        return_confidences: bool = False,
-    ) -> Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor]:
-        """Beam search with beam_width.
+        n: int = defaults.N,
+    ) -> Tuple[torch.Tensor, torch.Tensor]:
+        """Beam search with beam width.
 
         Note that we assume batch size is 1.
-
-        Args:
```
Adamits, line 176: It is unclear to me why we would remove the method signature here. Is this typical when inheriting a method from an interface?

yoyodyne/models/lstm.py
```diff
@@ -243,31 +229,37 @@ def beam_decode(
             # Constrains the next step to beamsize.
             for (
                 predictions,
```
Adamits, line 231: This was a poor choice by me. Can we maintain the name `logits` instead of changing it to `predictions` here? Sorry for the drive-by.

yoyodyne/models/lstm.py
```diff
@@ -243,31 +229,37 @@ def beam_decode(
...
                 # to 1. We squeeze off the first 2 dimensions to get a tensor
                 # of target_vocab_size.
                 predictions = predictions.squeeze(0).squeeze(0)
-                for j, prob in enumerate(predictions):
-                    if return_confidences:
-                        cl = char_likelihoods + [prob]
-                    else:
-                        cl = char_likelihoods
+
+                # Obtain the log-probabilities of the predictions (logits).
```
Adamits, line 242: "predictions (logits)" --> "logits"

yoyodyne/models/lstm.py
```diff
@@ -280,13 +272,19 @@ def beam_decode(
                 break
         # Returns the top-n hypotheses.
         histories = heapq.nlargest(n, hypotheses)
-        predictions = torch.tensor([h[1] for h in histories], self.device)
+
+        # Sometimes path lenghts does not match so it is neccesary to pad it
+        # all to same lenght to create a tensor.
+        max_len = max(len(h[1]) for h in histories)
+        predictions = torch.tensor(
+            [h[1] + [special.PAD_IDX] * (max_len - len(h[1])) for h in histories], device=self.device)
```
Adamits, line 280: I think we normally put the closing parenthesis on a new line here -- @kylebgorman you are more the code formatting person though :).

yoyodyne/models/lstm.py
```diff
@@ -280,13 +272,19 @@ def beam_decode(
...
+
         # Converts shape to that of `decode`: seq_len x B x target_vocab_size.
         predictions = predictions.unsqueeze(0).transpose(0, 2)
-        if return_confidences:
-            return predictions, torch.tensor([h[2] for h in histories])
-        else:
-            return predictions
+
+        # beam_serch now returns the log likelihood of each history
```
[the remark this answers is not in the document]
Adamits: I prefer if we remove the "rather" and the comment can just explain what is being returned as if this is the first implementation.

yoyodyne/models/lstm.py
```diff
@@ -298,26 +296,39 @@ def forward(
             batch (data.PaddedBatch).
 
         Returns:
-            torch.Tensor: tensor of predictions of shape seq_len x
-                batch_size x target_vocab_size.
+            Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor]: beam
+                search return a tuple with a tensor of predictions of shape
```
Adamits, line 300: return -> returns

yoyodyne/models/lstm.py
```diff
@@ -298,26 +296,39 @@ def forward(
             batch (data.PaddedBatch).
 
         Returns:
-            torch.Tensor: tensor of predictions of shape seq_len x
-                batch_size x target_vocab_size.
+            Union[Tuple[torch.Tensor, torch.Tensor], torch.Tensor]: beam
+                search return a tuple with a tensor of predictions of shape 
+                n x seq_len and tensor with the score for each prediction.
```
Adamits, line 301: Maybe replace "score" here with something more precise. Since it is common to normalize these scores by sequence length, we could say: "the unnormalized sum of symbol log-probabilities" or something like that?

yoyodyne/models/pointer_generator.py
```diff
@@ -401,10 +401,16 @@ def forward(
...
+                #     encoder_out=source_encoded,
+                #     mask=batch.source.mask,
+                #     beam_width=self.beam_width,
+                #     n=self.n
                 # )
-                raise NotImplementedError
+
+                # LSTM beam search does not work with pointer generator LSTM.
+                raise NotImplementedError(
+                    f"Beam search not implemented for {self.name} model.")
```
Adamits, line 413: Same formatting note about close paren.

Also, don't we inherit this from from the base class?

Adamits (changes requested):
I left a few nitpicks too, but looks good.

Please let us know how this performs in practice. I had wondered if using torch.argmax instead of keeping a heap would just be better in general, since it might allow us to skip some python loops and tensor methods are very fast.

If we make this change in the future, I think it will make batch_size > 1 beam_search possible.

An additional note: I don't think generalizing this to the other architectures (besides possible Transducer) will be that difficult, but this will be painfully slow for transformers without KV-caching.

kylebgorman:
> > I have one other general question: would it be helpful to build a class called Beam that encapsulates the the heapq calls and the list that is acting as the heap?
> 
> We should certainly do this eventually, and probably add some custom named tuples or dataclasses with it for tracking histories, etc.

@CarlosGolvano I for one would love this but am not expecting you to do it in this PR. This PR might help us think about what the interface should be though.

yoyodyne/models/pointer_generator.py
kylebgorman: I made a similar comment somewhere else; we might as well just call `self.beam_decode` (that way you document what its arguments ought to be) and then let the base class throw the exception.

yoyodyne/models/lstm.py
kylebgorman: For this, at the top-level directory:

```
pip install -r requirements.txt
black -l79 .
```

yoyodyne/models/pointer_generator.py
CarlosGolvano: The transformer and transducer call the `beam_decode` function of the base model. However, the pointer generator LSTM calls the LSTM `beam_decode`, resulting in a failure. Therefore, I had to comment on the call and put there the NotImplementedError.

kylebgorman: Oh, that makes then that pointer-generator has to override it. Thanks for clarification, resolving.

yoyodyne/predict.py
```diff
@@ -67,7 +67,15 @@ def get_model_from_argparse_args(
         models.BaseEncoderDecoder.
     """
     model_cls = models.get_model_cls_from_argparse_args(args)
-    return model_cls.load_from_checkpoint(args.checkpoint)
+
```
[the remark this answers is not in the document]
CarlosGolvano: I noticed that the beam_width is defined in the hparams file, and it overrides the beam_width parameter from the arguments. With Lightning, it's possible to override the hparams using the **kwargs.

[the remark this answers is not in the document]
kylebgorman: We're going to migrate to Lightning 2 (and LightningCLI) as soon as we close out all the outstanding PRs, so no huge worries about flag stuff...

yoyodyne/predict.py
```diff
@@ -104,24 +106,24 @@ def predict(
...
-    with open(output, "w", encoding=defaults.ENCODING) as sink:
-        for batch in trainer.predict(model, loader):
-            # batch -> (predictions, scores)
-            predictions = util.pad_tensor_after_eos(batch[0])
-            if batch[1] is None:
+    if model.beam_width > 1:
+        # Beam search
+        with open(output, "w", encoding=defaults.ENCODING) as sink:
+            tsv_writer = csv.writer(sink, delimiter='\t')
+            for batch in trainer.predict(model, loader):
```
kylebgorman, line 113: Could you please combine this and following line:

```python
for predictions, score in trainer.predict(model, loader):
    ...
```

Ditto on line 122-123.

yoyodyne/models/lstm.py
CarlosGolvano: I had the same question so I decided to follow the [PEP 257](https://peps.python.org/pep-0257/). 

> If a class subclasses another class and its behavior is mostly inherited from that class, its docstring should mention this and summarize the differences. Use the verb “override” to indicate that a subclass method replaces a superclass method and does not call the superclass method; use the verb “extend” to indicate that a subclass method calls the superclass method (in addition to its own behavior).

Anyway, I added some more information to the docstring.

yoyodyne/predict.py
```diff
@@ -133,8 +154,22 @@ def add_argparse_args(parser: argparse.ArgumentParser) -> None:
         required=True,
         help="Path to prediction output data TSV.",
     )
+
     # Prediction arguments.
-    # TODO: add --beam_width.
+    parser.add_argument(
```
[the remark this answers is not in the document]
CarlosGolvano: I think it is a good idea but finally I made `base.BaseEncoderDecoder.add_predict_argparse_args` in order not to break training related code.

CarlosGolvano:
I've made all the modifications you suggested in your comments—thank you for that! I'm still a bit unsure if I should address each comment individually since I'm new to PRs.

Regarding the performance of the beam search, it was really slow before I removed the 'n' parameter. Now, I don't know why, it seems a bit faster. We're planning to do some experiments these days so we can tell you more performance information. I also tried to swap out `heapq` for torch tensors, but it turned out to be trickier than I expected, so I decided to pause. I think that the Beam class you proposed will make this easier to implement.

yoyodyne/predict.py
```diff
@@ -100,9 +108,22 @@ def predict(
...
+                    print(prediction, file=sink)
+                    print(prediction)
+            else:
+                decoded_predictions = loader.dataset.decode_target(predictions)
+                for i, (prediction, score) in enumerate(zip(decoded_predictions, batch[1])):
+                    # print(prediction, file=sink)
+                    if i != 0:
+                        print('\t', end="", file=sink)
+                        print('\t', end="")
+                    print(f"{prediction}/{score}", end="", file=sink)
```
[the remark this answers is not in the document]
CarlosGolvano: I implemented the "collate" option. Now the output is:
`prediction_1\tscore_1\t...\tprediction_n\tscore_n`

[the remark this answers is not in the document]
kylebgorman: Great, resolving.

kylebgorman:
> I've made all the modifications you suggested in your comments—thank you for that! I'm still a bit unsure if I should address each comment individually since I'm new to PRs.

I have been doing this as you revise but it helps me if you click the “resolve” button once you commit a fix for an issue.

Broadly this looks good to me. I would like to do one more pass and test locally in the next 24 hours so I can confirm things.

@Adamits can you also check this one last time?

yoyodyne/models/lstm.py
```diff
@@ -280,13 +272,19 @@ def beam_decode(
                 break
         # Returns the top-n hypotheses.
         histories = heapq.nlargest(n, hypotheses)
-        predictions = torch.tensor([h[1] for h in histories], self.device)
+
+        # Sometimes path lenghts does not match so it is neccesary to pad it
```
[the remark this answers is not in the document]
CarlosGolvano: I noticed that sometimes hitories has different lenghts, so it is not possible to create a `torch.Tensor` from these predictions. Therefore, I had to pad all the predictions to match lengths.

kylebgorman:
I'll be AWK for a few days and if @Adamits hasn't had a chance to look at this by then I'll just test locally to make sure everything's good to go. Sorry for the delay.

Adamits:
Taking a look. Just having a few dependency issues since I haven't done a fresh yoyodyne install on my mac in a while :D.

Adamits:
> Regarding the performance of the beam search, it was really slow before I removed the 'n' parameter. Now, I don't know why, it seems a bit faster. We're planning to do some experiments these days so we can tell you more performance information.

Ah ok, I guess this skips some heapq.top_k calls, which may have been a bottleneck?

I also tried to swap out heapq for torch tensors, but it turned out 

> to be trickier than I expected, so I decided to pause. I think that the Beam class you proposed will make this easier to implement.

Cool! Let me know if you want to discuss sometime.

Adamits:
Sorry for delays. I have tested this for `attentive_lstm` and it seems to work as expected. One thing i'd like Kyle to weigh in on though---by default we write the predictions **and** scores when beam_width > 1, but if beam_width == 1, then we just write the predictions.

I don't nec. have a problem with this but it is a somewhat weird behavior. I think modifying what you do and do not write during prediction is easy for a user to change in python, and avoiding more CLI flags is probably a good move, but wanted to point this out anyway.

Adamits (approved):
LGTM. Approving pending Kyle's response.

CarlosGolvano:
Sorry for the inconvenience, I just made a last commit updating the README. I forgot to remove the 'n' parameter from the beam search and also added the beam search output format.

Thank you very much for your work and for taking time to accept the PR.  :blush:

kylebgorman:
> Sorry for delays. I have tested this for `attentive_lstm` and it seems to work as expected. One thing i'd like Kyle to weigh in on though---by default we write the predictions **and** scores when beam_width > 1, but if beam_width == 1, then we just write the predictions.

I'm okay with this state of affairs.

kylebgorman:
OK, just got the chance to test locally and it works well. I'll merge now.

We will put out version 0.2.13 now and it'll include this.

We'd love to

1) increase support for this in other models
2) generalize to batch sizes > 1, and/or
3) incorporate into training

in the future, and you @CarlosGolvano are welcome to contribute if you're so inclined.

PR #264, opened 2024-10-31
Removes `dtype=torch.long` declarations

### LINKED ISSUE #236: Casting to long
[Here](yoyodyne/data/datasets.py#L79) we cast encoded tensors to the `Long` dtype, i.e., int64. Why? Surely int32 would be enough. Is there any particular reason for this? Could we save memory by getting rid of the dtype spec?

kylebgorman:
I know of no reason to think these are necessary.

I also simplify the creation of the pointer distribution tensor. Since this is made with `zeros` one can just have dimension 1 be size 1 at creation; the `unsqueeze` call is unnecessary.

Closes #236.

### DIFF
--- a/yoyodyne/data/datasets.py
+++ b/yoyodyne/data/datasets.py
@@ -73,10 +73,7 @@ class Dataset(data.Dataset):
         Returns:
             torch.Tensor: the encoded tensor.
         """
-        return torch.tensor(
-            [self.index(symbol) for symbol in symbols],
-            dtype=torch.long,
-        )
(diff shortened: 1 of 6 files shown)


kylebgorman:
I have tested this on all architectures on both CPU and a CUDA GPU, and also spot-checked with different precisions on GPU, and everything is working as expected.

kylebgorman:
Hi @Adamits, gentle ping here.

Adamits:
Sorry! I had looked at this when it was still a draft, and forgot to come back. Checking now.

PR #265, opened 2024-11-20
Migrates to LightningCLI

### LINKED ISSUE #60: Torch and Lightning 2.0.0

### LINKED ISSUE #218: Unpin Numpy
#202 pinned numpy to < 2 because there was a conflict with PyTorch. This serves a reminder to remove this pin: most likely we can set numpy > 2.x.y (tbd) and also do the same for torch > 2.a.b. (also tbd) once the conflict is worked out. In fact this might already be possible.

kylebgorman:
Closes #60.
Closes #218.

LightningCLI removes our need to create separate training and prediction CLI programs, moving nearly all of that logic into the base model.

This commit in particular sets the stage:

* Updates dependencies.
* Increments minor version number.
* Creates an empty `cli.py` where the CLI-speific logic will live.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,9 +161,9 @@ provide any symbols of the form `<...>`, `[...]`, or `{...}`.
 Checkpointing is handled by
 [Lightning](https://pytorch-lightning.readthedocs.io/en/stable/common/checkpointing_basic.html).
 The path for model information, including checkpoints, is specified by
-`--model_dir` such that we build the path `model_dir/version_n`,
-where each run of an experiment with the same `model_dir` is namespaced with a
-new version number. A version stores all of the following:
+`--model_dir` such that we build the path `model_dir/version_n`, where each run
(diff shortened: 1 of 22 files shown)


kylebgorman:
Closing in favor of #327.

PR #266, opened 2024-11-25
Moves to mapper interface

### LINKED ISSUE #137: Move tensor encoding/decoding into the Index class
This is a notice of my plans to move the encoding methods (which take strings and make tensors) and decoding methods (which convert tensors back into strings) into the Index class itself. Currently this ability lives inside of the `DataSet` class and I really don't know why it belongs there. This should substantially reduce weird coupling, like for instance what happens [here](https://github.com/CUNY-CL/yoyodyne/blob/3d90fc5b71ca4b59df41e948bfde2219210bf2dd/yoyodyne/predict.py#L99) with all those chained method calls. 

As part of this, the Index class will have two class methods, one which reads the index from a `.pkl` file, and one which constructs the index from a list of file handles. This latter should give me a place to compute max tensor sizes, as needed for #50. 

I also think the index will then become a better object to unit test.

kylebgorman:
Borrowing a design element I used in UDTube, I decompose the dataset object into two pieces:

* a `Mapper` interface which knows how to map between lists of strings and tensors (to decode and encode)
* `DataSet`, as before.

There was no particular reason for the mapper functions to live inside the dataset. This separates the two pieces and uses the mapper object for prediction.

Closes #137. That issue says that the encoding/decoding should be moved to the index, but this actually makes those two even more modular.

(I also imported fix #272 and resolved some merge stuff from #268 etc.)

### DIFF
--- a/yoyodyne/predict.py
+++ b/yoyodyne/predict.py
@@ -103,7 +103,7 @@ def predict(
             # Beam search.
             tsv_writer = csv.writer(sink, delimiter="\t")
             for predictions, scores in trainer.predict(model, loader):
-                predictions = util.pad_tensor_after_eos(predictions)
+                predictions = util.pad_tensor_after_end(predictions)
                 # TODO: beam search requires singleton batches and this
                 # assumes that. Revise if that restriction is ever lifted.
                 targets = [
--- a/yoyodyne/predict.py
+++ b/yoyodyne/predict.py
@@ -27,7 +27,7 @@ def get_trainer_from_argparse_args(
 def get_datamodule_from_argparse_args(
     args: argparse.Namespace,
 ) -> data.DataModule:
-    """Creates the dataset from CLI arguments.
+    """Creates the datamoodule from CLI arguments.
 
     Args:
         args (argparse.Namespace).
--- a/yoyodyne/predict.py
+++ b/yoyodyne/predict.py
@@ -36,12 +36,16 @@ def get_datamodule_from_argparse_args(
         data.DataModule.
     """
     separate_features = args.features_col != 0 and args.arch in [
-        "pointer_generator_rnn",
+        "hard_attention_gru",
+        "hard_attention_lstm",
+        "pointer_generator_gru",
+        "pointer_generator_lstm",
         "pointer_generator_transformer",
-        "transducer",
+        "transducer_grm",
+        "transducer_lstm",
     ]
-    index = data.Index.read(args.model_dir)
     return data.DataModule(
+        model_dir=args.model_dir,
         predict=args.predict,
         batch_size=args.batch_size,
         source_col=args.source_col,
@@ -53,7 +57,6 @@ def get_datamodule_from_argparse_args(
         separate_features=separate_features,
         max_source_length=args.max_source_length,
         max_target_length=args.max_target_length,
-        index=index,
     )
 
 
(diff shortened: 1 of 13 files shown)


yoyodyne/predict.py
```diff
@@ -104,23 +96,43 @@ def predict(
...
-                decoded_predictions = loader.dataset.decode_target(predictions)
-                row = itertools.chain(
-                    *zip(decoded_predictions, scores.tolist())
+                # TODO: beam search requires singleton batches and this
+                # assumes that. Revise if that restriction is ever lifted.
+                targets = [
+                    parser.target_string(mapper.decode_target(target))
+                    for target in predictions
+                ]
+                row = itertools.chain.from_iterable(
```
Adamits, line 113: Could you add a comment here?

yoyodyne/predict.py
```diff
@@ -104,23 +96,43 @@ def predict(
...
             tsv_writer = csv.writer(sink, delimiter="\t")
             for predictions, scores in trainer.predict(model, loader):
                 predictions = util.pad_tensor_after_eos(predictions)
-                decoded_predictions = loader.dataset.decode_target(predictions)
-                row = itertools.chain(
-                    *zip(decoded_predictions, scores.tolist())
+                # TODO: beam search requires singleton batches and this
+                # assumes that. Revise if that restriction is ever lifted.
+                targets = [
+                    parser.target_string(mapper.decode_target(target))
```
Adamits, line 110: Just so I am clear, `mapper` goes from ints to strings, and `target_string` filters pads, etc and implements proper spacing? The method `target_string` is not very intuitive for me here.

Adamits (commented):
I don't really have a problem with it, but I don't actually follow the reasoning for needing another class. The "mapping" operations, to me, could intuitively live in the index, the dataset, or the datamodule. Is the ambiguity there around where it should live actually the problem? Or is this just an OOP principle that someone established?

yoyodyne/predict.py
kylebgorman: Done. 

```python
# Collates target strings and their scores.
```

yoyodyne/predict.py
kylebgorman: Yeah, it knows about the separators. That predates this PR.

kylebgorman:
> I don't really have a problem with it, but I don't actually follow the reasoning for needing another class. The "mapping" operations, to me, could intuitively live in the index, the dataset, or the datamodule. Is the ambiguity there around where it should live actually the problem? Or is this just an OOP principle that someone established?

The way to think of the design is this:

* The datamodule creates an index.
* The datamodule creates datasets.
* At the creation of each dataset, a mapper is made from the index and passed to the dataset. (This is necessary because the dataset generates tensors on the fly.)

I am moving it out of the dataset in this PR, because there are a number of places where you want to do what the mapping does but you don't need a reference to the full dataset. (For instance, you don't need the huge list that contains the actual data.) One example of this is in prediction: you need to map but there's no direct reference to the dataset. Putting it in a separate class helps with potential circularity issues.

The other obvious option is to make it *part* of the index but there are places where we need an index but not a mapper or vice versa. For instance the expert class doesn't need tensors or any of the padding, so it uses the index but not the mapper. I despaired of a way to separate out "string to integer" vs. "integers to tensors" except putting them in separate classes.

The general OOP principle at play here is the [Law of Demeter](https://en.wikipedia.org/wiki/Law_of_Demeter). In the old code (this was part of `predict.py`) had `loader.dataset.decode_target(...)`; this now reads `mapper.decode_target(target)`.

Adamits:
Thanks for describing your thinking.

> there are a number of places where you want to do what the mapping does but you don't need a reference to the full dataset

Yes good point. I think another place this is helpful is for runtime debugging. previously if I wanted to log encoded inputs/predicted outputs in the forward function at runtime, I needed the dataset.

Adamits (approved):
LGTM

PR #268, opened 2024-11-25
Removes separate "experiment" subdirectory

kylebgorman:
This adds an extra flag to every command and for what? If you want to keep model directories separate for different experiments in a way that goes beyond already done with the automated versioning, you can just append a subdirectory name to `--model_dir`, e.g., using `--model_dir models/foo` instead of `--model_dir models --experiment foo`.

The README is updated to reflect.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,9 +161,9 @@ provide any symbols of the form `<...>`, `[...]`, or `{...}`.
 Checkpointing is handled by
 [Lightning](https://pytorch-lightning.readthedocs.io/en/stable/common/checkpointing_basic.html).
 The path for model information, including checkpoints, is specified by
-`--model_dir` such that we build the path `model_dir/version_n`,
-where each run of an experiment with the same `model_dir` is namespaced with a
-new version number. A version stores all of the following:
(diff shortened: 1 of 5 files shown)


Adamits (approved):
Totally agree with this change and the code LGTM.

BUT I think this removes the ability to specify the wandb project to log to.  This was a confusing name for it anyway, and itd be good to also have an option for the wandb run_name, so probably a separate issue, just wanted to point it out.

kylebgorman:
> BUT I think this removes the ability to specify the wandb project to log to. This was a confusing name for it anyway, and itd be good to also have an option for the wandb run_name, so probably a separate issue, just wanted to point it out.

I'll test the W&B thing and fix it if so.

kylebgorman:
So there is an alternative way to set the project which is using the environment variable `$WANDB_PROJECT`. What if we just documented that and went ahead with the change?

kylebgorman:
I am going to merge this having added a note in the docs about the environmental variable. This sort of thing will become irrelevant with LightningCLI too (and I'm working on that now: the last week's PRs are all related to that migration).

PR #269, opened 2024-11-25
Use future annotations

kylebgorman:
There is just one location where this is useful.

For context, see [PEP 563](https://peps.python.org/pep-0563/). 

Note that this is backwards compatible all the way back to Python 3.7 (which we don't even support) and thus harmless.


kylebgorman:
This is a very minor change, no-op in fact, so I'm going to merge.

PR #272, opened 2024-11-30
Expands list of separate features models

kylebgorman:
These were missing.


kylebgorman:
This is safe so I'm going to push it in.

PR #273, opened 2024-11-30
Fixes prediction across all architectures

kylebgorman:
Two bugs found by testing prediction across all architectures:

* Hard attention models inherit the beam search implementation in `RNNModel`, but it's incompatible so it needs to be disabled. Failing to do this results in an inscrutable error instead of a `NotImplementedError`. This is fixed.
* The base model implementation of the prediction loop is made to have what seems like the obvious polymorphic return type which restores compatibility for hard attention. I suspect this is a minor regression introduced in #257.

Clean-ups done at the same time:

* Standardizes the name of the special symbol: it's called EOS at various points in the documentation but the code actually calls it `END` and its tag is `<E>`.
* Uses the names `beam_decode` and `greedy_decode` everywhere.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,7 +107,6 @@ One must specify the following required arguments:
 
 -   `--arch`: architecture, matching the one used for training
 -   `--model_dir`: path for model metadata
--   `--experiment`: name of experiment
 -   `--checkpoint`: path to checkpoint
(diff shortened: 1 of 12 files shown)


Adamits (approved):
Left one nit, but LGTM. 

Thanks for taking the time to do this, you're doing the lord's work.

PR #274, opened 2024-12-01
Cleanup to index writing

kylebgorman:
* Because the index is now all compromised of what `pickle` considers to be POD types (previously it consisted of separate `Vocabulary` objects), we can just load it directly without `setattr` now.
* To support this (and a few other things) we add a utility called `mkpath` which creates all the subdirectories needed and use this in all the cases where we are writing to a user-specified file.
* For simplicity we let the datamodule write the index. Less logic in `train` or `predict`: easier migration to LightningCLI.
* Gets rid of some unused stuff in the index.

### DIFF
--- a/yoyodyne/data/indexes.py
+++ b/yoyodyne/data/indexes.py
@@ -149,6 +149,10 @@ class Index:
             return len(self.SPECIAL) + len(self.source_vocabulary)
 
     @property
+    def features_vocab_size(self) -> int:
+        return len(self.features_vocabulary) if self.features_vocabulary else 0
+
+    @property
     def target_vocab_size(self) -> int:
         if self.tie_embeddings:
             return self.vocab_size
@@ -156,7 +160,3 @@ class Index:
             return len(special.SPECIAL) + len(self.target_vocabulary)
         else:
             return 0
-
-    @property
-    def features_vocab_size(self) -> int:
-        return len(self.features_vocabulary) if self.features_vocabulary else 0
--- a/yoyodyne/data/indexes.py
+++ b/yoyodyne/data/indexes.py
@@ -2,11 +2,10 @@
 
 from __future__ import annotations
 
-import os
 import pickle
 from typing import Dict, Iterable, List, Optional
 
-from .. import defaults, special
+from .. import defaults, special, util
 
 
 class Error(Exception):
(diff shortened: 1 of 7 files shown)


yoyodyne/data/indexes.py
```diff
@@ -160,25 +160,3 @@ def target_vocab_size(self) -> int:
             return len(special.SPECIAL) + len(self.target_vocabulary)
         else:
             return 0
-
-    @property
-    def features_vocab_size(self) -> int:
-        return len(self.features_vocabulary) if self.features_vocabulary else 0
-
-    # These are also recorded in the `special` module.
```
Adamits, line 168: How do we enforce that these  always occur in the same index position for these symbols? I assume in special the index variables are set manually?

yoyodyne/util.py
```diff
@@ -24,7 +24,39 @@ def __call__(
...
+        log_info(f"\t{arg}: {val!r}")
+
+
+def mkpath(path: str) -> None:
+    """Creates subdirectories for a path if they do not already exist.
+
+    Args:
+        path (str).
+    """
+    dirname = os.path.dirname(os.path.abspath(path))
```
Adamits, line 58: Is this how we created paths for artifacts previously? If not, is there a reason to prefer abs over rel paths?

Adamits (changes requested):
I just have a few quick questions (2 are inline comments) before we merge. Notably:

> Because the index is now all compromised of what pickle considers to be POD types (previously it consisted of separate Vocabulary objects), we can just load it directly without setattr now.

I actually don't follow what this means, would you mind explaining a bit more (e.g. what is a POD type?)?

yoyodyne/util.py
kylebgorman: No, I am being more perspicuous than before.

`abspath` is the standard way to get Python to correctly expand stuff like `..` in an OS-specific way; `relpath` doesn't do any such thing.

yoyodyne/data/indexes.py
kylebgorman: This hasn't changed: they are defined in `special` in a certain order and then we insert them at the beginning of the index.

Adamits (approved):
> "Plain ole' data", here I mean things Python knows how to pickle. Python knows how to pickle an Index; it didn't use to because an Index used to itself be made out of other Python classes.

Ahhh ok, nice. LGTM then.

PR #275, opened 2024-12-01
Uses simpler padding interface

kylebgorman:
This function works just fine for what we're trying to do here.


kylebgorman:
Going to leave this here in case we ever want to go this route.

PR #277, opened 2024-12-02
One necessary `long` conversion

kylebgorman:
This reverses a line in #264.

I think this only affects CPU or MPS; error without reads:

```
  File ".../site-packages/torch/nn/functional.py", line 2551, in embedding
    return torch.embedding(weight, input, padding_idx, scale_grad_by_freq, sparse)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: Expected tensor for argument #1 'indices' to have one of the following scalar types: Long, Int; but got CPUBoolType instead (while checking arguments for embedding)
```

### DIFF
--- a/yoyodyne/models/modules/transformer.py
+++ b/yoyodyne/models/modules/transformer.py
@@ -259,8 +259,11 @@ class FeatureInvariantTransformerEncoder(TransformerEncoder):
             embedded (torch.Tensor): embedded tensor of shape
                 B x seq_len x embed_dim.
         """
-        # Distinguishes features and chars; 1 or 0.
-        char_mask = symbols < (self.num_embeddings - self.features_vocab_size)
+        # Distinguishes features and chars; 1 or 0; embedding layer requires
+        # this to be integral.
+        char_mask = (
+            symbols < (self.num_embeddings - self.features_vocab_size)
+        ).long()
         type_embedding = self.esq * self.type_embedding(char_mask)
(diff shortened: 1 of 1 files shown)


Adamits (commented):
Good catch! But does `.long()` make it an `int64`? Can we just make it 8-bit? there might be a `tensor.int()` or similar? Or  `tensor.to(torch.int8)`

Doesn't matter a lot, but in the name of getting rid of unnecessary longs :)...

Adamits:
I did a quick test:

```
x = >>> x = torch.Tensor([True, False, True, False])
>>> x
--> tensor([1., 0., 1., 0.])
>>> x.long()
--> tensor([1, 0, 1, 0])
>>> x.long().dtype
--> torch.int64
>>> x.int().dtype
--> torch.int32
>>> x.to(torch.int8).dtype
--> torch.int8

kylebgorman:
The accelerators have an opinion about `int32` vs `int64` IIRC. I forget what they want though.

kylebgorman:
Torch also has lower-precision ones (8- and 16-bit) but most of them don't interoperate with CUDA well so I try not to get too hacky here.

Adamits:
> The accelerators have an opinion about `int32` vs `int64` IIRC. I forget what they want though.

Maybe we just keep it long to be safe then, this is extremely low impact.

PR #278, opened 2024-12-05
Removes unused char-likelihoods in beam search

kylebgorman:
The beam search implementation tracks, in addition to the (incomplete) log-likelihood of each candidate, a list of likelihoods for each character. However, this quantity is never used for anything and removing it does not change operation except using slightly less memory.


kylebgorman:
Since this is safe I'll put it in. Working to generalize beam search elsewhere.

PR #279, opened 2024-12-06
Adds some missing bibliography.

kylebgorman:
Also corrects a typo.


kylebgorman:
Submitting over the top of that fastapi/pydantic nonsense since this is quite safe.

PR #281, opened 2024-12-07
Adds a few design notes to the README

kylebgorman:


kylebgorman:
This is safe (README only) so submitting,

PR #292, opened 2024-12-11
Enforce abstraction with PEP 3119

kylebgorman:
PEP 3119 introduced decorators for declaring abstract methods. This PR uses them throughout the model, module, and evaluator hierarchies.

This is definitely superior to what we did before #254, which was to just declare but not implement these. Since that's the same as having a no-op null-return function, an unimplemented method in a concrete class just returns `None`. Usually, this will trigger failures somewhere downstream at run-time but the errors are completely uninformative and hard to trace down.

In #254, which added GRU support, I changed these to raise `NotImplementedError`. This means we get a fatal error (assuming we've wrapped things in an overly-general `try`/`except`) when the methods are called. 

With PEP 3119, we make `BaseModel`, `BaseModule`, etc. inherit (directly) from `abc.ABC` and then decorate methods that concrete classes must implement with `@abc.abstractmethod` (etc.). Then, Python gives an informative error when you attempt to instantiate a class which you thought was concrete but which in fact is abstract, and tells you the name of the method which hasn't been implemented. This is also "runtime" but happens at construction time and the error message is better. To review what this actually looks like, I recommend reading included commit [`ac8234f0ed0698e70adf0b86e95311156723b1f5`](https://github.com/kylebgorman/yoyodyne/commit/ac8234f0ed0698e70adf0b86e95311156723b1f5).

To actually track down the methods, I also imposed the following cleanups:

* Instead of using a static method to add args to the parser, I moved this to the module level, which cleans up things a bit. 
* I reordered methods roughly alphabetically; helper methods were made `_protected` (most already were) and are listed right after the methods(s) which call them. This is the recommendation from the "Clean Code" people and it helped me find all the missing pieces.

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,10 +408,8 @@ accelerator. Furthermore, since it is developed at "low-intensity" by a
 geographically-dispersed team, consistency is particularly important. Some
 consistency decisions made thus far:
 
--   The abstract/concrete class distinction is enforced at runtime by the
-    presence of methods raising `NotImplementedError`, though in the future we
-    may consider [PEP 3119](https://peps.python.org/pep-3119/)-style
-    "compile-time" enforcement.
+-   Abstract classes overrides are enforced using [PEP
+    3119](https://peps.python.org/pep-3119/).
 -   [`numpy`](https://numpy.org/) is used for basic mathematical operations and
     constants even in places where the built-in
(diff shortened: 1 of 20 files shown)


kylebgorman:
I've tested this heavily so submitting.

PR #293, opened 2024-12-12
Audits dropout layers

kylebgorman:
# Redundancy

Both base model and base module have a dropout layer, but the base model's dropout is unused. I remove it. We still pass the dropout probability to the modules, of course.

# In-place operation

The base module dropout is defined as an out-of-place operation, but there are several places where we do:

```python
self.dropout_layer = nn.Dropout(..., inplace=False)
...
output = ...
output = self.dropout_layer(output)
```

Knowing what I know about Python internals this is pretty expensive:

* The RHS of the final expression to be computed before the assignment, so we temporarily have two copies of `output` (one dropped out, one not) in memory.
 * This also "thrashes" the GC a bit: it has to deallocate the first one (its reference count will become zero after the reassignment) to clean up the mess.

In contrast:

```python
self.dropout_layer = nn.Dropout(..., inplace=True)
...
output = ...
self.dropout_layer(output)
```

uses half as much memory and doesn't exercise the GC either. Note that there are cases in the transformer where I now have to assign a variable to the combined embedding before dropping out, but that was happening implicitly at variable binding anyways and the new approach does not require any extra storage (plus it's mildly documentary).

### DIFF
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -108,7 +108,6 @@ class BaseModel(lightning.LightningModule):
         self.embeddings = self.init_embeddings(
             self.vocab_size, self.embedding_size
         )
-        self.dropout_layer = nn.Dropout(p=self.dropout, inplace=False)
         self.evaluators = [
             evaluators.get_evaluator(eval_metric)()
             for eval_metric in self.eval_metrics
--- a/yoyodyne/models/modules/base.py
+++ b/yoyodyne/models/modules/base.py
@@ -60,7 +60,7 @@ class BaseModule(lightning.LightningModule):
         self.num_embeddings = num_embeddings
         self.layers = layers
         self.hidden_size = hidden_size
-        self.dropout_layer = nn.Dropout(p=self.dropout, inplace=False)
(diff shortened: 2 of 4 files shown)


Adamits (approved):
Nice catch, thanks for the explanation. I was surprised to see how few times the code actually has to call the dropout layer---I might separately from this double check that the transformer decoder doesn't need an extra explicit dropout call.

LGTM

kylebgorman:
> Nice catch, thanks for the explanation. I was surprised to see how few times the code actually has to call the dropout layer---I might separately from this double check that the transformer decoder doesn't need an extra explicit dropout call.

The transformer of course has internal dropout between layers and there's also a call in the "embed" helper, which is shared between encoder and decoder modules. 

Testing suggests it still works as expected.

Adamits:
Right, I was wondering if we want something synonymous to https://github.com/CUNY-CL/yoyodyne/pull/293/files#diff-11ea8a753d8ac3315dcc44bb7543539e11d60fee367b507c02287c91b0c85412R253 though. Whether that should be there for either architecture implementation is probably a pretty small detail either way.

PR #294, opened 2024-12-13
Helper for emission probability

kylebgorman:
* Adds a helper for computing the emission probabilities.
* Gives consistent names to this and the transition probability helper.
* Uses positive rather than negative indices (which IMO are easier to debug with).


kylebgorman:
I believe this is pretty safe and have tested extensively so I am going to merge.

PR #297, opened 2025-01-14
Fixes copy-paste bug in train.

kylebgorman:
This forgot to pass `--source_col`.


kylebgorman:
This is safe and very small so I'll merge it.

PR #298, opened 2025-01-14
Reverts use of in-place dropout

kylebgorman:
This reverts an earlier change (#293) which added in-place dropout. On my PC (Windows 11, WSL, older Nvidia GPU) in-place dropout throws the inscrutable message:

```
RuntimeError: one of the variables needed for gradient computation has been modified by an inplace operation: [torch.cuda.FloatTensor [128, 1, 512]], which is output 0 of CudnnRnnBackward0, is at version 1; expected version 0 instead. Hint: enable anomaly detection to find the operation that failed to compute its gradient, with torch.autograd.set_detect_anomaly(True).
```

Tracing down this mysterious tensor of shape 128 x 1 x 512 I found that it was the tensor input to dropout.

Since I didn't see this on my laptop I suspect it is quite literally platform-dependent. 

Note however that I now instead avoid the expensive overwriting operation all the same, so we get most of the same benefits as #293.

### DIFF
--- a/yoyodyne/models/base.py
+++ b/yoyodyne/models/base.py
@@ -58,10 +58,9 @@ class BaseModel(abc.ABC, lightning.LightningModule):
     features_encoder_cls: Optional[modules.base.BaseModule]
     hidden_size: int
     source_encoder_cls: modules.base.BaseModule
-    # Constructed inside __init__.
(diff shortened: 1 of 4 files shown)


kylebgorman:
I have tested this against all models so it is a safe merge.

PR #300, opened 2025-01-17
Avoids 0-to-a-negative exponent error

kylebgorman:
When using the warmup inverse square root scheduler, if one specifies no warmup, you will hit the following error:

```
    return self.decay_factor * step**-0.5
                               ~~~~^^~~~~
ZeroDivisionError: 0.0 cannot be raised to a negative power
```

This can simply be avoided by treating step as 1-based. We already do this in the warmup clause, so we now do this in the warmed-up clause too. I also slightly improve the comments since this is an important point.

### DIFF
--- a/yoyodyne/schedulers.py
+++ b/yoyodyne/schedulers.py
@@ -68,7 +68,7 @@ class WarmupInverseSquareRoot(optim.lr_scheduler.LambdaLR):
             # +1 in numerator avoids a zero-LR first step.
             return (step + 1) / self.warmup_steps
         # +1 in base of exponent avoids an undefined operation (0 to a negative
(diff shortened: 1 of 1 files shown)


kylebgorman:
Hi @Adamits could you take a quick look at this, just to be sure?

Adamits (approved):

kylebgorman:
> So this would have triggered if you specifically requested this scheduler, but set warmup to 0? I am surprised I never had a sweep that did this, but maybe I just missed the error when it did.
> 
> Either way I think 1-based indexing makes sense here. LGTM

Yep, and it happened to me during a sweep.

PR #308, opened 2025-02-14
Robust subprocess sweep

### LINKED ISSUE #302: Even more robust sweeping
In [UDTube](https://github.com/CUNY-CL/udtube) the sweep script this loads an input YAML configuration file specified by `--config` and connects to W&B using the normal flags, then for each run, it inserts the W&B hyperparameter suggestions (overwriting if necessary) into the config and shells out to `udtube fit`. Its `sweep.py` parses just four flags: 3 specifying details for W&B, and one for the config file; the other ones are passed directly to `udtube fit` with the help of `argparse.ArgumentParser.parse_known_args`.

This design could easily be adapted for Yoyodyne and is more or less independent of LightningCLI migration. One advantage of this design is that when the shell call returns, the OS frees up all the associated memory. Right now, we try to empty the CUDA cache, but it's not completely obvious to me if this works as expected or not. Using a shell call instead ought to have a tiny bit additional overhead, but given that we're just running a single process at a time anyways, it's pretty minimal and should be more robust in terms of freeing memory.

kylebgorman:
As described in #302, this moves the sweep to a model where each sweep run is a subprocess, which seems likely to ensure the OS returns memory at the end of each run. This model is [one we piloted in UDTube](https://github.com/CUNY-CL/udtube/blob/master/examples/wandb_sweeps/sweep.py) and here the implementation is actually simpler because there's no need for temporary config files.

Testing confirms this works and that the data is being logged as expected to W&B, but I'd appreciate if a reviewer can try it out locally with an existing workflow. 

One nice side effect of this is we no longer need to maintain the `train.train` module interface, which has no place in LightningCLI anyways. 

Closes #302.

### DIFF
--- a/examples/wandb_sweeps/sweep.py
+++ b/examples/wandb_sweeps/sweep.py
@@ -1,79 +1,95 @@
 #!/usr/bin/env python
-"""Runs the sweep itself."""
+"""Runs a W&B sweep."""
 
 import argparse
 import functools
+import logging
+import subprocess
 import traceback
 import warnings
 
-import torch
+from typing import List
+
 import wandb
 
-from yoyodyne import train, util
 
 warnings.filterwarnings("ignore", ".*is a wandb run already in progress.*")
 
 
-def train_sweep(args: argparse.Namespace) -> None:
-    """Runs a single training run.
-
-    The wandb config data used here comes from the environment.
+def run(argv: List[str]) -> None:
+    """A single training run.
 
     Args:
-        args (argparse.Namespace).
+        argv (List[str]): partial command-line arguments (the command name
+            plus CLI arguments passed to this script but not declared below).
+    """
+    argv = populate_argv(argv)
+    process = subprocess.Popen(argv, stderr=subprocess.PIPE, text=True)
+    for line in process.stderr:
+        logging.info(line.rstrip())
+    wandb.finish(exit_code=process.wait())
+
+
+def populate_argv(argv: List[str]) -> List[str]:
+    """Populates argv with W&B arguments.
+
+    This copies and mutates to prevent side effects persisting beyond the
+    individual run.
+
+    Args:
+        argv (List[str]): partial command-line arguments (the command name
+            plus CLI arguments passed to this script but not declared below).
+
+    Returns:
+        List[str]: complete command-line arguments populated with W&B run
+            hyperparameters.
     """
     wandb.init()
-    # Model arguments come from the wandb sweep config and override any
-    # conflicting arguments passed via the CLI.
+    result = argv.copy()
     for key, value in wandb.config.items():
-        if key in args:
-            util.log_info(f"Overriding CLI argument: {key}")
-        setattr(args, key, value)
-    try:
-        train.train(args)  # Ignoring return value.
-    except RuntimeError as error:
-        # TODO: consider specializing this further if a solution to
-        # https://github.com/pytorch/pytorch/issues/48365 is accepted.
-        util.log_info(f"Runtime error: {error!s}")
-    finally:
-        # Clears the CUDA cache.
-        torch.cuda.empty_cache()
+        result.append(f"--{key}")
+        if value is not None:
+            result.append(f"{value}")
+    return result
 
 
-def main() -> None:
-    parser = argparse.ArgumentParser(description=__doc__)
-    train.add_argparse_args(parser)
-    parser.add_argument(
-        "--sweep_id",
-        required=True,
-        help="ID for the sweep.",
-    )
-    parser.add_argument("--entity", required=True, help="Entity name.")
-    parser.add_argument("--project", required=True, help="Project name.")
-    parser.add_argument(
-        "--count",
-        type=int,
-        help="Number of runs to perform.",
-    )
-    args = parser.parse_args()
-    # Forces log_wandb to True, so that the PTL trainer logs runtime metrics
-    # to wandb.
-    args.log_wandb = True
+def main(args: argparse.Namespace, argv: List[str]) -> None:
+    argv = ["yoyodyne-train", *argv]
     try:
         wandb.agent(
             sweep_id=args.sweep_id,
             entity=args.entity,
             project=args.project,
-            function=functools.partial(train_sweep, args),
+            function=functools.partial(run, argv),
             count=args.count,
         )
     except Exception:
         # Exits gracefully, so wandb logs the error.
-        util.log_info(traceback.format_exc())
+        logging.fatal(traceback.format_exc())
+        wandb.finish(exit_code=1)
         exit(1)
-    finally:
-        wandb.finish()
 
 
 if __name__ == "__main__":
-    main()
+    logging.basicConfig(
+        format="%(levelname)s: %(asctime)s - %(message)s",
+        datefmt="%d-%b-%y %H:%M:%S",
+        level="INFO",
+    )
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--sweep_id", required=True, help="ID for the sweep.")
+    parser.add_argument(
+        "--entity", required=True, help="The entity scope for the project."
+    )
+    parser.add_argument(
+        "--project", required=True, help="The project of the sweep."
+    )
+    parser.add_argument(
+        "--count",
+        type=int,
+        help="Number of runs to perform.",
+    )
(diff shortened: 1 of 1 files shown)


examples/wandb_sweeps/sweep.py
```diff
@@ -1,79 +1,97 @@
 #!/usr/bin/env python
-"""Runs the sweep itself."""
+"""Runs a W&B sweep."""
```
Adamits, line 2: Maybe "Trains a model in a W&B sweep" or similar?

examples/wandb_sweeps/sweep.py
```diff
@@ -1,79 +1,97 @@
...
-    main()
+    logging.basicConfig(
+        format="%(levelname)s: %(asctime)s - %(message)s",
+        datefmt="%d-%b-%y %H:%M:%S",
+        level="INFO",
+    )
+    parser = argparse.ArgumentParser(description=__doc__)
+    parser.add_argument("--sweep_id", required=True, help="ID for the sweep.")
+    parser.add_argument(
+        "--entity", required=True, help="The entity scope for the project."
```
Adamits, line 84: Can we add to the description that this will be either your user or team name under which the project is contained?

Adamits:
So I tested this locally and I forgot to set some required yoyodyne-train args. Because of this, we log an error, but then it keeps trying to queue another run. Perhaps this is the default behavior of `count = None`? I will keep testing.


Screenshot 2025-02-25 at 10 54 43 AM

Adamits (changes requested):

kylebgorman:
> So I tested this locally and I forgot to set some required yoyodyne-train args. Because of this, we log an error, but then it keeps trying to queue another run. Perhaps this is the default behavior of `count = None`? I will keep testing.
> 
> Screenshot 2025-02-25 at 10 54 43 AM

I think with `count = None` it just runs until you kill it. If you want it to crash on the first bad thing I can make that happen, too, but it may not be what we want. I don't know.

examples/wandb_sweeps/sweep.py
```diff
@@ -1,79 +1,97 @@
...
+    )
+    parser.add_argument(
+        "--project", required=True, help="The project of the sweep."
+    )
+    parser.add_argument(
+        "--count",
+        type=int,
+        help="Number of runs to perform.",
```
Adamits, line 92: Can we add that the default of `None` will keep starting new runs after each one finished (or fails), and to specify a number here?

examples/wandb_sweeps/sweep.py
kylebgorman: I don't know if those two options are actually the only things an entity can be. I think I took this from some W&B docs. Could I just put that kind of information in the README for this directory instead?

examples/wandb_sweeps/sweep.py
kylebgorman: I could just make it required. Even somebody who wants it to run "practically forever" can just say 1,000 or something.

Adamits: True -- although I just saw that you already added this description I requested to the README. SO I think we are good there, but you are probably right right that making this required is clearer.

Adamits:
> > So I tested this locally and I forgot to set some required yoyodyne-train args. Because of this, we log an error, but then it keeps trying to queue another run. Perhaps this is the default behavior of `count = None`? I will keep testing.
> > Screenshot 2025-02-25 at 10 54 43 AM
> 
> I think with `count = None` it just runs until you kill it. If you want it to crash on the first bad thing I can make that happen, too, but it may not be what we want. I don't know.

Yeah good point. I guess since sweeps seem like the kind of thing you queue on a server and then forget about for a bit, it feels like unintended behavior here could be really annoying for a user. I noticed oyu explained `--count` in the README though so I think we have the needed documentation here already.

examples/wandb_sweeps/sweep.py
kylebgorman: Just made it required---lmk what you think.

examples/wandb_sweeps/sweep.py
Adamits: Yeah the README seems like the right place.

kylebgorman:
> Yeah good point. I guess since sweeps seem like the kind of thing you queue on a server and then forget about for a bit, it feels like unintended behavior here could be really annoying for a user. 

I agree.

examples/wandb_sweeps/sweep.py
Adamits: Yeah this seems safer!

Adamits:
> Could you check one last thing for me? Do you see the stats logged on W&B for the test runs you did?

Which stats?

kylebgorman:
Literally any of them. Do you see loss, accuracy, whatever, from the runs?

Adamits:
@kylebgorman Ah nevermind. I was not actually letting my model train before. I now trigger an error when wandb tries to log the hyperparamters: `TypeError: Object of type set is not JSON serializable`

kylebgorman:
Haven't seen that before. The only place I think we use a set that might be logged is the list of evaluation metrics, and a cast could fix that.

Adamits:
My local wandb was super old (0.15.4 -- yoyodyne requires >=0.18.5,<0.19.0).

- The hyperparameters set by the sweep are logged (but I think the defaults are not -- in other work for instance, we log the `encoder_cls` and decoder `arch`, which I use to validate the architecture of my logged results)
- I don't think any runtime metrics are being logged (e.g. loss, validation acc, etc)

Adamits:
Ok adding `--log_wandb` does the trick and now we seem to log everything. I think this behavior is slightly confusing since without that flag, we DO log to wandb, but only params sampled by the sweep.

Additionally, I now seem to have these other empty columns related to the sweep parameters (e.g. `parameters.batch_size.distribution`) in the "Runs" grid on the W&B interface.

kylebgorman:
> Ok adding `--log_wandb` does the trick and now we seem to log everything. I think this behavior is slightly confusing since without that flag, we DO log to wandb, but only params sampled by the sweep.

I can force that.

> Additionally, I now seem to have these other empty columns related to the sweep parameters (e.g. `parameters.batch_size.distribution`) in the "Runs" grid on the W&B interface.

Are you sure that's new?

Adamits:
> > Ok adding `--log_wandb` does the trick and now we seem to log everything. I think this behavior is slightly confusing since without that flag, we DO log to wandb, but only params sampled by the sweep.
> 
> I can force that.
> 
> > Additionally, I now seem to have these other empty columns related to the sweep parameters (e.g. `parameters.batch_size.distribution`) in the "Runs" grid on the W&B interface.
> 
> Are you sure that's new?

I am going to assume it is not new, and was just hidden by default in my older sweeps. Either way, its not actually a problem.

kylebgorman:
Actually, it looks like I already forced `--log_wandb` on line 59 of `sweep.py`. Is that something I need to act on?

Adamits:
Oh you are right. I just reran without `--log_wandb` and I get the runtime stats/metrics. They were just laggy to log on the first run.

Here is something weird though -- and maybe its expected and I just have to think about it more: while the run is training, I have all of the hyperparameters in W&B. After the job finishes, everything except the hyperparams sampled in the sweep config disappear, and I also don't have the val accuracy nor loss in my "Runs" grid (though we do have them logged somewhere b/c they show up in the workspace graphs.). Can you reproduce this?

kylebgorman:
> Here is something weird though -- and maybe its expected and I just have to think about it more: while the run is training, I have all of the hyperparameters in W&B. After the job finishes, everything except the hyperparams sampled in the sweep config disappear, and I also don't have the val accuracy nor loss in my "Runs" grid (though we do have them logged somewhere b/c they show up in the workspace graphs.). Can you reproduce this?

I have this resolved now, so I will submit. 

#322 may have been an issue; it seems to be necessary to ensure `val_accuracy` is logged to W&B (seemed to be a crapshoot beforehand). The other thing is that yes, when it actually logs stuff is laggy and weird. My suggestion if you're monitoring a sweep is as follows:

1. Hide all columns.
2. Then, unhide the ones you care about: for me, that's created-time, runtime, all the hyperparameters, validation accuracy, and validation loss. (and sweep ID if I'm comparing across multiple sweeps tied to the same project, e.g., to compare random vs. Bayesian tuning).
3. Sort in descending order by validation accuracy (max). 

Then everything will be there. Example below.

image

PR #309, opened 2025-02-22
removes unused 'merge' linear layers in ptr gen

Adamits:


kylebgorman (approved):
LGTM, surprised it's that easy. I'll test locally before merging.

Adamits:
I ran a quick test on my mac with an ptr-gen LSTM. Did not wait for it to converge though. I think we replaced the initialization a while back and just missed these when we updated, so its just basic cleanup unless I missed something.

Issue #303, opened 2025-02-06
Unused linear layers in pointer-generators


kylebgorman:
The GRU pointer-generator implementation includes a linear layer called [`merge_h`](https://github.com/CUNY-CL/yoyodyne/blob/54af2003b092117ecaf426fab3f1ca2f3b2970c2/yoyodyne/models/pointer_generator.py#L183), which presumably is intended to be used to "merge" hidden state from the source and features encoders; the LSTM one also includes an additional layer called `merge_c`, presumably for the cell state. However, these are not actually ever used, and as far as I can tell, they never were (I looked back a bit) and should either be removed or perhaps wired in. 

@Adamits, I have assigned this to you because I'm frankly afraid to touch this module much, but maybe you'd be comfortable addressing this.

Adamits:
Yeah I can take a look. I remember the implementation (but not the variable names),  so shouldnt take long. I can probably do this next week, or the week after at the latest.

Adamits:
Ok I took a look, and I believe these were intended to be used for equation 1 in this paper: https://aclanthology.org/K18-3013.pdf "The decoder’s hidden state si is initialised by applying an affine transformation on the concatenation of the last hidden states of the lemma and the tag encoders"

See here: https://github.com/CUNY-CL/yoyodyne/commit/c2466278b427e74f3953a2e26675c8c73c863307#diff-6b59b567f9423d07b2b2390072886dae0413c107cd87c0853a0b51cb27766ce8R342

Not sure when we replaced this exactly, but if we don't think we need this initialization then I can just remove those.

Adamits:
And to be clear, I think for all RNN models, we inherit from rnn.py a randomly initialized `self.h0` and `self.c0` now.

kylebgorman:
I think we want to do away with this unless it causes a regression in performance. 

I worked back a ways to find out when that affine transformation step was "disconnected" from the actual flow of information and I couldn't find any time where it was used, actually, so this may never have worked as intended, or if it has, it hasn't for a long time.

Adamits:
I thought I linked it in a commit. But I didnt link the line, here it is again: https://github.com/CUNY-CL/yoyodyne/commit/c2466278b427e74f3953a2e26675c8c73c863307#diff-6b59b567f9423d07b2b2390072886dae0413c107cd87c0853a0b51cb27766ce8R342

[...]

PR #310, opened 2025-02-22
Always uses learned initialization for pointer-generator

### LINKED ISSUE #304: Complex non-parallelism in pointer-generators
I was curious about [this `else`-block](https://github.com/CUNY-CL/yoyodyne/blob/54af2003b092117ecaf426fab3f1ca2f3b2970c2/yoyodyne/models/pointer_generator.py#L412) and [this one too](https://github.com/CUNY-CL/yoyodyne/blob/54af2003b092117ecaf426fab3f1ca2f3b2970c2/yoyodyne/models/pointer_generator.py#L445).

In the common case where the source and/or feature encoders do have hidden state, actually using said hidden state is non-parallel to what we do with non-pointer generator models like ordinary GRU encoder-decoders. For ordinay GRU-GRU encoder-decoders (I'll focus on that case so I don't have to talk about cell state), we ignore the final encoder hidden state. Rather, the initial input for the decoder (i.e., at the first step) is made out of a $B \times 1$ tensor containing `START_IDX` indices and a $B \times 1$ tensor containing the learned  `h0` parameter.

However, in the pointer-generator case, we use this encoder hidden state (i.e., if the encoder is an RNN too), and only build the decoder input from the learned parameter if the encoder doesn't have hidden state (i.e., if the encoder is a transformer). 

I propose that we modify the pointer-generator RNN model that we always use `START_IDX` and `h0` as the initial input to the decoder and thus we make the `else` clause the only option. Furthermore, this ought not to be hard because the `h0` parameter and the ability to generate the initial hidden state for the decoder already is inherited even if it's not being used. And this ought to simplify the rather nasty `forward` method.

Alternatively, we can get rid of the learned decoder parameters altogether. I don't care much.

@Adamits what do you think? Do you want to take a shot at this or should I?

kylebgorman:
The implementation of pointer-generator RNNs contains disjunctions in the `forward` method which determine whether to use the encoder's hidden state as the input to the decoder, or the learned `h0` (and `c0` for LSTMs). We remove the disjucntion and just use the learned ones all the time (cf. #303). 57 lines of code are removed; my tests find no accuracy or latency regressions. 

Closes #304.

This will also support my effort to have less GRU/LSTM-specific code at the `model` level; more coming soon.

### DIFF
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -402,9 +402,7 @@ class PointerGeneratorGRUModel(PointerGeneratorRNNModel, rnn.GRUModel):
         """
         encoder_out = self.source_encoder(batch.source)
         source_encoded = encoder_out.output
-        last_hiddens = self.init_hiddens(
-            len(batch), self.source_encoder.layers
-        )
+        last_hiddens = self.init_hiddens(len(batch))
         if not self.has_features_encoder:
             if self.beam_width > 1:
                 # Will raise a NotImplementedError.
@@ -428,9 +426,6 @@ class PointerGeneratorGRUModel(PointerGeneratorRNNModel, rnn.GRUModel):
         else:
             features_encoder_out = self.features_encoder(batch.features)
             features_encoded = features_encoder_out.output
-            last_hiddens = self.init_hiddens(
-                len(batch), self.source_encoder.layers
-            )
             if self.beam_width > 1:
                 # Will raise a NotImplementedError.
                 return self.beam_decode(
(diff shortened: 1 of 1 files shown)


Adamits (approved):
LGTM.

If it is easy, could you share the tests you ran and the accuracies in the two setups?

Adamits (approved):

kylebgorman:
> If it is easy, could you share the tests you ran and the accuracies in the two setups?

I'll generate those before I merge.

The same issue crops up in the transducer but I'll take care of that in a separate PR along with some other things.

kylebgorman:
Simple before and after evaluation results---within the range of possible noise. I'll go ahead and merge.

```
before:
    mri LSTM: .576 val accuracy
    pol LSTM: .926 val accuracy                                                          

after:    
    mri LSTM: .565 val accuracy
    pol LSTM: .926 val accuracy                                                          
```

PR #311, opened 2025-02-23
RNN encapsulation

### LINKED ISSUE #267: Separate features in one place
The list of models that require separate features is currently specified in two places: `train.py` and `predict.py`, and the latter is actually out of date. 

* A simple fix is to create a tiny module that records this in one location; I will try this first.
* Then once this is working, I will make this a `@property` of the actual models if that's possible.

kylebgorman:
There's a lot of boilerplate code in RNNs to deal with the fact that GRUs and LSTMs have a different return type. This PR abstracts away from this by replacing `base.ModuleOutput` with a struct `RNNState` that contains the hidden state and possibly the cell state (but crucially, not the output). 

```python
class RNNState(nn.Module):
    """Represents the state of an RNN."""

    hidden: torch.Tensor
    cell: Optional[torch.Tensor]   # LSTMs only.

    def __init__(self, hidden, cell=None):
        super().__init__()
        self.register_buffer("hidden", hidden)
        self.register_buffer("cell", cell)
```

This is then used as follows. Here's a GRU encoder module derived from `nn.GRU`; note it no longer returns the hidden state, because nothing uses it:

```python
class WrappedGRUEncoder(nn.GRU, WrappedRNNEncoder):
    """Wrapps GRU API to work with packing."""

    def forward(
        self,
        sequence: torch.Tensor,
        lengths: torch.Tensor,
    ) -> torch.Tensor:
        packed, _ = super().forward(self._pack(sequence, lengths))
        return self._pad(packed)
```

And here's a LSTM decoder derived from `nn.LSTM`, which does return the hidden state and cell state, packed up nicely.

```python
class WrappedLSTMDecoder(nn.LSTM):
    """Wrapped LSTM API to work with RNNState."""

    def forward(
        self, symbol: torch.Tensor, state: RNNState
    ) -> Tuple[torch.Tensor, RNNState]:
        assert state.cell is not None, "expected cell state"
        decoded, (hidden, cell) = super().forward(
            symbol, (state.hidden, state.cell)
        )
        return decoded, RNNState(hidden, cell)
```

As a result, almost all the module and model RNN code becomes almost completely independent of RNN type. That is, there are no longer special `forward` methods for, say, `GRUModel` vs. `RNNModel`; the only custom code in those is related to "registering" them to use a GRU vs. an RNN encoder and/or module. This enables a few other simple improvements:

* `h0` and `c0` are moved to their respective decoders, where they belong, since they are notionally decoder parameters. 
* Uniform, inherited methods are used for the learned initial state for the decoder and for generating the tensor of `START_IDX` inputs that is the first input given to the decoder.
* Beam search is greatly simplified and moved to a submodule; as a demonstration I quickly added beam search support to pointer-generator RNNs! I believe the other classes of RNNs (hard attention and transducers) can also gain this functionality without great effort, and transformers with a bit of effort.
* The only remaining `Union` types are for polymorphic returns from beam vs. greedy search.
* Names are standardized throughout. 

Here's what beam search looks like now for RNNs decoders:

```python
        ...
        state = self.decoder.initial_state(batch_size)
        beam = beam_search.Beam(self.beam_width, state)
        for _ in range(self.max_target_length):
            for cell in beam.cells:
                if cell.final:
                    beam.push(cell)
                else:
                    symbol = torch.tensor([[cell.symbol]], device=self.device)
                    logits, state = self.decode_step(
                        encoded, mask, symbol, cell.state
                    )
                    scores = nn.functional.log_softmax(logits.squeeze(), dim=0)
                    for new_cell in cell.extensions(state, scores):
                        beam.push(new_cell)
            beam.update()
            if beam.final:
                break
        return beam.predictions(self.device), beam.scores(self.device)
```

Even though I added a lot of documentation, this shaves off 390 lines of code total, most of it of the dangerous copy-paste variety. Performance is neutral within epsilon.

Closes #267.


kylebgorman:
I have tested this extensively on both CPU and GPU so I will merge so I can move onto other things (like finishing #284).

PR #318, opened 2025-03-25
Use torchmetrics and embed them in base model

### LINKED ISSUE #158: Implement Pytorch Metrics

kylebgorman:
Closes #158. This is a continuation of #284 to avoid nasty merge issues.

* Loss is computed as before, but streamlined somewhat.
* `torchmetrics` implementation of exact match accuracy is lightly adapted. This does everything in tensor-land and should keep things on devices. My tests confirm that accuracy is what it was before.
* A `torchmetrics`-compatible implementation of symbol error rate (here defined as the edit distance divided by sum of target lengths) is inserted here. This is heavily documented and it is compatible with our existing implementation. The hot inner loop is still on CPU, but as mentioned in the documentation, this is probably the best option and I don't observe any obvious performance penalty when enabling this. Note that the computation of the old one was not as normally defined and gave the number of edits per word, not the number of edits per gold target symbol.
* We do away with the `evaluation` module altogether. Rather we treat the metrics objects as nullables living in the base class, a design adapted from UDTube. Documentation in `metrics` explains how to extend this for new metrics.
* `Both loss and the metrics expect logits tensors to be of the shape B x target_vocab_size x seq_len, but the RNN library code wants it to be B x seq_len x target_vocab_size. I choose the former to be the "default" representation and let the RNN dimensionality be an implementational detail.
* In a small driveby I make the hard attention greedy decoding use the same target length trick as other modules.

The CLI interface is unimpacted.

Reviewer: I recommend taking a look at `yoyodye/model/base.py` and `yoyodyne/metrics.py`. The rest is just compatibility stuff, especially that dealing with the return shape (etc.)


kylebgorman:
Hi @Adamits any last-minute objections to the reworked version? I'll submit if I don't hear from you in a day or so.

Issue #313, opened 2025-03-06
Unused separate feature encoder with RNNs and transformers


kylebgorman:
If a separate features encoder is requested, the `BaseModel` initializes the encoder. Then, the models either combine evidence from the source and feature encodings, or pass the separate source and feature encodings to the decoder, in their `forward` methods:

* [hard attention RNNs](https://github.com/CUNY-CL/yoyodyne/blob/ac1964f4f93540d0419386bd35fadb578a201935/yoyodyne/models/hard_attention.py#L300)
* [pointer-generator RNNs](https://github.com/CUNY-CL/yoyodyne/blob/ac1964f4f93540d0419386bd35fadb578a201935/yoyodyne/models/pointer_generator.py#L426)
* [transducer RNNs](https://github.com/CUNY-CL/yoyodyne/blob/ac1964f4f93540d0419386bd35fadb578a201935/yoyodyne/models/transducer.py#L580)

However, this is missing in the [RNN models (both inattentive and attentive)](https://github.com/CUNY-CL/yoyodyne/blob/ac1964f4f93540d0419386bd35fadb578a201935/yoyodyne/models/rnn.py#L180) and in [the transformer](https://github.com/CUNY-CL/yoyodyne/blob/master/yoyodyne/models/transformer.py); either even mention the features encoder. So if one does something like:

* `--arch lstm --features_encoder_arch lstm`
* `--arch lstm --features_encoder_arch linear`
* `--arch transformer --features_encoder_arch linear` 

(to name a few examples), the model has a features encoder, but it's never trained or used.

I think I know how to add this for RNNs but I may need some help with the transformer.

Adamits:
Is it your understanding that this never existed? E.g. I recall https://github.com/CUNY-CL/yoyodyne/issues/47 is related.

How would you want to do it for the RNN? would we attend to both and concatenate the attention representations during decoding?

kylebgorman:
I thought it existed but I don't think it ever did. 

I think you concatenate the source and features encodings, yes.

kylebgorman:
Hi @Adamits, I was able to do ths with vanilla RNNs but I'm a bit at a loss how to make this work for (vanilla) transformers and for pointer-generator transformers. Would you be willing to consult with me re: this when you have a minute? 

[...]

Issue #301, opened 2025-02-06
Multi-source embeddings


kylebgorman:
There are various ways one can provide "multiple sources" as inputs. For instance, you can use "tags" to separate out the pieces, or you can encode it into the source symbols ("I'm a g from source 1" vs. "I'm a g from source 2" via `g_1` or `g_2`), or in Yoyodyne, if you have exactly two sources, the second one can be provided as the feature column. All of these can be done basically by hacking around on your TSV files. But there's one other way you might do this that can't be done in this fashion: appending "source embeddings" to each symbol in the source tape.

Formally, let us suppose that the encoder produces "contextual embeddings" or "encodings" of each symbol, giving us a tensor of shape $B \times L \times H$ where $B$ is the batch size, $L$ the max length of the batch, and $H$ the size of the hidden layer. (I put aside things like bidirectionality or multiple hidden layers for simplicity.) A source embedding is a learned $S \times D$ embedding table where $S$ is the number of sources and $D$ is the size of the embedding. E.g., if there are two sources and the size is 32, it'll be $2 \cross 32$. An encoding augmented with source embeddings will then be of size $B \times L \times (H + D)$, with the source embedding of the corresponding symbol "appended" to the bottom of the size-$H$ encoding.

Compared to the tag system, this is more explicit, and feels like it might be a stronger inductive bias because the system doesn't have to learn to pass through or attend to (depending) to said tags.

Implementationally, one could imagine that the source indices tensor (0 if the symbol is from source 1, 1 if it's from source 2, etc.), ofshapesize $B \times L$, is produced during the data loading phase and is attached to the batch object. Then, the encoder is run as before, producing the encoding of shape $B \times L \times H$. Then, if a source indices tensor is present in the batch, these are used to retrieve a source embedding tensor of shape $B \times L \times D$, and then the encoding and source embedding tensors are concatenated along the last dimension. This could all live within the base model, I think, with a call-back like interface: if there is a source indices tensor in the batch and a source embedding in the model, create the source embedding and concatenate it.

[...]

PR #319, opened 2025-04-06
Supports separate encoders with vanilla RNNs

kylebgorman:
This PR adds support for separate encoders in vanilla RNNs; it addresses #313 (but doesn't touch transformers so it does not close it). We adopt the merging approach from the hard attention models, which works with both "linear" and RNN features encoders. Suppose that the source encoder generates $B \times L \times E$ and the features encoder generates $B \times L' \times E'$. To create the final encoding, we first average features encodings across all positions producing a tensor of size $B \times 1 \times E'$. We then expand this tensor to $B \times L \times E'$ by broadcasting it across all positions in the source tensor. We then concatenate along the encoding dimension, producing a new tensor of size $B \times L \times (E + E')$.  The effect can be likened to spreading the features encoder information across all positions and thus roughly doubling the size of the encoding. 

# Fixes

* The `LinearEncoder` layer is misnamed because it's just a non-contextual embedding, with no actual affine projections. It is renamed `EmbeddingEncoder` and documentation is adjusted to match. 
* A separate `attention_input_size` argument is no longer necessary because it is always the same as `decoder_input_size`.
* The `NotImplementedError` is removed because it is in fact implemented.

# Alternatives considered

I experimented with concatenating along the length dimension instead. This works when there are two RNNs of the same hidden size and dimensionality but would break if the features encoder is just an embedding because E and E' may not match (indeed, one common rule of thumb is to use a hidden layer size 4x of the embedding size). We _could_ also add an affine $E' \times E$ projection linear layer to the embedding encoder, but that'd only be applicable for vanilla RNNs and wouldn't make sense elsewhere. If one wants a length concatenation effect, one can probably just use a single source/features encoder and the default concatenation, and the proposal in #301 provides an even more principled generalization of that idea. So I reject this idea.

# Future work

Complete #313 by adding something appropriate for transformers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,7 +244,7 @@ present, the `--features_encoder_arch` flag. Valid values are:
     the transformer encoder used with features; it concatenates source and
     features and uses a learned embedding to distinguish between source and
     features symbols.
--   `linear`: a linear encoder.
+-   `embedding`: a non-contextual embedding encoder.
 -   `gru`: a GRU encoder.
 -   `lstm`: a LSTM encoder.
 -   `transformer`: a transformer encoder.
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -11,7 +11,7 @@ exclude = ["examples*"]
 
 [project]
 name = "yoyodyne"
-version = "0.3.1"
+version = "0.3.2"
 description = "Small-vocabulary neural sequence-to-sequence models"
 readme = "README.md"
 requires-python = ">= 3.10"
--- a/yoyodyne/models/hard_attention.py
+++ b/yoyodyne/models/hard_attention.py
@@ -180,6 +180,7 @@ class HardAttentionRNNModel(rnn.RNNModel):
             transitions = self._apply_mono_mask(transitions)
         return scores, transitions, state
 
+    # FIXME remove me
     @property
     def decoder_input_size(self) -> int:
         if self.has_features_encoder:
@@ -321,8 +322,9 @@ class HardAttentionRNNModel(rnn.RNNModel):
         encoded = self.source_encoder(batch.source)
         if self.has_features_encoder:
             features_encoded = self.features_encoder(batch.features)
-            # Averages to flatten, then expands across the interior dimension
-            # to match encoder output.
+            # Feature information is averaged across all positions, broadcast
+            # across the length of the source, and then concatenated with the
+            # source encoding along the encoding dimension.
             features_encoded = features_encoded.mean(dim=1, keepdim=True)
             features_encoded = features_encoded.expand(-1, encoded.size(1), -1)
             encoded = torch.cat((encoded, features_encoded), dim=2)
--- a/yoyodyne/models/hard_attention.py
+++ b/yoyodyne/models/hard_attention.py
@@ -180,17 +180,6 @@ class HardAttentionRNNModel(rnn.RNNModel):
             transitions = self._apply_mono_mask(transitions)
         return scores, transitions, state
 
-    # FIXME remove me
-    @property
-    def decoder_input_size(self) -> int:
-        if self.has_features_encoder:
-            return (
-                self.source_encoder.output_size
-                + self.features_encoder.output_size
-            )
-        else:
-            return self.source_encoder.output_size
-
     def greedy_decode(
         self,
         source_encoded: torch.Tensor,
--- a/yoyodyne/models/hard_attention.py
+++ b/yoyodyne/models/hard_attention.py
@@ -321,11 +321,10 @@ class HardAttentionRNNModel(rnn.RNNModel):
         encoded = self.source_encoder(batch.source)
         if self.has_features_encoder:
             features_encoded = self.features_encoder(batch.features)
-            # Averages to flatten embedding; this is done as an alternative to
-            # the linear projection used in the original paper.
+            # Averages to flatten, then expands across the interior dimension
+            # to match encoder output.
             features_encoded = features_encoded.mean(dim=1, keepdim=True)
             features_encoded = features_encoded.expand(-1, encoded.size(1), -1)
-            # Concatenates with the encoded source.
             encoded = torch.cat((encoded, features_encoded), dim=2)
         if self.training:
             return self.decode(
--- a/yoyodyne/models/modules/__init__.py
+++ b/yoyodyne/models/modules/__init__.py
@@ -10,7 +10,7 @@ from .hard_attention import ContextHardAttentionLSTMDecoder  # noqa: F401
 from .hard_attention import HardAttentionRNNDecoder  # noqa: F401
 from .hard_attention import HardAttentionGRUDecoder  # noqa: F401
 from .hard_attention import HardAttentionLSTMDecoder  # noqa: F401
-from .linear import LinearEncoder
+from .embedding import EmbeddingEncoder
 from .rnn import AttentiveGRUDecoder  # noqa: F401
 from .rnn import AttentiveLSTMDecoder  # noqa: F401
 from .rnn import GRUDecoder  # noqa: F401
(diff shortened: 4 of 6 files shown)


Adamits:
Sorry for my delay.

Before I look at code, just want to make sure I understand:

> We then concatenate along the encoding dimension, producing a new tensor of size B×L×(E+E′).

Here L is the sequence length right (I was reading it as Layer at first)?  and L' was the seq length of the features? So we average over the feature representations, and concatenate that averaged vector onto each source symbol hidden representation.

Then for the alternate consideration:
> ...experimented with concatenating along the length dimension instead.
I am sort of confused about what this means. Are you just concatenating the matrices of symbol and feature encodings so we have a single sequence again (though the encodings happen by separate modules here ofc)?

I will remark that a third possibility is that you do no extra steps when encoding, and the decoder must attend to both the symbols and the features separately and then combine them (concatenation, affine transformer). I think this is what the pointer generator did at first. However, this complicates Yoyodyne a bit since then the decoder has to know if a separate features encoder was requested, and then conditionally attend to both and combine.

kylebgorman:
> Here L is the sequence length right (I was reading it as Layer at first)? and L' was the seq length of the features? So we average over the feature representations, and concatenate that averaged vector onto each source symbol hidden representation.

Correct.

>> ...experimented with concatenating along the length dimension instead.
> I am sort of confused about what this means. Are you just concatenating the matrices of symbol and feature encodings so we have a single sequence again (though the encodings happen by separate modules here ofc)?

Yes, that is the alternative I considered.


Sorry for my delay.

Before I look at code, just want to make sure I understand:

We then concatenate along the encoding dimension, producing a new tensor of size B×L×(E+E′).

Here L is the sequence length right (I was reading it as Layer at first)? and L' was the seq length of the features? So we average over the feature representations, and concatenate that averaged vector onto each source symbol hidden representation.

Then for the alternate consideration:

...experimented with concatenating along the length dimension instead.
I am sort of confused about what this means. Are you just concatenating the matrices of symbol and feature encodings so we have a single sequence again (though the encodings happen by separate modules here ofc)?

> I will remark that a third possibility is that you do no extra steps when encoding, and the decoder must attend to both the symbols and the features separately and then combine them (concatenation, affine transformer). I think this is what the pointer generator did at first. However, this complicates Yoyodyne a bit since then the decoder has to know if a separate features encoder was requested, and then conditionally attend to both and combine.

This is just the second possibility for me: it would be necessary to add a linear layer to what I'm now calling the `EmbeddingEncoder` to make it work in general because you don't expect the embedding dimensionality and hidden layer dimensionality to match. I don't think it's much of a complication for vanilla RNNs (it can be handled in the embedding encoder module plus the shared `forward` method for the abstract RNN model class and just in a few lines), I just think it's unlikely to be meaningfully different than using a single encoder anyways. 

It would be a potential complication for transformers with their fixed attention context though.

Adamits:
> This is just the second possibility for me: it would be necessary to add a linear layer to what I'm now calling the `EmbeddingEncoder` to make it work in general because you don't expect the embedding dimensionality and hidden layer dimensionality to match. I don't think it's much of a complication for vanilla RNNs (it can be handled in the embedding encoder module plus the shared `forward` method for the abstract RNN model class and just in a few lines), I just think it's unlikely to be meaningfully different than using a single encoder anyways.

I think our attention module already does a projection though, no? Anyway, we probably want to replace that with a simpler KQV attention at some point..

> 
> It would be a potential complication for transformers with their fixed attention context though.

I am having trouble understanding why this complicates the transformer. Is that in the case that we have two separate encodings to attend to, since the cross-attention is built into the torch implementation of a transformer decoder already? I think for the transformer ptr-gen I had to subclass it (just realized that should probably be updated to match the LSTM ptr-gen changes).

Adamits (approved):
Code changes LGTM. I will remark that I spent time thinking through the semantics of the changes (the comments describing what is happening, and linear --> embedding) and agree with all of them.

Adamits:
One more though: does this generalize for the edit action transducer, and the pointer generator? I think those might have had their own mechanisms for handling features, and I wonder if this conflicts with them?

kylebgorman:
> I think our attention module already does a projection though, no? Anyway, we probably want to replace that with a simpler KQV attention at some point..

I don't see how that's relevant. Attention takes a source tensor as an argument, and you logically can't concatenate tensors B x L x E (source encoding from an RNN encoder) and B x L' x E' (features encoding from the misnamed "linear encoder") on the second dimension if E != E', as will usually be the case under the current design.

So the alternative design requires me to add a linear projection of shape E' x E to the so-called linear encoder. It would look up the embedding and then apply the linear projection. (Right now it just looks up the embedding.) This would only be applicable for vanilla RNNs (it wouldn't make sense for pointer-generators or hard monotonic models, which use different strategies) and my proposal is that we don't bother.

> I am having trouble understanding why this complicates the transformer.

The transformer is implemented with a max source length for the positional embedding, and the RNNs don't. We could exceed this limit if we're doing additional concatenation along the length dimension. For instance if the max length is 128 and we have a long source string of 125 symbols and there are 5 features it'll crash (probably in a nasty way without an informative error message).

kylebgorman:
> One more though: does this generalize for the edit action transducer, and the pointer generator? I think those might have had their own mechanisms for handling features, and I wonder if this conflicts with them?

It matches the choice made in the hard monotonic model and the transducer. [It is different from the pointer-generator, which has a separate features attention module](https://github.com/CUNY-CL/yoyodyne/blob/e43920ff30affd4b71e084627eecf77e33fcec4e/yoyodyne/models/pointer_generator.py#L229). The transformer doesn't support this approach yet and I haven't studied the problem there.

So for me there are two other possibilities at most:

* the "alternate solution" described in the OP
* the pointer-generator solution linked to in this post

Adamits:
> > One more though: does this generalize for the edit action transducer, and the pointer generator? I think those might have had their own mechanisms for handling features, and I wonder if this conflicts with them?
> 
> It matches the choice made in the hard monotonic model and the transducer. [It is different from the pointer-generator, which has a separate features attention module](https://github.com/CUNY-CL/yoyodyne/blob/e43920ff30affd4b71e084627eecf77e33fcec4e/yoyodyne/models/pointer_generator.py#L229). The transformer doesn't support this approach yet and I haven't studied the problem there.
> 
> So for me there are two other possibilities at most:
> 
> * the "alternate solution" described in the OP
> * the pointer-generator solution linked to in this post

"The transformer doesn't support this approach"  --> I am fairly certain the pointer-generator solution is the one I implemented for the pointer generator transformer to be clear.

> The transformer is implemented with a max source length for the positional embedding

Oh so you are talking about the self-attention portion?

kylebgorman:
> "The transformer doesn't support this approach" --> I am fairly certain the pointer-generator solution is the one I implemented for the pointer generator transformer to be clear.

Oh, I was referring to the "vanilla transformer" here. [The pointer-generator transformer doesn't support separate features](https://github.com/CUNY-CL/yoyodyne/blob/e43920ff30affd4b71e084627eecf77e33fcec4e/yoyodyne/models/pointer_generator.py#L542) either though. (I added that check but that's because the features were being ignored otherwise.) 
 
> Oh so you are talking about the self-attention portion?

I wasn't making a distinction because I thought both had a length limit but I may be mistaken.

Adamits:
> > One more though: does this generalize for the edit action transducer, and the pointer generator? I think those might have had their own mechanisms for handling features, and I wonder if this conflicts with them?
> 
> It matches the choice made in the hard monotonic model and the transducer. [It is different from the pointer-generator, which has a separate features attention module](https://github.com/CUNY-CL/yoyodyne/blob/e43920ff30affd4b71e084627eecf77e33fcec4e/yoyodyne/models/pointer_generator.py#L229). The transformer doesn't support this approach yet and I haven't studied the problem there.

Ah ok. So I guess it supports separate features (https://github.com/CUNY-CL/yoyodyne/blob/e43920ff30affd4b71e084627eecf77e33fcec4e/yoyodyne/models/modules/transformer.py#L558C7-L558C32) but not a separate encoder.

> So for me there are two other possibilities at most:
> 
> * the "alternate solution" described in the OP
> * the pointer-generator solution linked to in this post


Ok, we can continue discussing for future work I guess. I think this PR looks good to go -- your solution implemented here seems like a good one.

PR #321, opened 2025-05-02
Coerces pointer generator layer matching

kylebgorman:
Currently the pointer-generator RNN implementation raises an error if the number of encoder and decoder layers doesn't match. This is sensible behavior in a library, but from the user perspective you might rather it just does the right thing. This also is a major issue for tuning the pointer-generator; in the past both @Adamits and I have just made a minor fork of the library to enforce the matching. What I propose here is that we enforce it by setting the number of decoder layers to match the number of encoder layers and then logging that to stderr. 

One alternative considered was to continiue raising the exception but then catching it (somewhere) and fixing it manually. But I'm not sure exactly how this would work.

### DIFF
--- a/yoyodyne/models/pointer_generator.py
+++ b/yoyodyne/models/pointer_generator.py
@@ -5,7 +5,7 @@ from typing import Callable, Optional, Tuple, Union
 import torch
 from torch import nn
 
-from .. import data, special
+from .. import data, special, util
 from . import base, beam_search, defaults, modules, rnn, transformer
 
 
@@ -99,8 +99,8 @@ class PointerGeneratorRNNModel(PointerGeneratorModel, rnn.RNNModel):
             self.decoder_layers = self.encoder_layers
             util.log_info(
                 "The number of encoder and decoder layers must match "
-                f"({self.encoder_layers} != {self.decoder_layers})."
-                "Setting both to {self.encoder_layers}".
+                f"({self.encoder_layers} != {self.decoder_layers}). "
+                f"Setting both to {self.encoder_layers}."
             )
         # Uses the inherited defaults for the source embeddings and encoder.
         if self.has_features_encoder:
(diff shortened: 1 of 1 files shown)


Adamits:
@kylebgorman Can you confirm that this dynamic layer update will be reflected in the logged hyperparameters?

kylebgorman:
> @kylebgorman Can you confirm that this dynamic layer update will be reflected in the logged hyperparameters?

It'll log, e.g.: 

```
	decoder_layers: 1
	encoder_layers: 2
```

and similarly in `hparams.yaml`, but it'll later log:

`The number of encoder and decoder layers must match (2 != 1). Setting both to 2.`

I don't know how you feel about this. I do think it sort of needs to happen automatically to allow for tuning though: do you have any other ideas?

kylebgorman:
@Adamits gentle ping on this, any thoughts?

Adamits:
Hey Kyle,

From the conversation above, my feeling is that I am likely to use the W&B csv as a sort of report for results, so as long as the layers are correct there (I.e. the call to their API with initialization hyperparams happens after the dynamic change) then LGTM!

Adamits (approved):

kylebgorman:
So the CSV there will be wrong (it'll say 2 encoder layers but 1 decoder layer, but it'll train with 2x2) but if you reuse those arguments again, the same thing will happen (it'll train wtih 2x2) so the "wrongness" in the record has no downstream impact. I don't see any obvious way to make things happen in the opposite direction since we can't make W&B do this for us.

kylebgorman:
On reflection, this is exclusively an issue for tuning. I think the solution, then, is to incorporate the ability to link values of one variable to another, and to do so within the tuning software itself rather than within Yoyodyne proper.

I will enforce this after #327, which is much better about catching (i.e., reporting as an informative error) model misconfigurations early.

Closing.

PR #326, opened 2025-07-07
Separate features everywhere

### LINKED ISSUE #47: Feature encoding enhancements

### LINKED ISSUE #136: Add option to concatenate features for pointer generator
I was just thinking, though our pointer-generator implementation(s) take care to encode features separately so that they are not used in the attention distribution for the pointer probabilities, I think it is worth making it easy to just consider features as other input symbols along with the lemmas. 

That is, to concatenate the features with the input just like we do for the 'vanilla' seq2seq models. This is just for comparison as sometimes these models learn things on their own without much intervention.

### LINKED ISSUE #313: Unused separate feature encoder with RNNs and transformers

kylebgorman:
This applies a plan outlined in ABQ:

1. Use a separate features encoder for all models where features are provided.
2. Enhance the `Embedding` module to be a `Linear` module again to fit this design.
3. Let the model handle the appropriate fusion of source and features encodings in a model-specific way; usually this is just a few lines in `forward` but in the case of the pointer-generator model, it lives inside of decoding.
4. Add a concatenation "example" script and docs for those who want the original behavior; it can be framed as a simple pre-processing step without much issue.
5. The feature-invariant transformer is considerably simplified because we have separate source and features modules now eliminating the need to "look up" which symbol is which.
6. There are substantial reductions in what needs to be passed around (`separate_features`, `features_vocab_size`, etc. no longer need to be).

A few open questions in this design:

1. In the linear module, should we also perform averaging across length? [right now no.]
2. Should we make the source and features encoder the same module if the source and features encoder are of the same type? (i.e., both LSTMs)? [right now no; easy to do though.]
3. What is the best practice for the "attention" in the inattentive RNN given that the RNN module concatenates encodings lengthwise? [the old logic is no longer expressible, and doing at the literal end of the sequence, as Sutskever claim to do (maybe as a bit of misdirection?!), doesn't work. I slightly modify the logic so we add up the embeddings at ENDs.]
4. Can one combine a RNN encoder and transformer encoder? [in theory yes, _but not generally_ because they will have different dimensionalities on the the output dimension; this PR works iff that dimension just happens to be the same and raises an informative error otherwise; note this is not a regression! Ultimately the solution here should be simply that users can configure source and features encoders separately and we justr check to make sure the configuration coheres; but this is going to involve adding a bunch of new flags to the library and I'd rather do this post #60.]

Evaluations are essentially all no-change (for the models that previously defaulted to symbol concatenation), suggesting the "one way to do it" solution is good design here.

- Closes #313.
- Closes #136 (as irrelevant).
- Closes #47.

### DIFF
--- a/yoyodyne/models/modules/transformer.py
+++ b/yoyodyne/models/modules/transformer.py
@@ -209,14 +209,14 @@ class WrappedTransformerDecoder(nn.TransformerDecoder):
         self,
         source_encoded: torch.Tensor,
         source_mask: torch.Tensor,
-        target_embedded: torch.Tensor,
+        target: torch.Tensor,
         target_mask: torch.Tensor,
         causal_mask: torch.Tensor,
     ) -> torch.Tensor:
         return super().forward(
             memory=source_encoded,
             memory_key_padding_mask=source_mask,
-            tgt=target_embedded,
+            tgt=target,
             tgt_key_padding_mask=target_mask,
             tgt_is_causal=True,
             tgt_mask=causal_mask,
@@ -261,9 +261,9 @@ class TransformerDecoder(TransformerModule):
             Tuple[torch.Tensor, torch.Tensor]: decoder outputs and the
                 embedded targets.
         """
-        target_embedded = self.embed(target)
+        embedded = self.embed(target)
         causal_mask = nn.Transformer.generate_square_subsequent_mask(
-            target_embedded.size(1),
+            embedded.size(1),
             device=self.device,
             dtype=bool,
         )
@@ -271,11 +271,11 @@ class TransformerDecoder(TransformerModule):
         decoded = self.module(
             source_encoded,
             source_mask,
-            target_embedded,
+            embedded,
             target_mask,
             causal_mask,
         )
-        return decoded, target_embedded
+        return decoded, embedded
 
     def get_module(self) -> WrappedTransformerDecoder:
         decoder_layer = nn.TransformerDecoderLayer(
@@ -357,7 +357,7 @@ class SeparateFeaturesTransformerDecoderLayer(nn.TransformerDecoderLayer):
         self,
         source_encoded: torch.Tensor,
         source_mask: torch.Tensor,
-        target_embedded: torch.Tensor,
+        target: torch.Tensor,
         target_mask: torch.Tensor,
         features_encoded: torch.Tensor,
         features_mask: torch.Tensor,
@@ -371,7 +371,7 @@ class SeparateFeaturesTransformerDecoderLayer(nn.TransformerDecoderLayer):
         Args:
             source_encoded (torch.Tensor): encoded source sequence.
             source_mask (torch.Tensor): mask for source.
-            target_embedded (torch.Tensor): current embedded target, which
+            target (torch.Tensor): current embedded target, which
                 may be the full target or previous decoded, of shape
                 B x seq_len x hidden_size.
             target_mask (torch.Tensor): mask for target.
@@ -383,9 +383,9 @@ class SeparateFeaturesTransformerDecoderLayer(nn.TransformerDecoderLayer):
             torch.Tensor.
         """
         output = self.norm2(
-            target_embedded
+            target
             + self._sa_block(
-                self.norm1(target_embedded),
+                self.norm1(target),
                 causal_mask,
                 target_mask,
                 is_causal=True,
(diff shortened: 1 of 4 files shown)


yoyodyne/models/modules/linear.py
```diff
@@ -0,0 +1,44 @@
...
+        bidirectional (bool).
+        *args: passed to superclass.
+        **kwargs: passed to superclass.
+    """
+
+    linear: nn.Linear
+
+    def __init__(self, *args, output_size: int, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.linear = nn.Linear(self.embedding_size, output_size)
```
bonham79, line 25: add flag for toggling the linear bias. (there's some random edge papers that explore adding bias to features so easier if we keep as a secret argument

yoyodyne/models/modules/transformer.py
```diff
@@ -340,127 +339,77 @@ def __init__(self, *args, nfeature_heads, **kwargs):
...
-        if self.norm_first:
-            x = x + self._sa_block(
-                self.norm1(x),
+        output = self.norm2(
+            target_embedded
+            + self._sa_block(
+                self.norm1(target_embedded),
+                causal_mask,
                 target_mask,
-                target_key_padding_mask,
```
bonham79, line 406: @Adamits can you check if the attention logic gets preserved after the removal?

bonham79 (approved):
LGTM

yoyodyne/models/modules/linear.py
bonham79: That's my point, just add a default `bias=True`. Never have to mess with it but don't have to remember that default.

PR #327, opened 2025-07-16
LightningCLI migration

### LINKED ISSUE #5: Testing

### LINKED ISSUE #60: Torch and Lightning 2.0.0

### LINKED ISSUE #151: Friendly predictor interface
I am working on a simple wrapper class which loads the model and yields predictions, but the interface in `predict.py` is somewhat unfriendly for this...it has nice abstractions but they're not the ones I need.

I propose to improve `predict.py` and related bits to make this process more friendly, and put my completed "predictor" class could go into the examples directory.

### LINKED ISSUE #328: Shared source and features encoder when it makes sense
As discussed in #326, it makes sense to allow users to share the source and features encoder (which already have the same notional vocabulary) and to provide a sensible interface for this.

kylebgorman:
We use LightningCLI to run things, eliminating argparse logic and all the complexities of `train.py` (which no longer is needed) and `predict.py` (which is moved into a callback). 

LightningCLI performs the command-line parsing (and config file parsing) by introspecting into module `__init__`s and docstrings. This means that there are no need to "wire" options for the datamodule; they just magically appear. Furthermore, there is no neeed to keep model arguments consistent or to place them all in the base model. The result is hundreds of lines of Python deleted.

Users now specify the source encoder module via arguments, and when features are being used, can either say `features_encoder: true` to simply reuse the source encoder for features (`features_encoder: false`, or no features encoder, is the default), or can specify a separate one. Introspection also takes care of all the wiring here. Each model checks whatever it believes the preconditions to be and is responsible for blending source and features as per #326.

Users gain a lot of new flexibility in specifying the encoders and decoders. Just a few examples ought to suffice, *all new* and in addition to already supported behaviors.

* A linear features encoder, can, for hard attention, pointer-generator models, and transducer models, be just about any desired output size. (For vanilla RNNs and transformers, an informative error is returned if it is not the expected size.)
* Pointer-generators can use different modules for source and features encoders, or different architectures, or even different numbers of layers, without any trouble.
* One can use LSTM source encoder and a transformer features encoder, if properly specified.
* Encoders and decoders can have different hidden sizes.
* Encoders and decoders can have different dropout probabilities.

Users can use arbitrary optimizers or schedulers (including all of those in `torch.nn`) thanks to introspection. We only include our warmup-based LR scheduler and a default dummy scheduler for completeness.

We inherit LightningCLI's subcommand ("verb") interface. The supported verbs are:

* `fit`: train
* `validate`: run the validation loop and log metrics; specify model with `--ckpt_path`
* `test`: predict on test and compute test accuracy (etc.); specify model with `--ckpt_path`
* `predict`: predict the targets; specify model with `--ckpt_path`

A huge number of example configuration files are provided and the process is documented in great detail. 

The sweep system is rewritten (it is somewhat simpler now and more compatible with Yoyodyne-Pretrained and UDTube).

A large integration test suite is added and documented. It uses a mix of toy data sets and two "realistic" ones and exercises every model-level architecture (and at least some of the "Franken-" options too), as well as some tests for misconfiguration. This uses basic config files with simple hyperparameters, but in future work, they can be rewritten to use hyperparameters discovered with tuning. Software to regenerate all data is provided.

If accepted, the current `master` branch will be renamed `legacy` in case it's needed in the future. 

Closes #5.
Closes #60.
Closes #151.
Closes #328.


configs/.bad/BAD_attentive_lstm.yaml
bonham79, line 1: Did you intend to add your .bads ?

configs/README.md
```diff
@@ -0,0 +1,129 @@
...
+    shared source and features encoder.
+
+## Putting it all together
+
+[`algother.yaml`](altogether.yaml) is a single combined configuration file with
+trainer, checkpoint, data, model, and prediction configuration information.
+
+To use, run:
+
+    yoyodyne fit --config /path/to/config.yaml
```
bonham79, line 129: Can you point me to an example line for inference post trainign?

configs/altogether.yaml
```diff
@@ -0,0 +1,53 @@
+trainer:
+  precision: bf16-mixed
```
bonham79, line 2: did you ever compare bf16-mixed vs bf16-true to see memory overhead change son latter?

configs/altogether.yaml
```diff
@@ -0,0 +1,53 @@
...
+      logging_interval: epoch
+  - class_path: lightning.pytorch.callbacks.EarlyStopping
+    init_args:
+      monitor: val_loss
+      min_delta: 0.0
+      patience: 10
+      verbose: true
+      mode: min
+  max_epochs: 100
+  max_time: 00:06:00:00
```
bonham79, line 22: remove max time example. Users may try to naievely do a 24 hr run.

configs/altogether.yaml
```diff
@@ -0,0 +1,53 @@
+trainer:
+  precision: bf16-mixed
+  logger:
+  - class_path: lightning.pytorch.loggers.CSVLogger
+    init_args:
+      save_dir: /Users/Shinji/models
+  - class_path: lightning.pytorch.loggers.WandbLogger
+    init_args:
+      save_dir: /Users/Shinji/models
```
bonham79, line 9: What's with this user profile?

configs/attentive_gru2lstm.yaml
```diff
@@ -0,0 +1,24 @@
...
+      dropout: .3
+      embedding_size: 300
+      hidden_size: 100
+  decoder_dropout: .3
+  decoder_hidden_size: 100
+  embedding_size: 300
+  optimizer:
+    class_path: torch.optim.Adadelta
+    init_args:
+      lr: 1
```
bonham79, line 24: LR 1?

configs/trainer.yaml
```diff
@@ -0,0 +1,21 @@
...
+      logging_interval: epoch
+  - class_path: lightning.pytorch.callbacks.EarlyStopping
+    init_args:
+      monitor: val_loss
+      patience: 10
+      verbose: true
+logger:
+  - class_path: lightning.pytorch.loggers.CSVLogger
+    init_args:
+      save_dir: /Users/Shinji/models
```
bonham79, line 17: weird directory

pyproject.toml
```diff
@@ -1,41 +1,25 @@
...
 license = { text = "Apache 2.0" }
+readme = "README.md"
+requires-python = ">= 3.10,<3.141"
 authors = [
-    {name = "Adam Wiemerslage"},
-    {name = "Kyle Gorman"},
-    {name = "Travis Bartley"},
+    { name = "Adam Wiemerslage" },
+    { name = "Kyle Gorman", email = "kylebgorman@gmail.com" },
+    { name = "Travis Bartley" },
```
bonham79, line 15: small vanglorious thing, i use my middle initial M.

pyproject.toml
```diff
@@ -1,41 +1,25 @@
...
-include-package-data = false
-
-[tool.setuptools.packages.find]
-include = ["yoyodyne*"]
-exclude = ["examples*"]
-
 [project]
 name = "yoyodyne"
-version = "0.3.3"
+version = "0.4.0"
```
bonham79, line 7: I think this can be 1 beta at this stage no?

requirements.txt
```diff
@@ -1,16 +1,12 @@
-black>=24.10.0
```
bonham79, line 1: do we have no over numpy versioning?

yoyodyne/data/collators.py
```diff
@@ -19,9 +19,9 @@ class Collator:
 
     has_features: bool
     has_target: bool
-    max_source_length: int = defaults.MAX_SOURCE_LENGTH
-    max_features_length: int = defaults.MAX_FEATURES_LENGTH
-    max_target_length: int = defaults.MAX_TARGET_LENGTH
+    max_source_length: int = defaults.MAX_LENGTH
+    max_features_length: int = defaults.MAX_LENGTH
+    max_target_length: int = defaults.MAX_LENGTH
```
bonham79, line 24: Feature length makes sense but target and source should be different length vars. (Zh/Ko/Ja tokens)

yoyodyne/models/expert.py
```diff
@@ -422,92 +421,19 @@ def find_prefixes(
...
-            sed_params_path.
+        path (str): path to SED parameter .pkl file.
+        oracle_factor (int): scaling factor to determine rate of expert
+            rollout sampling.
 
     Returns:
         expert.Expert.
     """
-
-    def _generate_data(
```
bonham79, line 449: where we move this?

bonham79 (changes requested):
Configs have some weird pathings that will cause a crash or weird artifacts. Can you replace with just a local path?

Major changes to expert, where that all go?

Which schedulers did you try inheriting from? (Just to keep track of any need for import guards).

Need Adam also to review, too big for just my eyes.

configs/.bad/BAD_attentive_lstm.yaml
kylebgorman: Yes. The [README in this directory](https://github.com/kylebgorman/yoyodyne/blob/lightning/configs/.bad/README.md) explains why:

```
This contains bad model configurations, i.e., those which are known to fail are instantiation time due to misconfiguration. They are mostly useful for testing.
```

configs/README.md
kylebgorman: See [here](https://github.com/kylebgorman/yoyodyne/tree/lightning?tab=readme-ov-file#inference-predict) in README for that:

```
yoyodyne predict --config path/to/config.yaml --ckpt_path path/to/checkpoint.ckpt
```

configs/altogether.yaml
kylebgorman: Yeah. Neither were a visible improvement but I tend to opt towards showing more rather than less features since discoverability is always so-so here.

configs/altogether.yaml
kylebgorman: If someone changes it to 24, I assume they understand what's going on.

pyproject.toml
kylebgorman: Can you submit a 1-line PR and then I'll merge in?

pyproject.toml
kylebgorman: You're thinking like I am; I will advance it to that as soon as I have #5 (good tests) going. I know how to integration-test this from previous work with LightningCLI.

requirements.txt
kylebgorman: Not that I'm aware of but I'll leave this comment unresolved until I test locally. 

I think that "over-pin" or whatever you want to call it was in place because of an unresolved issue with an ancient Lightning not playing nicely with very new Numpy.

yoyodyne/data/collators.py
kylebgorman: They can be, they just have the same default value. (And always did, I just sloppily gave them different names a long time ago.)

kylebgorman:
> Configs have some weird pathings that will cause a crash or weird artifacts. Can you replace with just a local path?

You're referring specifically to paths for files or directories? One absolutely can use local paths with this all but I don't and wouldn't recommend it myself. I could add a note to that effect though. 

> Major changes to expert, where that all go?

The bit of code in `train.py` that called the Maxwell trainer is gone, because there is no `train.py` anymore as instantiation is not stateful. (I promise this is worth the headache. That was like 200 lines of incredibly hairy code). Because of that we no longer need the code in the expert that mostly duplicated what's in `maxwell-train` to do looping over data files. So users will have to do it  beforehand. The reader is reminded of this twice in the README and I added [more detailed documentation about what they have to do here](https://github.com/kylebgorman/yoyodyne/tree/lightning/examples/maxwell).

> Which schedulers did you try inheriting from? (Just to keep track of any need for import guards).

It will work with any scheduler that is a subclass of `torch.optim.lr_scheduler.LRScheduler` whose first argument is the optimizer. I have tried the warmup inverse square root scheduler and the plateau one myself. I don't find any of the other ones useful. 

Similarly, the optimizer can be any subclass of `torch.optim.Optimizer` whose first argument is the model params. I've tried Adam and Adadelta.

This functionality for plugging in scheduler and optimizer is part of Lightning, not my creation, and they test it. 

> Need Adam also to review, too big for just my eyes.

Yeah I'm sorry. I'd have made it smaller if I could have ;)

configs/altogether.yaml
bonham79: Oh goddamnit. Why do you pilot the Yoyodyne Kyle?

Nah but let's clean these ups for non-otaku user directories. (e.g. ~/Documents should be neutral enough)

requirements.txt
bonham79: If you wouldn't mind, yeah those tests would be appreciated. Numpy versioning is just bane of my existence so would rather pin that down now before they or the PTL people decide to torment us

bonham79:
> You're referring specifically to paths for files or directories? One absolutely can use local paths with this all but I don't and wouldn't recommend it myself. I could add a note to that effect though.

Well the configs are referring to default directory structures no? I'd rather them be generic for more plug and play behavior than having users need to write in each time. Also, it's more a security issue exposing your local directory structure (yes, this is just a research codebase for a singular lab so no real issue, but I feel the specters of infosec past present and future).

> data files. So users will have to do it beforehand. The reader is reminded of this twice in the README and I added [more detailed documentation about what they have to do here](https://github.com/kylebgorman/yoyodyne/tree/lightning/examples/maxwell).

fair enough, it's not really critical for model training so enjoy the slim down. 

> Yeah I'm sorry. I'd have made it smaller if I could have ;)

no worries, just know my limitations. Not seeing anything on my end. All on you @Adamits

kylebgorman:
> Well the configs are referring to default directory structures no? I'd rather them be generic for more plug and play behavior than having users need to write in each time. Also, it's more a security issue exposing your local directory structure (yes, this is just a research codebase for a singular lab so no real issue, but I feel the specters of infosec past present and future).

I don't think it's possible to have it be both an absolute path and "just work" on an arbitrary system. Things like "~" are interpreted as literals, not in the OS-specific way, or I would have done "~/Documents/MyYoyodyneModels" or something. So I have two options:

* Don't mention the need to provide paths, even though if you fail to provide a path the config won't work.
* Provide a hypothetical path and remind users to set it.

I have added notes to the config README and the YAML examples that indicate which paths need to be set in a user-specific fashion.

requirements.txt
kylebgorman: This is resolved; I tested it with the oldest and newest Numpy (> than the old pin) and it works.

bonham79 (approved):
LGTM!

Following on offline discussion, let's make current main a 'legacy' branch before pushing. Then we enable 'user side unit testing.'

PR #388, opened 2026-08-20
Implementing optional student forcing

your-future-self:
These changes are in order to implement optional student forcing in the rnn and transformer models.


kylebgorman (commented):
Just focusing on files that don't need included here:

* Under `tests/testdata/data/*/` you have several `lightning_logs` subdirectories. Remove those and their contents from the PR.
* Also in these same set of subdirectories I see index.pkl files. Remove those.

I generally recommend using `git add` slowly, file by file, because otherwise you end up checking in a bunch of things you don't need. 

That'll shrink the size of the CL quite a bit, too.

kylebgorman (commented):
Just FYI, I still see the lightning_log subdirectories in the PR.

kylebgorman (commented):
One more thing to remove: `tests/testdata/*/models` and their contents. 

Get rid of any of the `.DS_Store` files. Most of them are in the aforementioned subdirectories, but check the list of files in the PR to be sure you don't miss any.

I also see something called `identity.zip`; IDK what it is but get rid of it.

yoyodyne/models/transformer.py
```diff
@@ -224,34 +224,86 @@ def get_decoder(
...
+    def global_decode(
+        self,
+        encoded: torch.Tensor,
+        encoded_mask: torch.Tensor,
+        target: torch.Tensor,
+        target_mask: torch.Tensor,
+    ) -> torch.Tensor:
+        """Decodes globally using teacher forcing.
+
+        This is only possible during strict teacher
```
kylebgorman, line 236: Weird line wraps, you have 79 characters...

yoyodyne/models/transformer.py
```diff
@@ -224,34 +224,86 @@ def get_decoder(
...
+        forcing training and validation;
+        Student forcing requires greedy decoding.
+
+        Args:
+        TODO: specify
+
+        Returns:
+        TODO: specify
```
kylebgorman, line 244: fill this in

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "yoyodyne"
-version = "0.5.27"
+version = "0.5.99"
```
kylebgorman, line 7: You can call this 0.5.28 since the previous version is 0.5.27 and there's nothing pending.

yoyodyne/models/transformer.py
```diff
@@ -224,34 +224,86 @@ def get_decoder(
...
+        target_mask: torch.Tensor,
+    ) -> torch.Tensor:
+        """Decodes globally using teacher forcing.
+
+        This is only possible during strict teacher
+        forcing training and validation;
+        Student forcing requires greedy decoding.
+
+        Args:
+        TODO: specify
```
kylebgorman, line 241: fill this in.

yoyodyne/models/pointer_generator/rnn.py
```diff
@@ -377,8 +378,8 @@ def greedy_decode_train_validate(
                 weights.
             source_encoded (torch.Tensor).
             source_mask (torch.Tensor).
-            features_encoded (torch.Tensor, optional).
-            features_mask (torch.Tensor, optional).
+            features_encoded (torch.Tensor, optional).  NOT IMPLEMENTED
```
kylebgorman, line 381: Is this still not implemented?

yoyodyne/models/pointer_generator/rnn.py
```diff
@@ -377,8 +378,8 @@ def greedy_decode_train_validate(
...
+            features_mask (torch.Tensor, optional).     NOT IMPLEMENTED
```
kylebgorman, line 382: ditto?

yoyodyne/models/transformer.py
```diff
@@ -27,23 +27,24 @@ class TransformerModel(base.BaseModel):
...
         self,
         *args,
-        decoder_attention_heads: int = defaults.ATTENTION_HEADS,
+        attention_heads: int = defaults.ATTENTION_HEADS,
         decoder_positional_encoding: (
             modules.BasePositionalEncoding | None
         ) = None,
-        teacher_forcing: bool = defaults.TEACHER_FORCING,
+        teacher_forcing: float = defaults.TEACHER_FORCING,
+        # teacher_forcing: bool = True,
```
kylebgorman, line 43: remove

yoyodyne/models/transformer.py
```diff
@@ -496,7 +585,8 @@ def forward(self, batch: data.Batch) -> torch.Tensor:
         prefix = batch.source.tensor
         if batch.has_features:
             prefix = torch.cat((prefix, batch.features.tensor), dim=1)
-        if (self.training or self.validating) and self.teacher_forcing:
+        raise NotImplementedError
```
kylebgorman, line 588: this looks like an indentation error but maybe you just didn't finish implementing this bit. should be straightforward?

yoyodyne/models/transformer.py
```diff
@@ -364,15 +452,16 @@ class CausalTransformerModel(base.BaseModel):
...
     classifier: nn.Linear
 
     def __init__(
         self,
         *args,
         attention_heads: int = defaults.ATTENTION_HEADS,
         positional_encoding: modules.BasePositionalEncoding | None = None,
-        teacher_forcing: bool = defaults.TEACHER_FORCING,
+        teacher_forcing: float = defaults.TEACHER_FORCING,
+        # teacher_forcing: bool = True,
```
kylebgorman, line 464: remove

kylebgorman (commented):
Looks like you got all the files that needed deleted. Here are some superficial comments. You can either do these now or do the manual merge step, up to you.

yoyodyne/models/transformer.py
your-future-self: Looking back at this, I'm pretty sure you're right about it being an indentation error, but to be honest I'm not completely sure. That said I'm pretty sure the student forcing stuff doesn't function for data with features, so I think a NotImplemented belongs somewhere but I'm not exactly sure where? Also, is making studentforcing etc. work for data with features within the scope of this project? Am I misunderstanding how complicated that would be?

yoyodyne/models/transformer.py
your-future-self: done

yoyodyne/models/transformer.py
kylebgorman: > Looking back at this, I'm pretty sure you're right about it being an indentation error, but to be honest I'm not completely sure.

If it's rendering correctly in GitHub (and this is not a guaranteed) then it's strictly speaking invalid Python since after a line-final colon, you have to increase the indentation. 

> That said I'm pretty sure the student forcing stuff doesn't function for data with features, so I think a NotImplemented belongs somewhere but I'm not exactly sure where? Also, is making studentforcing etc. work for data with features within the scope of this project? Am I misunderstanding how complicated that would be?

I may be missing something but I don't think it'll really interact in a deep way. The student/teacher distinction really only pertains to how the decoder sees the earlier target symbols, but by the time you're calling the decoding methods, the system has already FULLY processed the source and features symbols.

The only way in which you _might_ have to think about this at all is that if there are some simple interface changes inside of `forward` methods (which I think this is), and if, within these methods, there is a partial split into source+features vs. just source as input, you have to more or less duplicate the interface change along both paths. Does that make sense? I don't know if that actually applies here, but it could come up.

kylebgorman:
Right now, it looks like the `black` formatting is correct but there are `flake` errors in the uploaded version. Just FYI; addressing those is probably your next step.

kylebgorman:
> Right now, it looks like the `black` formatting is correct but there are `flake` errors in the uploaded version. Just FYI; addressing those is probably your next step.

And that seems to have done it, nice.

There are still some manual conflicts to resolve; see instructions under "This branch has conflicts that must be resolved" and let me know if you get stuck here.

